PR #2, opened 2018-01-10
Update to use Polymer's properties-mixin

sorvell:
NOTE, do not merge yet. Depends on Polymer 3.x being released on npm.


src/@polymer/polymer-lit-element.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+import { PropertiesMixin } from '../../@polymer/polymer/lib/mixins/properties-mixin.js';
+import { TemplateResult } from '../../lit-html/lit-html.js';
+import { render } from '../../lit-html/lib/lit-extended.js';
+
+export { html } from '../../lit-html/lit-html.js';
+export class PolymerLitElement extends PropertiesMixin(HTMLElement) {
+
+  ready() {
+    this.attachShadow({mode: 'open'});
```
TimvdLippe: Given this line, I think we can add a `shadowRoot: ShadowRoot` as property on this class. This should remove the need for `(el.shadowRoot as ShadowRoot)` in every usage of this field.

README.md
```diff
@@ -1,3 +1,55 @@
...
+    this.addEventListener('click', async (e) => {
+      this.whales++;
+      await this.nextRendered;
+      this.dispatchEvent(new CustomEvent('whales', {detail: {whales: this.whales}}))
+    });
+    super.ready();
+  }
+
+  // Render method should return a `TemplateResult` using the provided lit-html `html` tag function
+  render(props, html) {
```
justinfagnani: remove `html`

justinfagnani: Rather than `props` as an argument, I'd use destructuring:

```ts
render({foo}) {
   return html`... ${foo}`;
}
```

demo/polymer-lit-element.html
```diff
@@ -0,0 +1,75 @@
...
+        this.addEventListener('click', async (e) => {
+          this.whales++;
+          await this.nextRendered;
+          this.dispatchEvent(new CustomEvent('whales', {detail: {whales: this.whales}}))
+          console.log(this.shadowRoot.querySelector('.count').textContent);
+        });
+        super.ready();
+      }
+
+      render(props) {
```
justinfagnani: use destructuring

src/@polymer/polymer-lit-element.ts
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  invalidate() {
+    this._invalidateProperties();
+    return this.nextRendered;
+  }
+
+  get nextRendered() {
+    if (!this._nextRendered) {
+      this._nextRendered = new Promise((resolve) => {
```
justinfagnani: Later I think it'll be good to reject in case there's an exception during rendering. ie:

```ts
_flushProperties() {
  super._flushProperties();
  try {
    const result = this.render(this.__data);
    if (result) {
      render(result, this.shadowRoot!);
    }
  } catch (e) {
    if (this._nextRenderedReject) {
      this._nextRenderedReject(e);
    }
  }
```

PR #3, opened 2018-01-17
Fix the npm lock file

abdonrd:
1. `@polymer/polymer-lit-element` > `@polymer/lit-element`.

2. Bump packages versions.


abdonrd:
Friendly ping to @sorvell. :)

PR #13, opened 2018-01-21
WIP: Add decorator support when used from TypeScript

kenchris:
- Add TypeScript example in demo/ which uses decorators
- Add decorator for customElements.define and adding properties

Missing:
- Documentation
- Tests (the current tests don't run locally for me yet)

I would like initial feedback on the approach, especially about how the build system should work. I also created issues for problems that I ran into.

### DIFF
--- a/src/@polymer/lit-element-decorators.ts
+++ b/src/@polymer/lit-element-decorators.ts
@@ -1,0 +1,25 @@
+export type LitElementPropertyType = BooleanConstructor|DateConstructor|NumberConstructor|StringConstructor|ArrayConstructor|ObjectConstructor;
+
+export interface PropertyOptions {
+  type: LitElementPropertyType;
+}
+
+export const customElement = (tagname: keyof HTMLElementTagNameMap) => {
+  return <T extends { new (...args: any[]): HTMLElement }>(clazz: T) => {
+    window.customElements.define(tagname, clazz);
+    return clazz;
+  };
+}
+
+export const property = (options: PropertyOptions | LitElementPropertyType) => {
+  return (prototype: any, propertyName: string): any => {
+    createProperty(prototype, propertyName, options);
+  };
+}
+
+function createProperty(prototype: any, propertyName: string, options: PropertyOptions | LitElementPropertyType): void {
+  if (!prototype.constructor.hasOwnProperty('properties')) {
+    Object.defineProperty(prototype.constructor, 'properties', { value: {} });
+  }
+  prototype.constructor.properties[propertyName] = options;
+}
--- a/demo/ts/lit-element.html
+++ b/demo/ts/lit-element.html
@@ -1,0 +1,19 @@
+<!DOCTYPE html><!--
+@license
+Copyright (c) 2015 The Polymer Project Authors. All rights reserved.
+This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+Code distributed by Google as part of the polymer project is also
+subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+--><html><head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, minimum-scale=1.0, initial-scale=1.0, user-scalable=yes">
+  <title>lit-element demo</title>
+</head>
+<body>
+  <script type="module">
+    import { MyElement } from "./my-element.js"
+  </script>
+  <my-element bar="5" whales="5">Hi</my-element>
+</body></html>
--- a/demo/ts/my-element.ts
+++ b/demo/ts/my-element.ts
@@ -1,0 +1,58 @@
+// @ts-check
+import { LitElement, html } from '../../lit-element.js';
+import { customElement, property } from '../../lit-element-decorators.js'
+
+declare global {
+  interface HTMLElementTagNameMap {
+    "my-element": MyElement;
+  }
+}
+
+@customElement('my-element')
+export class MyElement extends LitElement {
+  @property(String) foo: string = 'foo';
+  @property({ type: Number }) bar: number;
+  @property(Number) whales: number = 0;
+  @property({ type: Array }) friends: any;
+
+  constructor() {
+    super();
+  }
+
+  ready() {
+    this.addEventListener('click', async () => {
+      this.whales++;
+      await this.nextRendered;
+      this.dispatchEvent(new CustomEvent('whales', {detail: {whales: this.whales}}))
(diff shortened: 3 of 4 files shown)


src/@polymer/lit-element-decorators.ts
```diff
@@ -0,0 +1,46 @@
...
+
+function createProperty(prototype: any, propertyName: string, options: PropertyOptions = {}): void {
+  if (!prototype.constructor.hasOwnProperty('properties')) {
+    Object.defineProperty(prototype.constructor, 'properties', { value: {} });
+  }
+  prototype.constructor.properties[propertyName] = options;
+}
+
+function reflectType(prototype: any, propertyName: string): any {
+  const { hasMetadata = () => false, getMetadata = () => null } = Reflect;
```
justinfagnani: Give how large The Reflect polyfill is and how the metadata API is no longer even speculative, I'd prefer to remove support altogether.

src/@polymer/lit-element-decorators.ts
```diff
@@ -0,0 +1,46 @@
...
+   */
+  type?: BooleanConstructor|DateConstructor|NumberConstructor|StringConstructor|ArrayConstructor|ObjectConstructor;
+  notify?: boolean;
+  reflectToAttribute?: boolean;
+  readOnly?: boolean;
+  computed?: string;
+  observer?: string|((val: any, old: any) => void);
+}
+
+export function customElement(tagname: string) {
```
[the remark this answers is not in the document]
kenchris: I am having the problem that TypeScript doesn't recognize HTMLElementConstructor, but it works with with Function or other types like StringConstructor. I assume that I need to add HTMLElementContructor to TS definitions somehow. Do you know where?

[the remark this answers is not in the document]
kenchris: interface HTMLElement is defined here: https://github.com/Microsoft/TypeScript/blob/master/lib/lib.d.ts#L8719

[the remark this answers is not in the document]
kenchris: OK, this seems to be working

```
export const customElement = (tagname: keyof HTMLElementTagNameMap) => {
  return <T extends { new (...args: any[]): HTMLElement }>(clazz: T) => {
    window.customElements.define(tagname, clazz);
    return clazz;
  };
}
```

src/@polymer/lit-element-decorators.ts
kenchris: ok, done

kenchris:
How do I add my element to `HTMLElementTagNameMap`?

I tried with

```
declare global {
  interface HTMLElementTagNameMap {
    "my-element": MyElement,
  }
}

@customElement('my-element')
export class MyElement extends LitElement {
...
}
```

But it still complains

dkundel:
Hey @kenchris this looks really cool but what made you choose lowercase names for the decorators? Stencil and Angular use caps and I feel like it would create a nicer experience to have sort of the same experience across the board. Just my two cents :)

justinfagnani:
@dkundel I'm not sure why Angular and Stencil chose caps, but every other case I've seen, including the Polymer decorators, and Google internal decorators, have been lowercase. In JS, generally just constructors are uppercase, so decorators wouldn't be.

dkundel:
Yeah I generally agree with the lowercase part. Those were just the only two occurrences where I've seen them actually being used in the wild so far. I checked with the proposal out of curiosity and they actually use lowercase for the same example as well. Just that they called it `defineElement` instead of `customElement`.

kenchris:
@justinfagnani your example decorators here: https://stackblitz.com/edit/lit-element-typescript?file=decorators.ts basically do everything that I did, but better. Why don't you add that to the project?

You are not using `tagname: keyof HTMLElementTagNameMap` as you suggested to me above though (which I couldn't figure out how to make work anyways :))

QuackMartins:
Hey guys, any updates on this?
We are working on our second project with lit-element... these decorators would really help us from repeating our class properties into `get properties()` method

pshihn:
@QuackMartins https://github.com/pshihn/alit-element in the meantime

QuackMartins:
Great, gonna use it

MatthieuLebigre:
I'd love to have that also !

Lodin:
There is almost finished [implementation](https://github.com/babel/babel/pull/7976) of decorator spec at Babel. Wouldn't it be more reasonable to use new decorator proposal that is going to enter Stage 3 instead of implementing deprecated one?

blikblum:
The current decorator proposal, being implemented by babel, is at stage 2. Only after some time, it will be moved to stage 3. This can take years.

typescript has a policy of implementing only stage 3. Anyway the new babel implementation supports the legacy syntax  with legacy option

Lodin:
@blikblum In accordance with Daniel Ehrenberg's [tweet](https://twitter.com/littledan/status/1017690405163405312) decorators can be moved to Stage 3 at July 2018. And we can always use Babel and Typescript together without waiting for their implementation of new proposal

43081j:
I'm not too sure of the differences between the spec TypeScript implements and the spec which is being adopted, but it seems like the usage will remain the same (of decorators).

So I think it is probably a good idea to still get this merged as it sounds like only implementation on our end will change rather than usage of our decorators. Also waiting for an undefined amount of time we _think_ isn't far away seems like a good way to get a PR going stale.

src/@polymer/lit-element-decorators.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export const customElement = (tagname: keyof HTMLElementTagNameMap) => {
+  return <T extends { new (...args: any[]): HTMLElement }>(clazz: T) => {
+    window.customElements.define(tagname, clazz);
+    return clazz;
+  };
+}
+
+export const property = (options: PropertyOptions | LitElementPropertyType) => {
+  return (prototype: any, propertyName: string): any => {
```
43081j: could we use something like `{ new(): LitElement }` here instead of `any`?

src/@polymer/lit-element-decorators.ts
```diff
@@ -0,0 +1,25 @@
...
+}
+
+export const customElement = (tagname: keyof HTMLElementTagNameMap) => {
+  return <T extends { new (...args: any[]): HTMLElement }>(clazz: T) => {
+    window.customElements.define(tagname, clazz);
+    return clazz;
+  };
+}
+
+export const property = (options: PropertyOptions | LitElementPropertyType) => {
```
HitkoDev: Could we add optional support for reflect-metadata? Perhaps something similar to https://github.com/pshihn/alit-element/blob/3d1cf69434295ed1037e2d8a30833960952a0703/src/alit-element-decorators.ts#L31

That would be great for code sanity as there will be no need maintain consistency between decorator and type definition

kenchris:
Not relevant any longer with the refactor

PR #15, opened 2018-02-08
Add `didRender` method 

sorvell:
Fixes https://github.com/PolymerLabs/lit-element/issues/5

* changes `nextRendered` promise to `renderComplete`
* alters render so that it's called within `_propertiesChanged`
* adds render helpers: `renderAttributes`, `classString`, `styleString`

### DIFF
--- a/src/@polymer/lit-element.ts
+++ b/src/@polymer/lit-element.ts
@@ -75,52 +75,76 @@ export class LitElement extends PropertiesMixin(HTMLElement) {
 
   private __renderComplete: Promise<any>|null = null;
   private __resolveRenderComplete: Function|null = null;
+  private __isInvalid: Boolean = false;
 
   protected ready() {
     this.attachShadow({mode: 'open'});
     super.ready();
   }
 
-  protected _shouldPropertiesChange() { return true; }
-
-  // TODO(sorvell): push to PropertiesChanged
-  protected _flushProperties() {
-    if (this._shouldPropertiesChange()) {
-      let changedProps = this.__dataPending;
-      this.__dataPending = null;
-      this._propertiesChanged(this.__data, changedProps, this.__dataOld);
-    }
+  /**
+   * Override which always returns true so that `_propertiesChanged`
+   * is called whenver properties are invalidated. This ensures `render`
+   * is always called in response to `invalidate`.
+   * @param {*} _props Current element properties
+   * @param {*} _changedProps Changing element properties
+   * @param {*} _prevProps Previous element properties
+   * @returns {boolean} Default implementation always returns true.
+   */
+  protected _shouldPropertiesChange(_props: any, _changedProps: any, _prevProps: any) {
+    return true;
   }
 
-  protected _propertiesChanged(props: any, ...args) {
-    super._propertiesChanged(props, ...args);
+  /**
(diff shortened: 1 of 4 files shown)


src/@polymer/lit-element.ts
```diff
@@ -12,53 +12,167 @@
...
+  /**
+   * Returns a promise which resolves after the element next renders.
+   */
+  get renderComplete() {
+    if (!this.__renderComplete) {
       // TODO(sorvell): handle rejected render.
-      this._nextRendered = new Promise((resolve) => {
-        this._nextRenderedResolver = resolve;
+      this.__renderComplete = new Promise((resolve) => {
+        this.__resolveRenderComplete = resolve;
```
kevinpschaaf: Rather than chaining off of the promise to clear the resolve function & promise, it could be done by wrapping resolve here right?

```js
this.__resolveRenderComplete = () => {
  resolve();
  this.__renderComplete = this.__resolveRenderComplete = null;
}
```
I guess it does a similar thing since `this.__renderComplete.then(...` is the first thing chained off the fresh promise and guaranteed to go first?

The async'ness is hard to reason about, i.e. whether there's a hole of time where someone could request the promise after it's been resolved, before it's been reset...

src/@polymer/lit-element.ts
```diff
@@ -12,53 +12,167 @@
...
+   * Override which always returns true so that `_propertiesChanged`
+   * is called whenver properties are invalidated. This ensures `render`
+   * is always called in response to `invalidate`.
+   * @param {*} _props Current element properties
+   * @param {*} _changedProps Changing element properties
+   * @param {*} _prevProps Previous element properties
+   * @returns {boolean} Default implementation always returns true.
+   */
+  protected _shouldPropertiesChange(_props: any, _changedProps: any, _prevProps: any) {
+    return true;
```
kevinpschaaf: This is fine, but just noted there's a closure warning on master where the default `_shouldPropertiesChange` returns `changedProps` instead of casting to a `Boolean`. Was going to ask to fix that here but just realized that it's in a separate repo. :(

src/@polymer/lit-element.ts
```diff
@@ -12,53 +12,167 @@
...
       // TODO(sorvell): handle rejected render.
-      this._nextRendered = new Promise((resolve) => {
-        this._nextRenderedResolver = resolve;
+      this.__renderComplete = new Promise((resolve) => {
+        this.__resolveRenderComplete = resolve;
+      });
+      this.__renderComplete.then(() => {
+        this.__resolveRenderComplete = this.__renderComplete = null;
       });
+      if (!this.__isInvalid && this.__resolveRenderComplete) {
```
kevinpschaaf: What is the case where `this.__resolveRenderComplete` could be null here?

src/@polymer/lit-element.ts
sorvell: Bug in Polymer. We can fix separately.

src/@polymer/lit-element.ts
sorvell: Appeases TypeScript.

src/@polymer/lit-element.ts
sorvell: PTAL

PR #19, opened 2018-02-17
Adds _createRoot method

### LINKED ISSUE #18: Provide a way to use lit without shadow dom
https://github.com/PolymerLabs/lit-element/blob/fdb3ba6b78253c9be4f4c2b917fa668af12d1c9c/src/%40polymer/lit-element.ts#L81

It would be great if we had a way to control if we want to use shadow dom like in polymer itself.

https://www.polymer-project.org/2.0/docs/devguide/dom-template#stamp-templates-in-light-dom - like here.

sorvell:
Fixes #18 

Allows users to specify a render target which defaults to the element shadowRoot. Just return `this` to render into the elements childNodes.

Also improves console warning describing that properties are set re-entrantly. Stack trace should now better help identify offending code.

### DIFF
--- a/src/@polymer/lit-element.ts
+++ b/src/@polymer/lit-element.ts
@@ -17,7 +17,6 @@ import { TemplateResult } from '../../lit-html/lit-html.js';
 import { render } from '../../lit-html/lib/lit-extended.js';
 
 export { html } from '../../lit-html/lib/lit-extended.js';
-
 /**
  * Renders attributes to the given element based on the `attrInfo` object where
  * boolean values are added/removed as attributes.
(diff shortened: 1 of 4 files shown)


src/@polymer/lit-element.ts
```diff
@@ -103,23 +113,26 @@ export class LitElement extends PropertiesMixin(HTMLElement) {
...
   }
 
-  protected _flushProperties() {
-    super._flushProperties();
-    if (this.__dataPending) {
-      console.warn(`Setting properties in response to properties changing considered harmful. Offending properties: ${Object.keys(this.__dataPending)}.`);
+  _shouldPropertyChange(property: String, value: any, old: any) {
+    const change = super._shouldPropertyChange(property, value, old);
+    if (change && this.__isChanging) {
+      console.warn(`Setting propertyes in response to other properties changing considered harmful. Setting '${name}' from '${this._getProperty(name)}' to '${value}'. See stack trace for more info.`);
```
kevinpschaaf: [sic] propertyes

Could make this `console.trace` instead of `console.warn` and then remove the "See stack trace for more info"

kevinpschaaf: [sic] propertyes

Could make this `console.trace` instead of `console.warn` and then remov e the "See stack trace for more info"

PR #20, opened 2018-02-21
Polymer TS types

e111077:
added typings to polymer mixins used by lit element.

NOTE: This is a stopgap method to get TS compiling nicely until the official polymer types come out.

Typings stolen from polymer 2 master branch types + a small few of my own.

There is an `import * as Polymer` [hack](https://github.com/PolymerLabs/lit-element/compare/PolymerTypes?expand=1#diff-da69066bf3e2239dfe94e19e90c87792R14) ~which we cannot figure out why we can't remove (causes using private type errors), but it all compiles and works!~ that is necessary because of the way typescript compiles the types for our mixin implementation.

Additionally did some extra work to make the tests compile as well (though there seem to be some problems with WCT not being able to load the correct things).


samuelli:
@e111077 what's the status of this?

samuelli:
Is this enough @e111077? This is already checked in: https://github.com/PolymerLabs/lit-element/blob/master/custom_typings/polymer.d.ts

e111077:
Nice, looks like this can be closed, then

PR #26, opened 2018-03-23
Use package names to import modules and fix lots of typings.

justinfagnani:
This also removes some weird folder structure and the bower_components symlink.

### DIFF
--- a/src/lit-element.ts
+++ b/src/lit-element.ts
@@ -11,12 +11,29 @@
  * subject to an additional IP rights grant found at
  * http://polymer.github.io/PATENTS.txt
  */
-import { PropertiesMixin } from '../../@polymer/polymer/lib/mixins/properties-mixin.js';
-import { camelToDashCase } from '../../@polymer/polymer/lib/utils/case-map.js';
-import { TemplateResult } from '../../lit-html/lit-html.js';
-import { render } from '../../lit-html/lib/lit-extended.js';
+import {
+  PropertiesChangedConstructor
+} from '@polymer/polymer/lib/mixins/properties-changed.js';
+import {PropertiesMixin} from '@polymer/polymer/lib/mixins/properties-mixin.js';
+import {
+  PropertiesMixinConstructor
+} from '@polymer/polymer/lib/mixins/properties-mixin.js';
+import {camelToDashCase} from '@polymer/polymer/lib/utils/case-map.js';
+import {render} from 'lit-html/lib/lit-extended.js';
+import {TemplateResult} from 'lit-html/lit-html.js';
 
-export { html } from '../../lit-html/lib/lit-extended.js';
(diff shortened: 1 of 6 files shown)


src/lit-element.ts
```diff
@@ -70,26 +87,26 @@ export function styleString(styleInfo: any) {
...
 
-
-
   /**
    * Returns an
-   * @returns {Node|ShadowRoot} Returns a node into which to render.
-  */
-  protected _createRoot() {
-    return this.attachShadow({mode: 'open'});
+   * @returns {Element|DocumentFragment} Returns a node into which to render.
```
rictic: Types can (and IMO should) be left out of jsdoc in ts files

src/lit-element.ts
```diff
@@ -112,13 +129,13 @@ export class LitElement extends PropertiesMixin(HTMLElement) {
    * @param {*} changedProps Changing element properties
    * @param {*} prevProps Previous element properties
    */
-  protected _propertiesChanged(props: any, changedProps: any, prevProps: any) {
+  _propertiesChanged(props: any, changedProps: any, prevProps: any) {
     this.__isChanging = true;
     this.__isInvalid = false;
     super._propertiesChanged(props, changedProps, prevProps);
     const result = this.render(props);
```
rictic: Where is render defined? I don't see it anywhere here.

src/lit-element.ts
```diff
@@ -101,7 +118,7 @@ export class LitElement extends PropertiesMixin(HTMLElement) {
    * @param {*} _prevProps Previous element properties
    * @returns {boolean} Default implementation always returns true.
    */
-  protected _shouldPropertiesChange(_props: any, _changedProps: any, _prevProps: any) {
+  _shouldPropertiesChange(_props: any, _changedProps: any, _prevProps: any) {
```
rictic: Prefer `object` to `any`, here and on `_propertiesChanged`

src/lit-element.ts
justinfagnani: done

src/lit-element.ts
justinfagnani: it's in here... it's abstract-ish

src/lit-element.ts
justinfagnani: done here and elsewhere

kevinpschaaf:
@justinfagnani @sorvell Sorry after the fact, but would it make sense to add a note to the readme that points out previous versions of lit-html used path-based references to dependencies, current version uses names, and thus name-to-path conversion is necessary before loading in the browser (with pointers to polymer-cli and/or other tools)?  Seems worth pointing out the change maybe?

samuelli:
The readme is also incorrect now, right?

PR #27, opened 2018-03-23
Update README

justinfagnani:


README.md
```diff
@@ -47,11 +70,59 @@ class MyElement extends LitElement {
...
+
+The one caveat is that LitElement imports Polymer using Polymer's package name,
+like so:
+
+```js
+import {PropertiesMixin} from '@polymer/polymer/lib/mixins/properties-mixin.js';
+```
+
+Browsers do not yet natively support such imports - they only import modules
+by full URL or absolute or relative path. Yet many tools - like bundlers,
```
kevinpschaaf: Can this section be somehow targeted at developers who are explicitly trying to use `LitElement` directly in the browser?  i.e., don't want to scare off people already in the Rollup/WebPack workflow by making this all sound somehow Polymer-specific.

README.md
```diff
@@ -47,11 +70,59 @@ class MyElement extends LitElement {
...
+```bash
+$ npm i -g polymer-cli@next
+```
+
+Then serve your project with module specifier rewriting turned on:
+```bash
+$ polymer serve --npm --module-resolution=node
+```
+
+You can add `"npm": true` and `"moduleResultion": "node"` to your polymer.json
```
samuelli: s/moduleResultion/moduleResolution/

README.md
justinfagnani: Do you have a suggestion? I don't read this as scaring off anyone, just that our old path pattern isn't that well supported.

README.md
justinfagnani: fixed

README.md
kevinpschaaf: Well, I was thinking something along the lines of, 

> If you're using WebPack, Rollup, or another tool that already handles compiling out package names from import specifiers for development, you can skip this section.

🤷‍♂️

sorvell:
Already incorporated similar text via another PR and consulted the language here so closing this.

PR #32, opened 2018-04-18
Render api

sorvell:
Adds a render-based API. Needs reivew. Users have complained that having some methods named for "properties" and others for "rendering" is confusing.


src/lit-element.ts
```diff
@@ -161,18 +162,28 @@ export class LitElement extends PropertiesMixin
...
     throw new Error('render() not implemented');
   }
 
+  protected _shouldRender(_props: object, _changedProps: object, _prevProps: object) {
+    return true
+  }
+
+  protected _willRender(_props: object, _changedProps: object, _prevProps: object) {}
+
+  protected _applyRender(result: TemplateResult, _root: Node) {
```
kevinpschaaf: Need API docs for `_shouldRender`, `_willRender`, `_applyRender`.

Think we also need a section in README on lifecycle, and what one might do in each callback (i.e. what's its intended use case/purpose).  The one I'm most unclear on is `_willRender`.  What good thing can you do here that you can't do in `_render`?

README.md
```diff
@@ -1,6 +1,59 @@
...
+the entire rendering state declaratively, and lit-html ensures that updates
+are fast by creating the static DOM once and smartly updating only the parts of
+the DOM that change.
+
+For basic usage:
+
+  1. Create a class that extends `LitElement`.
+  1. Implement a static `properties` getter that returns the element's properties
+  (which automatically become observed attributes).
+  1. Then implement a `_render(props)` method and use the element's
```
samuelli: So is this the only required method to implement for a class that extends `LitElement`?

README.md
```diff
@@ -1,6 +1,59 @@
...
+  Implement to directly control rendered DOM. Typically this is not needed as `lit-html`
+  can be used in the `_render` method to set properties, attributes, and
+  event listeners. This method can, however, be useful to call methods
+  on rendered elements or to specifically react to the rendered
+  state of the DOM.
+
+  * `_createRoot()`: Implement to customize where the
+  element's template is rendered by returning an element into which to
+  render. By default this creates a shadowRoot for the element.
+  To render into the element's childNodes, return `this`.
```
samuelli: Gut reaction to these APIs is that they're private and I shouldn't really be touching these when writing my element.

README.md
```diff
@@ -1,6 +1,59 @@
...
+  * `_render(props)`: Implement to describe the element's DOM using `lit-html`. Ideally,
+  the `_render` implementation is a pure function using only `props` to describe
+  the element template.
+
+  * `_didRender(props, changed, old)`: Called after element DOM has been rendered.
+  Implement to directly control rendered DOM. Typically this is not needed as `lit-html`
+  can be used in the `_render` method to set properties, attributes, and
+  event listeners. This method can, however, be useful to call methods
+  on rendered elements or to specifically react to the rendered
+  state of the DOM.
```
samuelli: This reads really confusingly to me. I think this API is necessary but is not really captured by the description here. I don't really understand what/why `props, changed, old` are.

One use case I've encountered is usually around needing to operate on the rendered DOM. Writing the `_render()` function exposes a string, but there's cases in which you need to operate on the produced DOM. For example, immediately focusing on a child element; this isn't possible from the `_render` function.

README.md
```diff
@@ -1,6 +1,59 @@
...
+
+  * `_createRoot()`: Implement to customize where the
+  element's template is rendered by returning an element into which to
+  render. By default this creates a shadowRoot for the element.
+  To render into the element's childNodes, return `this`.
+
+  * `invalidate`: Call to force the element to asynchronously re-render regardless
+  of whether or not any property changes are pending.
```
samuelli: Reaction: validation/invalidation introduces a new dimension that isn't described anywhere else. I'd implemented a similar method called `requestRender()` or something like `forceRender()` seems like a more suitable from an intuitive API design standpoint.

README.md
```diff
@@ -1,6 +1,59 @@
...
+
+LitElement uses [lit-html](https://github.com/Polymer/lit-html) to render into the
+element's Shadow DOM and [Polymer's](https://github.com/Polymer/polymer)
+PropertiesMixin to help manage element properties and attributes.
+
+LitElement uses a functional reactive pattern where any changes to element
+properties are batched and then rendered asynchronously. Users describe
+the entire rendering state declaratively, and lit-html ensures that updates
+are fast by creating the static DOM once and smartly updating only the parts of
+the DOM that change.
```
samuelli: Readme comments:
 * would be great to put a minimal code snippet which demonstrates the useful of utilising this base class.
 * Pull this paragraph out into highlighted dot points - 
   * **Batched rendering** - LitElement watches changes to properties and asynchronously renders these batched changes to reduce rendering overheads, 
   * **Declarative state**
  etc. with extended descriptions for each.
 * "functional reactive"? Sounds scary.

README.md
```diff
@@ -1,6 +1,59 @@
-# lit-element
+# LitElement
 
-## Base class for creating custom elements using Polymer and lit-html.
+## A simple base class for creating custom elements rendered with lit-html.
+
+LitElement uses [lit-html](https://github.com/Polymer/lit-html) to render into the
+element's Shadow DOM and [Polymer's](https://github.com/Polymer/polymer)
+PropertiesMixin to help manage element properties and attributes.
```
notwaldorf: does this need a'linkin?

README.md
notwaldorf: yah!

README.md
notwaldorf: s/changed/listOfChangedProperties maybe? And maybe show a snippet that has a simple/useful `didRender` example?

lit-element.d.ts
```diff
@@ -0,0 +1,113 @@
+/**
+ * @license
+ * Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
```
frankiefu: 2018

README.md
```diff
@@ -51,7 +100,6 @@ customElements.define('my-element', MyElement);
   <my-element whales="5">hi</my-element>
 ```
 
-## Known Issues
-* This element does not yet work with the ShadyCSS polyfill. Support is coming soon!
-* API is subject to minor changes.
-* See [lit-html](https://github.com/Polymer/lit-html) for more info.
+Known Issues:
```
notwaldorf: why no `##`

demo/lit-element.html
```diff
@@ -10,7 +10,7 @@
   <meta charset="UTF-8">
   <meta name="viewport" content="width=device-width, minimum-scale=1.0, initial-scale=1.0, user-scalable=yes">
   <title>lit-element demo</title>
-  <script src="../node_modules/@webcomponents/webcomponentsjs/webcomponents-lite.js"></script>
+  <script src="../../node_modules/@webcomponents/webcomponentsjs/webcomponents-lite.js"></script>
```
notwaldorf: is this called `-bundle` now?

lit-element.js
```diff
@@ -0,0 +1,193 @@
...
+    constructor() {
+        super(...arguments);
+        this.__renderComplete = null;
+        this.__resolveRenderComplete = null;
+        this.__isInvalid = false;
+        this.__isChanging = false;
+    }
+    ready() {
+        this._root = this._createRoot();
+        super.ready();
```
frankiefu: The compiled js seems to be out of date. For example I don't see the `this._firstRendered();` here

README.md
kenchris: But in a way, it would be bad to use names without prefix as then we cannot use those in the future in standards as it might break the web

README.md
sorvell: Done

README.md
sorvell: Added note.

sorvell:
@samuelli PTAL

README.md
```diff
@@ -3,60 +3,98 @@
...
   * `_createRoot()`: Implement to customize where the
   element's template is rendered by returning an element into which to
   render. By default this creates a shadowRoot for the element.
   To render into the element's childNodes, return `this`.
 
-  * `invalidate`: Call to force the element to asynchronously re-render regardless
+  * `requestRender`: Call to request the element to asynchronously re-render regardless
   of whether or not any property changes are pending.
 
   * `renderComplete`: Returns a promise which resolves after the element next renders.
```
samuelli: Would be great to sort this API by the lifecycle. eg.
```
_shouldRender()
_render()
_didRender()
renderComplete()
```

README.md
```diff
@@ -3,60 +3,98 @@
...
 
 
-Example:
+## Bigger Example
 
 ```javascript
-import {LitElement, html} from 'node_modules/@polymer/lit-element/lit-element.js'
+import {LitElement, html} from '@polymer/lit-element';
 
```
samuelli: s/Synched/synced/. Former is past tense.

PR #44, opened 2018-05-07
Make ts compilable

e111077:
it seems as if it wont compile correctly if you have `"modules": "not commonjs"`, so you have to specify to resolve modules with `"moduleResolution": "commonjs"` to make it compile yet output es2015 modules

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 *.js
 *.js.map
 *.d.ts
+!custom_typings/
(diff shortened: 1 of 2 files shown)


justinfagnani:
We shouldn't ignore `/test`, in fact the current `.gitignore` looks pretty messed up. Can you change it to ignore *.js, *.d.ts and *.js.map? You'll have to exclude custom_typings/ and add an empty .npmignore too.

e111077:
oops didn't see that test/index.html was a thing beforehand

PR #59, opened 2018-05-13
Export svg template literal from lit-extended

### LINKED ISSUE #56: Export svg from lit-extended as well as html
I've been trying to use lit-element to render svg icons within a component, but found it strange to have to import `svg` from lit-html. I had some trouble getting the right version of both dependencies in sync so that rendering would work right.

I think it makes sense to import both rendering functions from the same place and not have to install a separate dependency with all the versioning headaches that causes.

TimvdLippe:
This also exports the `svg` template literal as defined in https://github.com/Polymer/lit-html/blob/1dc43255d172f412fd19784c190764d05605cc16/src/lib/lit-extended.ts#L25-L29

Fixes #56

### DIFF
--- a/src/lit-element.ts
+++ b/src/lit-element.ts
@@ -28,7 +28,7 @@ export {
 export {
   PropertiesMixinConstructor
 } from '@polymer/polymer/lib/mixins/properties-mixin.js';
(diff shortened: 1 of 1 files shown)


TimvdLippe:
Actually, should we maybe export everything from `lit-html`?

PR #60, opened 2018-05-14
chore(README.md): properties-mixin.html -> properties-mixin.js

noahlaux:
This PR fixes `PropertiesMixin` link from `properties-mixin.html` to `properties-mixin.js`.


rajsite:
Was about to make the same pull request 👍

PR #78, opened 2018-05-25
Assign HTMLElementWithProperties as constant

joncass:
When using the closure compiler on a compiled version of LitElement, it is complaining that `The class in an extends clause must be a qualified name`. I believe the issue is that `class LitElement extends PropertiesMixin(HTMLElement)` attempts to extend the result of a function rather than a defined symbol. I think this change might fix that issue.

Please let me know if you think there is a different way to solve this, or if my code requires changes of any kind.

Thanks!

### DIFF
--- a/src/lit-element.ts
+++ b/src/lit-element.ts
@@ -88,8 +88,8 @@ export function styleString(styleInfo: {[name: string]: any}) {
   return o.join('; ');
 }
 
-export class LitElement extends PropertiesMixin
-(HTMLElement) {
+const HTMLElementWithProperties = PropertiesMixin(HTMLElement);
+export class LitElement extends HTMLElementWithProperties {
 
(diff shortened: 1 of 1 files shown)


joncass:
I just linked my github account to my google ldap, so CLA should be all set now. Let me know if there are any issues.

joncass:
Thanks @TimvdLippe. I don't have write access - is this a change you want to merge, or do you want to wait for someone else to do so?

TimvdLippe:
I want to wait for the maintainers of this element to take a look. I just
wanted to state my approval and coin the idea of running closure on Travis.

joncass:
I'm no longer convinced that this change is safe. Closing this PR while I investigate further.

joncass:
Created as an issue instead: #80

PR #92, opened 2018-06-10
Extract WCT configuration to wct.conf.json

TimvdLippe:
Instead of passing the arguments manually in the `package.json` script, it is now in the JSON file. This also enables `FireFox` testing.


TimvdLippe:
This is included in #97 now, as these changes are required to get the Travis build passing.

PR #95, opened 2018-06-10
Add [hidden] convention to example docs

### LINKED ISSUE #50: ([hidden]) convention
In some of your examples, readme and demo, you have `display: block`, which is fine but, is it worth including hidden in these examples, just because it's good form?

```CSS
<style>
  :host {display: block;}
  :host([hidden]) {display: none;}
</style>
```

https://developers.google.com/web/fundamentals/web-components/best-practices#add-a-host-display-style-that-respects-the-hidden-attribute

TimvdLippe:
Fixes #50

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,9 @@ class MyElement extends LitElement {
         :host {
           display: block;
         }
+        :host([hidden]) {
+          display: none;
+        }
       </style>
       <h4>Foo: ${foo}</h4>
       <div>whales: ${'🐳'.repeat(whales)}</div>
--- a/demo/lit-element.html
+++ b/demo/lit-element.html
@@ -49,6 +49,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
               display: block;
             }
 
+            :host([hidden]) {
+              display: none;
+            }
+
             .count {
               color: green;
             }


web-padawan:
FYI: https://meowni.ca/hidden.is.a.lie.html

`!important` is missing.

TimvdLippe:
Hm yes that is true. Should we even endorse hidden then?!

Westbrook:
I always understood the `[hidden] { display: none !import;}` as being a reaction to the code in this PR being missing. So that when it wasn't internal to a component, you would have to use `!important` to override from the outside Doesn't the `:host([hidden])` get the `display: none;` specificity high enough to match what we'd expect and leave any external styling to effect the element as if it were any native tag? Looking at prepping a generator for new components at my office where clarity on this point would be of great benefit.

Link2Twenty:
Westbrook is correct important is not required as long as you do `:host([hidden]) {display:none;}`

`:host {display:block}` overrides the build in hidden styles but `:host([hidden]) {display:none;}` overrides the override.

Link2Twenty:
It might be worth mentioning somewhere that by default elements will be `display:inline;` so it you can avoid setting display do.

Also if we wanted to cut down on lines something like this would work too.

```html
<style>
  :host(:not([hidden])) {
    display: block;
  }
</style>
```

Here's a super simple [example](https://jsfiddle.net/link2twenty/06kjtL17/).

Issue #42, opened 2018-05-06
Non-shadow dom API


LarsDenBakker:
As a long-time user of Polymer, one of my biggest issues with Polymer.Element is it's hard buy-in with shadow dom. There is a method that you can overwrite to avoid creating a shadow root, but you really have to know what you're doing and then there are still issues around styling. Especially on browsers that need shady dom / css.

Shadow dom is a powerful tool and makes a lot of sense for building re-usable UI components. When building large complex applications, it is useful to split code up into separate components But in our experience, using shadow dom for all these private components has a lot of downsides. Closely associated components are unnecessarily isolated making tests more complicated, shared styles are duplicated many times over and especially on browser that do not have native shadow dom performance is negatively impacted. I think in the applications we build, over half our components do not need to use shadow dom. 

The current support for elements without shadow dom is to overwrite _createRoot and return 'this'. I'm worried that this is not a very intuitive API and already requires some in-depth knowledge of how this works internally. I'm hoping that there can be more official support for creating elements without shadow dom and that it's use cases are covered by documentation, demos and tests. I'd be happy to contribute this where possible.

moebiusmania:
@LarsDenBakker this is a real good point, and for my experience a shared one. BTW YouTube itself *doesn't* use full Shadow DOM for the same reasons.

The `_createRoot => this` trick may be intuitive or not, but at least as opposite of Polymer it let you decide to use Shadow DOM or not individually, the main downside is that you loose support for `<slot>` element.  

It would be nice to have a sort of bultin `slot` polyfill activated when you opt to not use Shadow DOM.

yavorski:
Started a new project with `lit-html` and `LitElement` 2 days ago.
I opted out of the shadow dom mainly because I wanted to use external stylesheet for my elements and to not have duplicated inlined css.
I also need to support firefox as well, and web components there are hidden behind a flag (really hope to be enabled by default soon).
Now `<link rel="stylesheet" href="my-element-style.css">` is supported in chrome/safari, but again shady css is really messy with firefox i am not sure it can be even pollyfilled.

sorvell:
We'd like to encourage a default of using Shadow DOM because it's the right option when making a re-usable element since these need encapsulated dom and css. 

You have a point that when making an application, it is sometimes reasonable not to use Shadow DOM, and this is why we provide the `_createRoot` override point.

We're open to making this API simple and would welcome a proposal / PR.

LarsDenBakker:
[...]

PR #99, opened 2018-06-12
Add LitElementLight

LarsDenBakker:
Splits the LitElement class into an abstract LitElementBase class, which is implemented by LitElement (renders to shadow root) and LitElementLight (renders to light dom). See https://github.com/Polymer/lit-element/issues/42 for use cases for LitElementLight.

I'm not sure about the name 'LitElementLight', but I could not think of anything else.

I think this results in a clearer API than overriding methods. It clearly explains the intent to developers, and leaves the project flexible for different strategies in the future. For example when/if https://github.com/w3c/webcomponents/issues/468 becomes a thing, LitElementLight will have a use case within shared components as well.

Fixes https://github.com/Polymer/lit-element/issues/42

### DIFF
--- a/src/lit-element.ts
+++ b/src/lit-element.ts
@@ -19,7 +19,8 @@ import {
   PropertiesMixinConstructor
 } from '@polymer/polymer/lib/mixins/properties-mixin.js';
 import {camelToDashCase} from '@polymer/polymer/lib/utils/case-map.js';
-import {render} from 'lit-html/lib/shady-render.js';
+import {render as shadyRender} from 'lit-html/lib/shady-render.js';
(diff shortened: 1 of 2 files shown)


LarsDenBakker:
I think this could be split up into separate files, but it would mean exposing multiple files and it seems like there is quite a lot in one file already. I'm not sure about the approach that should be taken, for instance which files should contain classString and styleString and which files should export html and svg from lit-html.

sorvell:
Sorry for the delay and thanks for the PR. We've elected to maintain the current way of customizing the rendering element (now by implementing `createRenderRoot`) as noted [here](https://github.com/Polymer/lit-element/issues/42#issuecomment-416306482).

PR #105, opened 2018-06-17
Changes properties change trace to warning

bennypowers:
LitElement warns when properties are changed as a result of property changes. That's all well and good, but the way it goes about it is disruptive. Instead of using the standard `console.warn`, LitElement logs a `console.trace`. This can be really annoying at times, as it fills up the console with  information which may not be immediately relevant. A console warning, which is in any event preferred for this type of message, will also print the trace, but collapsed behind a drill-down control, and should be used instead.

Example:
Developer has some improper data flow which causes this warning to log. However, it's not causing UX problems, so she's de-prioritized that for now. However, an important UI bug has come up and she needs to log information to the console to fix it. In this case, the log stack traces from `_shouldRender` are disruptive to her work.

### DIFF
--- a/src/lit-element.ts
+++ b/src/lit-element.ts
@@ -207,7 +207,7 @@ export class LitElement extends PropertiesMixin
   _shouldPropertyChange(property: string, value: any, old: any) {
     const change = super._shouldPropertyChange(property, value, old);
     if (change && this.__isChanging) {
-      console.trace(
+      console.warn(
           `Setting properties in response to other properties changing ` +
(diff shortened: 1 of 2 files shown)


sorvell:
@AndreasGalster The idea here is that rendering an element should not cause it to render again. At worst this can be an infinite loop and at best it's inefficient. That said, it's possible there may be some cases when there's a valid reason to do this. If you'd like us to consider this more, I recommend filing a separate issue, thanks.

Issue #110, opened 2018-06-22
Prettier


merlinnot:
Hey, I'd like to propose switching from clang-format to [Prettier](https://prettier.io). I was recently going through the code of LitElement and the output of clang-format seems less clean and readable, at least to me. Other Google projects are happily using Prettier, like `firebase-js-sdk`.

What do you guys thinks? I'd be more than happy to make a PR.

sorvell:
Thanks for the feedback. Not saying you're wrong, but we're going to stick with clang-format for now since it's got some configuration options we need.

PR #112, opened 2018-06-22
Changed types of parameters of lifecycle callbacks to be more strict.

### LINKED ISSUE #111: More strict types of callback parameters
Types of lifecycle callback parameters are very generic, namely `object`. We could leverage polymorphic `this` and mapped (`Partial` in particular) types to make it more strict.

I'll follow up with a PR.

merlinnot:
We now leverage polymorphic `this` and mapped types.

Two files were reformatted by clang-format (latest version since it's not defined in `devDependencies`). You can use `?w=1` query to see GitHub diff with no whitespace changes.

Resolves #111

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@polymer/lit-element",
-  "version": "0.5.2",
+  "version": "0.6.0",
   "description": "Polymer based lit-html custom element",
   "license": "BSD-3-Clause",
   "repository": "Polymer/lit-element",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@polymer/lit-element",
-  "version": "0.6.0",
+  "version": "0.5.2",
   "description": "Polymer based lit-html custom element",
   "license": "BSD-3-Clause",
   "repository": "Polymer/lit-element",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@polymer/lit-element",
-  "version": "0.6.0",
+  "version": "0.5.2",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@polymer/lit-element",
-  "version": "0.5.2",
+  "version": "0.6.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,14 +26,14 @@
         "@babel/template": "7.0.0-beta.51",
         "@babel/traverse": "7.0.0-beta.51",
         "@babel/types": "7.0.0-beta.51",
-        "convert-source-map": "^1.1.0",
-        "debug": "^3.1.0",
-        "json5": "^0.5.0",
-        "lodash": "^4.17.5",
-        "micromatch": "^3.1.10",
-        "resolve": "^1.3.2",
-        "semver": "^5.4.1",
-        "source-map": "^0.5.0"
+        "convert-source-map": "1.5.1",
+        "debug": "3.1.0",
+        "json5": "0.5.1",
+        "lodash": "4.17.10",
+        "micromatch": "3.1.10",
+        "resolve": "1.7.1",
+        "semver": "5.5.0",
+        "source-map": "0.5.7"
       },
       "dependencies": {
         "source-map": {
@@ -51,10 +51,10 @@
       "dev": true,
       "requires": {
         "@babel/types": "7.0.0-beta.51",
-        "jsesc": "^2.5.1",
-        "lodash": "^4.17.5",
-        "source-map": "^0.5.0",
-        "trim-right": "^1.0.1"
+        "jsesc": "2.5.1",
+        "lodash": "4.17.10",
+        "source-map": "0.5.7",
+        "trim-right": "1.0.1"
       },
       "dependencies": {
         "source-map": {
@@ -103,7 +103,7 @@
       "requires": {
         "@babel/helper-function-name": "7.0.0-beta.35",
         "@babel/types": "7.0.0-beta.35",
-        "lodash": "^4.2.0"
+        "lodash": "4.17.10"
       },
       "dependencies": {
         "@babel/code-frame": {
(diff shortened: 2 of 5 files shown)


merlinnot:
In the end I had to manually alter formatting to align with tslint rules. I've created an issue so we can discuss switching to Prettier: #110

merlinnot:
@TimvdLippe will do.

merlinnot:
@TimvdLippe It actually caught some typing errors in tests already: ec9628840ce877b741cfe16239eb1b106064c536

I can understand that you'd like to see a test that is failing with `object` in type definitions and is passing with `this`. `object` is more generic than `this`, tho, so I see no way to test it using the TypeScript compiler itself, I'd probably have to go with something like [this](https://github.com/Microsoft/types-publisher/blob/master/src/tester/test.ts). Is it worth it?

TimvdLippe:
@merlinnot Ah no, that is actually a really good catch :+1: I wonder if we can use the same `Partial` solution for `static get properties()`, but let's put that in a separate PR

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@polymer/lit-element",
-  "version": "0.5.2",
+  "version": "0.6.0",
```
TimvdLippe: Let's not do that right now but in a separate PR

merlinnot:
Hi @TimvdLippe @sorvell, any chance to get it merged? :)

TimvdLippe:
Steve is still on vacation :wink:

merlinnot:
How about this week? 😃

justinfagnani:
So I'm not so sure about these typings here. The `this` type isn't actually an accurate description of what can appear in the arguments to the callbacks.

 1. By default the keys there are only the keys of the `properties` block, not the keys of the whole class. For instance, the `this` type will include `_render`, `_shouldRender`, etc.
 2. Additional keys may appear if `setProperty` is called with names not in the properties block, as might happen with custom getter/setters.
 3. We might be able to use mapped types to creating more accurate typings directly from the `properties` block, but that work work with the upcoming decorators. As far as I know, there's no way to have conditional types take into account decorators.

I think the current types, which admit that we don't actually know the type of the properties that well might be the most accurate.

cc @rictic who worked on some mapped typings for the Polymer properties block as well.

TimvdLippe:
@justinfagnani I have been playing with this and I came up with the following mapped types and conditional types: [TypeScript playground](http://www.typescriptlang.org/play/#src=%2F%2F%20Lit%20Element%20source%20code%0D%0Aabstract%20class%20LitElement%3CProperties%20%3D%20object%3E%20%7B%0D%0A%20%20protected%20abstract%20_render(props%3A%20Properties)%3A%20string%3B%0D%0A%7D%0D%0A%0D%0Atype%20ToConstructor%3CT%3E%20%3D%0D%0A%20%20T%20extends%20string%20%3F%20typeof%20String%20%3A%0D%0A%20%20T%20extends%20number%20%3F%20typeof%20Number%20%3A%0D%0A%20%20T%20extends%20boolean%20%3F%20typeof%20Boolean%20%3A%0D%0A%20%20T%3B%0D%0A%0D%0Atype%20PropertiesConstructor%3CProperties%3E%20%3D%20%7B%0D%0A%20%20%5BProp%20in%20keyof%20Properties%5D%3A%20ToConstructor%3CProperties%5BProp%5D%3E%0D%0A%7D%0D%0A%0D%0A%2F%2F%20User%20code%20source%20code%0D%0A%0D%0Ainterface%20Object%20%7B%20%7D%0D%0A%0D%0Aclass%20AnotherClass%20%7B%20%7D%0D%0A%0D%0Ainterface%20SubElementProperties%20%7B%0D%0A%20%20foo%3A%20string%3B%0D%0A%20%20bar%3A%20boolean%3B%0D%0A%20%20obj%3A%20Object%3B%0D%0A%20%20clazz%3A%20AnotherClass%3B%0D%0A%7D%0D%0A%0D%0Aclass%20SubElement%20extends%20LitElement%3CSubElementProperties%3E%20implements%20SubElementProperties%20%7B%0D%0A%20%20foo%3A%20string%3B%0D%0A%20%20bar%3A%20boolean%3B%0D%0A%20%20obj%3A%20Object%3B%0D%0A%20%20clazz%3A%20AnotherClass%3B%0D%0A%0D%0A%20%20static%20get%20properties()%3A%20PropertiesConstructor%3CSubElementProperties%3E%20%7B%0D%0A%20%20%20%20return%20%7B%0D%0A%20%20%20%20%20%20foo%3A%20String%2C%0D%0A%20%20%20%20%20%20bar%3A%20Boolean%2C%0D%0A%20%20%20%20%20%20obj%3A%20Object%2C%0D%0A%20%20%20%20%20%20clazz%3A%20AnotherClass%0D%0A%20%20%20%20%7D%0D%0A%20%20%7D%0D%0A%0D%0A%20%20_render(%7B%20foo%2C%20bar%20%7D%3A%20SubElementProperties)%20%7B%0D%0A%20%20%20%20if%20(bar)%20%7B%0D%0A%20%20%20%20%20%20return%20''%20%2B%20foo%3B%0D%0A%20%20%20%20%7D%0D%0A%20%20%20%20return%20'no'%0D%0A%20%20%7D%0D%0A%7D%0D%0A%0D%0Aconst%20foo%20%3D%20new%20SubElement()%3B%0D%0Aconst%20num%3A%20boolean%20%3D%20foo.bar%3B)

Still, there is a need to declare the types of the `properties` separately and TypeScript does not allow these types to be mixed in into the class. Thus, you need to duplicate these field types in the class as well. This adds the following burden to the user:

1. Declare the properties interface
1. Implement the interface in the class
1. Use the `PropertiesConstructor` type to obtain the correct `static get properties` type

justinfagnani:
@TimvdLippe I'm pretty sure there's a way to use the inferred type of the `properties` getter without having to manually define the interface. Regardless, that's still not accurate as any call to `setProperty()` can introduce new keys in the the callbacks. And that technique won't work with class fields and decorators.

Basically, we can assert that the type _usually_ has a particular shape, or has at least some fields, but maybe others, so `object` is actually about as good as we can do.

sorvell:
Really sorry for the delay on this and thanks for the PR. The API has changed a good bit (merged to master and to be released in version 0.6) so I think it makes sense to close this. We can re-evaluate based on the new codebase.

Issue #136, opened 2018-07-31
Misleading warning about property changes


AusIV:
### Description
Under certain conditions, the warning:

```Setting properties in response to other properties changing considered harmful```

gets logged (often repeatedly) when an error occurred during the render process, rather than when properties changed in response to other properties

#### Live Demo
https://stackblitz.com/edit/lit-element-example-tjfppw

#### Steps to Reproduce
1. Throw an error in the _render() function.
2. Some time later, set the value of a property.

#### Expected Results
The error from the render function should be obvious in the browser's development console, and there should be no errors related to setting properties.

#### Actual Results
Any properties set after an error in the _render() function trigger the "Setting properties in response to other properties changing considered harmful" warning, which includes a stack trace. This pushes the original error up past the top of the console, making it look like the problem pertains to setting properties, distracting from the real problem

#### Cause

When properties are set, it checks to see if `element.__isChanging` is true, and logs a warning if it is (see https://github.com/Polymer/lit-element/blob/master/src/lit-element.ts#L209). 

In _flushProperties(), it sets `__isChanging = true`, calls `super._flushProperties()`, and after a success sets `__isChanging = false`. This means failures in `super._flushProperties()` leave `__isChanging == True`, so the warning will trigger until the next successful call to `_flushProperties()`.

### Browsers Affected

- [ x] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari 11
- [ ] Safari 10
- [ ] IE 11

(I'm pretty sure it will effect all of the above, but I've only test on Chrome)

### Versions

- lit-element: v0.5.2
- webcomponents: vX.X.X

sorvell:
This issue should no longer reproduce since we've removed the warning for the upcoming `0.6` release, currently in preview as `dev`. See the new [update lifecycle](https://github.com/Polymer/lit-element#advanced-update-lifecycle) for more info.

PR #137, opened 2018-07-31
Ensure that __isChanging gets set to false after errors

AusIV:
This commit addresses #136

Wrapping the call to `super._flushProperties()` in a try{}finally{}
ensures that this.__isChanging gets set back to false even if an
error occurs within `super._flushProperties()`, which will avoid
distracting from the real error by logging reentrancy warnings.

I haven't been able to figure out how to write a good test for this.
If I throw an exception in the render function it fails the test
even thoug that's an expected behavior. As the render process
happens asynchronously, I can't trivially catch the error to avoid
failing the test.

### DIFF
--- a/src/lit-element.ts
+++ b/src/lit-element.ts
@@ -192,8 +192,11 @@ export class LitElement extends PropertiesMixin
   _flushProperties() {
     this.__isChanging = true;
     this.__isInvalid = false;
-    super._flushProperties();
-    this.__isChanging = false;
+    try {
+      super._flushProperties();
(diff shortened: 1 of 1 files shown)


sorvell:
Sorry for the delay and thanks for the PR. The code has significantly changed in master (and the `dev` preview of the upcoming `0.6` release). The warning and flag have been removed. See the new [update lifecycle](https://github.com/Polymer/lit-element#advanced-update-lifecycle) for more info.

PR #145, opened 2018-08-13
add a typescript transform

### LINKED ISSUE #8: Add TypeScript custom transform to generator accessors at compile time.
Once we have runtime decorators with #7, we should write a custom tsc transform that compiles the decorators into accessors, which should reduce the class definition time of LitElements.

43081j:
Fixes #8.

## DO NOT MERGE

This is stupidly experimental, it may not belong here and we don't yet have a way to use it! I'm placing it here mostly so i can get some code review and suggestions of how we deal with it.

## 🎈 

So I wrote this transform which can be used with the typescript compiler to apply both `@customElement` and `@property` at build time.

Few problems:

* Where do we put it?
* We would need to introduce a build script people can optionally use (as the TS team said they will _never_ allow transforms in tsconfig afaik)
* It needs to produce a commonjs output I suppose, so our existing tsconfig won't build it correctly

How it works:

### `@customElement`

It takes the first argument of the decorator and appends the `customElements.define` statement after the class. 

### `@property`

It maps the typescript type to a constructor type (falls back to `Object`) and adds an entry to the properties getter.

If the properties getter doesn't exist, it makes it.

If the properties getter exists but has no body, it makes it.

---

Also left a comment or two to show you where im stuck and the code i don't like.

cc @aomarks @justinfagnani

### DIFF
--- a/src/ts-transform.ts
+++ b/src/ts-transform.ts
@@ -67,7 +67,7 @@ class CustomElementVisitor extends Visitor implements IDecoratorVisitor {
   }
 
   visit(node: ts.Node): ts.VisitResult<ts.Node> {
-    if (!ts.isClassDeclaration(node)) {
+    if (!ts.isClassDeclaration(node) || !node.name) {
       return ts.visitEachChild(node, (child) =>
         this.visit(child), this.context);
     }
@@ -87,11 +87,11 @@ class CustomElementVisitor extends Visitor implements IDecoratorVisitor {
     }
 
     const name = args[0];
-    const defineCall = ts.createCall(
+    const defineCall = ts.createStatement(ts.createCall(
       ts.createPropertyAccess(
-        ts.createIdentifier('define'),
-        'customElements'
-      ), undefined, [name]);
+        ts.createIdentifier('customElements'),
+        ts.createIdentifier('define')
+      ), undefined, [name, node.name]));
 
     return [node, defineCall];
   }
@@ -209,7 +209,7 @@ class PropertyVisitor extends Visitor implements IDecoratorVisitor {
   }
 }
 
-export function decoratorTransformer(): ts.TransformerFactory<ts.SourceFile> {
+export function decoratorTransformer<T extends ts.Node>(): ts.TransformerFactory<T> {
   return (context) => {
     const visitors = [
       new CustomElementVisitor(context),
--- a/src/ts-transform.ts
+++ b/src/ts-transform.ts
@@ -1,0 +1,231 @@
+import * as ts from 'typescript';
+
+abstract class Visitor {
+  protected context: ts.TransformationContext;
+
+  constructor(context: ts.TransformationContext) {
+    this.context = context;
+  }
+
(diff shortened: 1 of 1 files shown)


43081j:
To test this you can make it build as a commonjs module then throw up a script like so:

```ts
let source = `
@customElement('x-foo')
class XFoo extends HTMLElement {
  static get properties() { return {}; }

  @property
  public a: string;

  @property
  public b: number;

  @property
  public c: boolean;

  @property
  public d: string[];

  @property
  public e: Array<Date>;

  @property
  public f: object;

  public g: string;
}
`;
let result = ts.transpileModule(source, {
  compilerOptions: {module: ts.ModuleKind.CommonJS},
  transformers: {before: [decoratorTransformer()]}
});

console.log(result.outputText);
```

src/ts-transform.ts
```diff
@@ -0,0 +1,231 @@
...
+    stripDecorator(node, this.kind);
+
+    const args = getDecoratorArguments(matchingDecorators[0]);
+
+    if (args.length === 0) {
+      return node;
+    }
+
+    const name = args[0];
+    const defineCall = ts.createCall(
```
43081j: this is wrong at the minute. i think its either backwards or its missing the second argument!

HitkoDev:
If you make `decoratorTransformer` the default export and compile this file as a commonJS module, it can be used with [ttypescript CLI](https://github.com/cevek/ttypescript) (and most tools that allow overriding typescript compiler path)

```json
{
    "compilerOptions": {
        "plugins": [
            { "transform": "@polymer/lit-element/ts-transform.js", "type": "config" }
        ]
    }
}
```

43081j:
funnily enough i was wondering if someone had made a wrapper around the cli to inject transforms.

nice to know someone did it already!

its a potential solution then (us recommending use of ttypescript) unless we choose to use a build script or some polymer cli solution instead.

HitkoDev:
Minimal tooling is best tooling 😄 It should be simple enough to just proxy `ttsc` command through 
```sh
$ polymer-cli compile
```

(speaking of which, maybe it would be better to put transformers under https://github.com/Polymer/tools monorepo?)

That way anyone can use transformers without relying on polymer-specific build systems, something I really miss when working with Angular AOT compiler.

eterna2:
Hey @43081j

I encountered some issues using this with `ttypescript`.

If `static get properties ()` is not declared, and u are emitting type `declaration`, the compiler will throw an undefined error.

I suspect when u insert the `properties` into the ast tree, something is not updated somewhere.

The source code works properly, the error only happens when it is generating the type declarations.

Any ideas?

justinfagnani:
Hey @43081j, this is awesome, and I think we're going to get to a transform pretty soon, but I'm cleaning up issues and PRs, so I'm going to close this for now since it's not mergeable. I'll loop you in when we figure out what's up here. Thanks!

43081j:
sure, happy to pick it back up or make a new one if thats the way you wanna go :+1:

PR #152, opened 2018-08-25
Clarify `update()` signature.

Westbrook:
It is currently unclear where to manage non-rendering state change operations with the documentation deep in the explanation here rather than in the method signature.

In other places (i.e. render()), in text references to method calls seem to omit their arguments, so I did the same with update() and super.update(). Let me know if there is an alternate notation approach is preferred.


sorvell:
This was rolled into https://github.com/Polymer/lit-element/pull/161. Thanks for catching that.

PR #159, opened 2018-08-29
Add @query, @queryAll and @customElement decorators

### LINKED ISSUE #158: Consider adding a `@query` decorator for locating rendered elements
This would remove the need to find nodes in rendered output. Something like:

```
@query('.foo`)
private _fooNode: Element
```

This would likely be implemented as a getter, but we may want to have a memoization option.

justinfagnani:
Fixes #158

### DIFF
--- a/src/lib/decorators.ts
+++ b/src/lib/decorators.ts
@@ -1,0 +1,79 @@
+
+/**
+ * @license
+ * Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {LitElement} from '../lit-element.js';
+
+export type Constructor<T> = {
+  new (...args: unknown[]): T
+};
+
+/**
+ * Class decorator factory that defines the decorated class as a custom element.
+ *
+ * @param tagName the name of the custom element to define
+ *
+ * In TypeScript, the `tagName` passed to `customElement` must be a key of the
+ * `HTMLElementTagNameMap` interface. To add your element to the interface,
+ * declare the interface in this module:
+ *
+ *     @customElement('my-element')
+ *     export class MyElement extends LitElement {}
+ *
+ *     declare global {
+ *       interface HTMLElementTagNameMap {
+ *         'my-element': MyElement;
+ *       }
+ *     }
+ *
+ */
+export const customElement = (tagName: keyof HTMLElementTagNameMap) =>
+    (clazz: Constructor<HTMLElement>) => {
+      window.customElements.define(tagName, clazz);
+      // Cast as any because TS doesn't recognize the return type as being a
+      // subtype of the decorated class when clazz is typed as
+      // `Constructor<HTMLElement>` for some reason. `Constructor<HTMLElement>`
+      // is helpful to make sure the decorator is applied to elements however.
+      return clazz as any;
+    };
+
+/**
+ * A property decorator that converts a class property into a getter that
+ * executes a querySelector on the element's renderRoot.
+ */
+export const query = _query((target: NodeSelector, selector: string) =>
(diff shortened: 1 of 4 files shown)


MaKleSoft:
Would it be worthwhile to cache the result of the `querySelector` call? One could opt in or out or caching via a second decorator argument.

Also, instead of `@customElement` why not just use `@element`? 🚲🏠

src/lib/decorators.ts
```diff
@@ -0,0 +1,53 @@
...
+/*
+ * Base-implementation of `@query` and `@queryAll`.
+ *
+ * @param queryFn exectute a `selector` (ie, querySelector or querySelectorAll)
+ * against `target`.
+ */
+function _query<T>(queryFn: (target: NodeSelector, selector: string) => T) {
+  return (selector: string) => (proto: any, propName: string) => {
+    Object.defineProperty(proto, propName, {
+      get(this: HTMLElement) { return queryFn(this.shadowRoot!, selector); },
```
craigjennings11: Just wondering, but shouldn't this be `this.renderRoot` instead of `this.shadowRoot` in case someone overrides `createRenderRoot()` to return something else?

src/lib/decorators.ts
```diff
@@ -0,0 +1,53 @@
...
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+/**
+ * Class decorator that defines the decorated class a custom element.
+ *
+ * @param tagName The name of the custom element to define.
+ */
+export const customElement = (tagName: string) => (clazz: any) => {
```
kenchris: Have you considered how this will extend to scoped custom element registries?

justinfagnani: Not very thoroughly, but this is basically the same as a `customElements.define()` statement in the module. That will define on the global registry, and redefinitions of the tag name in scoped registries should be ok.

src/lib/decorators.ts
justinfagnani: Yeah...

src/lib/decorators.ts
```diff
@@ -0,0 +1,76 @@
...
+import {LitElement} from '../lit-element.js';
+
+export type Constructor<T> = {
+  new (...args: unknown[]): T
+};
+
+/**
+ * Class decorator factory that defines the decorated class as a custom element.
+ *
+ * @param tagName The name of the custom element to define.
```
sorvell: Remove capitalization and period.

src/lib/decorators.ts
```diff
@@ -0,0 +1,76 @@
...
+ *         'my-element': MyElement;
+ *       }
+ *     }
+ *
+ */
+export const customElement = (tagName: keyof HTMLElementTagNameMap) =>
+    (clazz: Constructor<HTMLElement>) => {
+      window.customElements.define(tagName, clazz);
+      // Cast as any
+      return clazz as any;
```
sorvell: Why the `any` cast?

src/test/lib/decorators_test.ts
```diff
@@ -0,0 +1,116 @@
...
+          <div>Not this one</div>
+          <div id="blah">This one</div>
+        `;
+      }
+    }
+
+    test('returns an element when it exists', async () => {
+      const c = new C();
+      container.appendChild(c);
+      await Promise.resolve();
```
sorvell: `await c.updateComplete`...
 `+` a few other places below

src/test/lib/decorators_test.ts
```diff
@@ -0,0 +1,116 @@
...
+  teardown(() => {
+    if (container !== undefined) {
+      container.parentElement!.removeChild(container);
+      (container as any) = undefined;
+    }
+  });
+
+  suite('@customElement', () => {
+    test('defines an element', () => {
+      @customElement('lit-element-test-1' as keyof HTMLElementTagNameMap)
```
sorvell: Use [generateElementName](https://github.com/Polymer/lit-element/blob/master/src/test/test-helpers.ts#L19) to ensure uniqueness of element name... in all tests that define elements.

justinfagnani: done

src/test/lib/decorators_test.ts
justinfagnani: done

src/lib/decorators.ts
justinfagnani: done

src/lib/decorators.ts
justinfagnani: done

justinfagnani:
@sorvell PTAL

PR #161, opened 2018-08-29
Update lifecycle based on feedback

sorvell:
**Adds**
* `updated`: called after the element updates and used to directly manipulate element DOM, for example focusing an element. Previously this code was written in `update` after calling `super.update`.

**Changes**
* `firstUpdated`: This was previously `firstRendered` and was changed to expose only one entry point using the "render" terminology.
* `requestUpdate`: This was previously `invalidate` and `invalidateProperty`. This one method now does double duty and was changed so we consistently use the "update" concept.
* `hasChanged`: This was previously `shouldInvalidate` and was changed to indicate meaning rather than what action is taken as a result.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,11 +176,10 @@ into the element. This is the only method that must be implemented by subclasses
 
   * `updateComplete`: Returns a Promise that resolves when the element has completed
   updating that resolves to a boolean value that is `true` if the element completed the
-  update without triggering another update. This can happen if a property is set in
-  `update()` after the call to `super.update()` for example. This getter can be
-  implemented to await additional state. For example, it is sometimes useful to
-  await a rendered element before fulfilling this promise. To do this, first
-  await `super.updateComplete` then any subsequent state.
+  update without triggering another update. This happens if a property is set in
+  `updated()`. This getter can be implemented to await additional state.
+  For example, it is sometimes useful to await a rendered element before fulfilling
+  this promise. To do this, first await `super.updateComplete` then any subsequent state.
 
   * `requestUpdate(name?, oldValue?)`: Call to request the element to asynchronously
   update regardless of whether or not any property changes are pending. This should
--- a/README.md
+++ b/README.md
@@ -161,22 +161,20 @@ into the element. This is the only method that must be implemented by subclasses
   By default, this method always returns true, but this can be customized as
   an optimization to avoid updating work when changes occur, which should not be rendered.
 
-  * `update()` (protected): This method calls `render()` and then uses `lit-html` in order to
-  render the template DOM. Implement to directly control rendered DOM.
-  Typically this is not needed as `lit-html` can be used in the `render` method
-  to set properties, attributes, and event listeners. However, it is sometimes useful
-  for calling methods on rendered elements, for example focusing an input:
-  `this.shadowRoot.querySelector('input').focus()`. The `changedProperties` argument is a Map
-  with keys for the changed properties pointing to their previous values.
-  Note, calling `super.update()` is required. Before calling `super.update()`,
-  changes made to properties do not trigger `invalidate()`, after calling `super.update()`,
-  changes do trigger `invalidate()`.
+  * `update(changedProperties)` (protected): This method calls `render()` and then uses `lit-html`
+  in order to render the template DOM. It also updates any reflected attributes based on
+  property values. Setting properties inside this method will *not* trigger the element to update.
 
-  * `firstRendered()`: (protected) Called after the element's DOM has been
+  * `firstUpdated()`: (protected) Called after the element's DOM has been
   updated the first time. This method can be useful for capturing references to rendered static
-  nodes that must be directly acted upon, for example in `update()`.
+  nodes that must be directly acted upon, for example in `updated()`. Setting properties
+  inside this method will trigger the element to update.
 
-  * `updateComplete`:  Returns a Promise that resolves when the element has completed
+  * `updated(changedProperties)`: (protected) Called whenever the element's DOM has been
+  updated and rendered. Implement to perform post updating tasks via DOM APIs, for example,
+  focusing an element. Setting properties inside this method will trigger the element to update.
+
+  * `updateComplete`: Returns a Promise that resolves when the element has completed
   updating that resolves to a boolean value that is `true` if the element completed the
   update without triggering another update. This can happen if a property is set in
   `update()` after the call to `super.update()` for example. This getter can be
@@ -184,7 +182,7 @@ into the element. This is the only method that must be implemented by subclasses
   await a rendered element before fulfilling this promise. To do this, first
   await `super.updateComplete` then any subsequent state.
 
-  * `invalidate`: Call to request the element to asynchronously update regardless
+  * `invalidate()`: Call to request the element to asynchronously update regardless
   of whether or not any property changes are pending. This should only be called
   when an element should update based on some state not stored in properties,
   since setting properties automatically calls `invalidate`.
@@ -207,14 +205,15 @@ and the property's `shouldInvalidate(value, oldValue)` returns true.
 of the event loop, before the next paint).
 * `shouldUpdate(changedProperties)`: The update proceeds if this returns true, which
 it does by default.
-* `update(changedProperties)`: Updates the element. Setting properties inside
-update is handled specially. Before calling `super.update()`, setting properties
-will *not* trigger an update. After calling `super.update()` setting properties will
-trigger an update.
+* `update(changedProperties)`: Updates the element. Setting properties inside this
+method will *not* trigger the element to update.
   * `render()`: Returns a `lit-html` TemplateResult (e.g. <code>html\`Hello ${world}\`</code>)
-  to render element DOM. Setting properties in `render()` does not trigger an update.
-  * `firstRendered()`: Called after the DOM is rendered the first time.
-  Setting properties in `firstRendered()` does trigger an update.
+  to render element DOM. Setting properties inside this method will *not* trigger
+  the element to update.
+* `firstUpdated()`: Called after the element is updated the first time.
+  Setting properties inside this method will trigger the element to update.
+* `updated()`: Called whenever the element is updated. Setting properties inside
+this method will trigger the element to update.
 * `updateComplete` promise is resolved with a boolean that is `true` if the
 element is not pending another update, and any code awaiting the element's
 `updateComplete` promise runs and observes the element in the updated state.
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ and renders declaratively using `lit-html`.
     if you're using a compiler that supports them, like [TypeScript](https://www.typescriptlang.org/) or [Babel](https://babeljs.io/).
     * With a static `properties` getter.
     * By manually writing getters and setters. This can be useful if tasks should
-    be performed when a property is set, for example validation. Call `invalidateProperty(name, oldValue)`
+    be performed when a property is set, for example validation. Call `requestUpdate(name, oldValue)`
     in the setter to trigger an update and use any configured property options.
 
     Properties can be given an options argument which is an object that describes how to
@@ -29,27 +29,27 @@ and renders declaratively using `lit-html`.
 
     Property options include:
 
-    * `attribute`: Describes how and whether the property becomes an observed attribute.
+    * `attribute`: Indicates how and whether the property becomes an observed attribute.
     If the value is `false`, the property is not added to `observedAttributes`.
     If true or absent, the lowercased property name is observed (e.g. `fooBar` becomes `foobar`).
     If a string, the string value is observed (e.g `attribute: 'foo-bar'`).
-    * `type`: Describes how to serialize and deserialize the attribute to/from a property.
+    * `type`: Indicates how to serialize and deserialize the attribute to/from a property.
     If this value is a function, it is used to deserialize the attribute value
     a the property value. If it's an object, it can have keys for `fromAttribute` and
     `toAttribute` where `fromAttribute` is the deserialize function and `toAttribute`
     is a serialize function used to set the property to an attribute. If no `toAttribute`
     function is provided and `reflect` is set to `true`, the property value is set
     directly to the attribute.
-    * `reflect`: Describes if the property should reflect to an attribute.
+    * `reflect`: Indicates if the property should reflect to an attribute.
     If `true`, when the property is set, the attribute is set using the
     attribute name determined according to the rules for the `attribute`
     propety option and the value of the property serialized using the rules from
     the `type` property option.
-    * `shouldInvalidate`: Describes if setting a property should trigger
-    invalidation and updating. This function takes the `newValue` and `oldValue` and
-    returns `true` if invalidation should occur. If not present, a strict identity
-    check is used. This is useful if a property should be considered dirty only
-    if some condition is met, like if a key of an object value changes.
+    * `hasChanged`: Indicates if a property should be considered changed when it's set.
+    This function takes the `newValue` and `oldValue` and returns `true` if an
+    update should be requested. If not present, a strict identity check is used.
+    This is useful if a property should be considered dirty only if some condition
+    is met, like if a key of an object value changes.
 
   * **React to changes:** LitElement reacts to changes in properties and attributes by
   asynchronously rendering, ensuring changes are batched. This reduces overhead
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -20,7 +20,7 @@ and renders declaratively using `lit-html`.
     if you're using a compiler that supports them, like [TypeScript](https://www.typescriptlang.org/) or [Babel](https://babeljs.io/).
```
kevinpschaaf: Suggest removing Babel reference until the included decorators support the TC39/Babel API.

README.md
```diff
@@ -29,27 +29,27 @@ and renders declaratively using `lit-html`.
...
     If `true`, when the property is set, the attribute is set using the
     attribute name determined according to the rules for the `attribute`
     propety option and the value of the property serialized using the rules from
     the `type` property option.
-    * `shouldInvalidate`: Describes if setting a property should trigger
-    invalidation and updating. This function takes the `newValue` and `oldValue` and
-    returns `true` if invalidation should occur. If not present, a strict identity
-    check is used. This is useful if a property should be considered dirty only
-    if some condition is met, like if a key of an object value changes.
+    * `hasChanged`: Indicates if a property should be considered changed when it's set.
```
kevinpschaaf: The tense here makes it sound like it is a function you call, rather than a function you supply.  

Suggestion: 
>`hasChanged`: A function that indicates if a property should be considered changed when it is set. The function should take the `newValue` and `oldValue` and return `true` if an update should be requested.

README.md
```diff
@@ -153,46 +153,41 @@ into the element. This is the only method that must be implemented by subclasses
   the `render` implementation is a [pure function](https://en.wikipedia.org/wiki/Pure_function) using only the element's current properties
   to describe the element template. This is the only method that must be implemented by subclasses.
   Note, since `render()` is called by `update()`, setting properties does not trigger
-  `invalidate()`, allowing property values to be computed and validated.
+  an update, allowing property values to be computed and validated.
 
   * `shouldUpdate(changedProperties)` (protected): Implement to control if updating and rendering
```
kevinpschaaf: I would basically consider basically all of these API's (outside of `render`) "advanced"...

src/lib/updating-element.ts
```diff
@@ -120,15 +119,15 @@ const defaultPropertyDeclaration: PropertyDeclaration = {
...
   reflect: false,
-  shouldInvalidate: notEqual
+  hasChanged: notEqual
 };
 
 const microtaskPromise = new Promise((resolve) => resolve(true));
 
 const STATE_HAS_UPDATED = 1;
-const STATE_IS_INVALID = 1 << 2;
+const STATE_IS_UPDATE_REQUESTED = 1 << 2;
```
kevinpschaaf: Nit: `STATE_UPDATE_REQUESTED`

src/lib/updating-element.ts
```diff
@@ -246,13 +245,13 @@ export abstract class UpdatingElement extends HTMLElement {
...
-   * Returns true if a property should cause invalidation and update.
-   * Called when a property value is set and uses the `shouldInvalidate`
+   * Returns true if a property should request an update.
+   * Called when a property value is set and uses the `hasChanged`
    * option for the property if present or a strict identity check.
    */
-  private static _valueShouldInvalidate(value: unknown, old: unknown,
-      shouldInvalidate: ShouldInvalidate = notEqual) {
-    return shouldInvalidate(value, old);
+  private static _valueHasChanged(value: unknown, old: unknown,
```
kevinpschaaf: Isn't the point of having this static here for overriding? Why private?

src/lib/updating-element.ts
```diff
@@ -371,7 +370,7 @@ export abstract class UpdatingElement extends HTMLElement {
     if ((this._updateState & STATE_HAS_UPDATED) && window.ShadyCSS !== undefined) {
       window.ShadyCSS.styleElement(this);
     }
-    this.invalidate();
+    this.requestUpdate();
```
kevinpschaaf: Only need on first connected

src/lib/updating-element.ts
```diff
@@ -529,22 +545,32 @@ export abstract class UpdatingElement extends HTMLElement {
   /**
    * Updates the element. By default this method reflects property values to attributes.
    * It should be overridden to render and keep updated DOM in the element's root.
-   * Within `update()` setting properties does not trigger `invalidate()`, allowing
-   * property values to be computed and validated before DOM is rendered and updated.
-   * This means in an override of `update()`, before calling `super.update()`
-   * setting properties will not trigger another update, but after calling `super.update()`
-   * setting properties will trigger another update.
+   * Setting properties inside this method will *not* trigger the element to update.
```
kevinpschaaf: ...will not trigger a *second* round of `update`

README.md
sorvell: Done

src/lib/updating-element.ts

src/lib/updating-element.ts
sorvell: Holding off on this for now until the need is verified.

src/lib/updating-element.ts

src/lib/updating-element.ts

src/lib/updating-element.ts
sorvell: Changed.

CaptainCodeman:
re: the `hasChanged` property function, I find it a bit unintuitive because my understanding is that it will run whenever the property _hasChanged_ and it's really meant to indicate if that change _should trigger an update_ ... so maybe something like `triggerUpdate` would be clearer?

README.md
justinfagnani: I believe that the current decorators are supported in Babel 7 with the `legacy `  flag. I'd link to the Babel docs: https://babeljs.io/docs/en/babel-plugin-proposal-decorators

justinfagnani:
@CaptainCodeman `hasChanged` will be called for any set of the property, even if it's the same object instance, so `hasChanged` is responsible for signaling whether the object really has any changes in it.

README.md
```diff
@@ -153,46 +151,41 @@ into the element. This is the only method that must be implemented by subclasses
...
-  should occur when property values change or `invalidate` is called. The `changedProperties`
-  argument is an object with keys for the changed properties pointing to their previous values.
+  should occur when property values change or `requestUpdate()` is called. The `changedProperties`
+  argument is a map with keys for the changed properties pointing to their previous values.
```
justinfagnani: capitalize Map, since it's a class

src/lib/updating-element.ts
```diff
@@ -529,22 +546,32 @@ export abstract class UpdatingElement extends HTMLElement {
...
       this._reflectingProperties = undefined;
     }
-    // Before this (before calling super)...
-    this._markUpdated();
+  }
+
+  /**
+   Invoked whenever the element is updated. Implement to perform
+   post updating tasks via DOM APIs, for example, focusing an element.
+   Setting properties inside this method will trigger the element to update.
```
justinfagnani: Add a newline before the last sentence, here and below. And maybe add some more detail after "to update", like "to update against this update cycle is complete"

src/lit-element.ts
```diff
@@ -32,14 +31,9 @@ export abstract class LitElement extends UpdatingElement {
...
    * Updates the element. This method reflects property values to attributes
-   * and calls `render` to render DOM via lit-html. It should be overridden to
-   * perform tasks on rendered DOM. Within `update()` setting properties does
-   * not trigger `invalidate()`, allowing property values to be computed and
-   * validated before DOM is rendered and updated. This means in an override
-   * of `update()`, before calling `super.update()` setting properties will not
-   * trigger another update, but after calling `super.update()` setting
-   * properties will trigger another update.
-   * * @param changedProperties Map of changed properties with old values
+   * and calls `render` to render DOM via lit-html. Setting properties inside
```
justinfagnani: I think these docs should more closely match UpdatingElement's version, ie. mention reflection.

src/lib/updating-element.ts
```diff
@@ -424,23 +424,33 @@ export abstract class UpdatingElement extends HTMLElement {
...
+   * configured property options are honored.
    *
-   * @param name {PropertyKey}
-   * @param oldValue {any}
+   * @param name {PropertyKey} (optional) name of requesting property
+   * @param oldValue {any} (optional) old value of requesting property
+   * @returns {Promise} A promise that is resolved when the update completes.
    */
-  protected invalidateProperty(name: PropertyKey, oldValue: any, options?: PropertyDeclaration) {
+  requestUpdate(name?: PropertyKey, oldValue?: any, options?: PropertyDeclaration) {
```
kenchris: options doesn't seem to be documented

src/lib/updating-element.ts
```diff
@@ -517,9 +534,9 @@ export abstract class UpdatingElement extends HTMLElement {
   }
 
   /**
-   * Controls whether or not `update` should be called when the element invalidates.
-   * By default, this method always returns true, but this can be customized to
-   * control when to update.
+   * Controls whether or not `update` should be called when the element requests
+   * an update. By default, this method always returns true, but this can be
```
kenchris: `true`

src/lib/updating-element.ts
```diff
@@ -529,22 +546,32 @@ export abstract class UpdatingElement extends HTMLElement {
...
       }
       this._reflectingProperties = undefined;
     }
-    // Before this (before calling super)...
-    this._markUpdated();
+  }
+
+  /**
+   Invoked whenever the element is updated. Implement to perform
+   post updating tasks via DOM APIs, for example, focusing an element.
```
kenchris: post-updating ?

README.md
```diff
@@ -153,46 +151,41 @@ into the element. This is the only method that must be implemented by subclasses
...
-
-  * `firstRendered()`: (protected) Called after the element's DOM has been
+  * `update(changedProperties)` (protected): This method calls `render()` and then uses `lit-html`
+  in order to render the template DOM. It also updates any reflected attributes based on
+  property values. Setting properties inside this method will *not* trigger another update..
+
+  * `firstUpdated()`: (protected) Called after the element's DOM has been
   updated the first time. This method can be useful for capturing references to rendered static
-  nodes that must be directly acted upon, for example in `update()`.
+  nodes that must be directly acted upon, for example in `updated()`. Setting properties
```
kenchris: ...`updated()`, which is called immediately after?

README.md
```diff
@@ -153,46 +151,41 @@ into the element. This is the only method that must be implemented by subclasses
...
+  inside this method will trigger the element to update.
+
+  * `updated(changedProperties)`: (protected) Called whenever the element's DOM has been
+  updated and rendered. Implement to perform post updating tasks via DOM APIs, for example,
+  focusing an element. Setting properties inside this method will trigger the element to update.
 
-  * `updateComplete`:  Returns a Promise that resolves when the element has completed
+  * `updateComplete`: Returns a Promise that resolves when the element has completed
   updating that resolves to a boolean value that is `true` if the element completed the
```
kenchris: double resolve here... "resolves when ... updateing that resolves" - hard to read

README.md
```diff
@@ -153,46 +151,41 @@ into the element. This is the only method that must be implemented by subclasses
...
-  * `invalidate`: Call to request the element to asynchronously update regardless
-  of whether or not any property changes are pending. This should only be called
-  when an element should update based on some state not stored in properties,
-  since setting properties automatically calls `invalidate`.
-
-  * `invalidateProperty(name, oldValue)` (protected): Triggers an invalidation for
-  a specific property. This is useful when manually implementing a property setter.
-  Call `invalidateProperty` instead of `invalidate` to ensure that any configured
-  property options are honored.
+  update without triggering another update. This happens if a property is set in
```
kenchris: in -> inside?

README.md
```diff
@@ -153,46 +151,41 @@ into the element. This is the only method that must be implemented by subclasses
...
+  `updated()`. This getter can be implemented to await additional state.
+  For example, it is sometimes useful to await a rendered element before fulfilling
+  this promise. To do this, first await `super.updateComplete` then any subsequent state.
+
+  * `requestUpdate(name?, oldValue?)`: Call to request the element to asynchronously
```
kenchris: That method seems to have a third argument `options?`

README.md
```diff
@@ -202,19 +195,20 @@ into the element. This is the only method that must be implemented by subclasses
 ## Advanced: Update Lifecycle
 
 * When the element is first connected or a property is set (e.g. `element.foo = 5`)
-and the property's `shouldInvalidate(value, oldValue)` returns true.
-* `invalidate()`: Updates the element after waiting a [microtask](https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/) (at the end
+and the property's `hasChanged(value, oldValue)` returns true.
```
kenchris: `true`

README.md
```diff
@@ -202,19 +195,20 @@ into the element. This is the only method that must be implemented by subclasses
...
+* `requestUpdate()`: Updates the element after waiting a [microtask](https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/) (at the end
```
kenchris: awaiting?

src/lib/updating-element.ts
```diff
@@ -452,62 +462,69 @@ export abstract class UpdatingElement extends HTMLElement {
...
       this._applyInstanceProperties();
     }
     if (this.shouldUpdate(this._changedProperties)) {
-      this.update(this._changedProperties);
+      const changedProperties = this._changedProperties;
+      this.update(changedProperties);
+      const needsFirstUpdate = !(this._updateState & STATE_HAS_UPDATED);
+      this._markUpdated();
+      if (needsFirstUpdate) {
+        this.firstUpdated();
```
kenchris: Any reason why firstUpdates doesn't take changedProperties as argument as well?

README.md
sorvell: Done.

README.md
sorvell: Done.

README.md
sorvell: Done.

README.md
sorvell: Fixed.

README.md
sorvell: Done.

README.md
sorvell: This argument should not be set by users. That's confusing so removing it.

README.md
sorvell: Done.

README.md
sorvell: Done.

README.md
```diff
@@ -202,19 +195,20 @@ into the element. This is the only method that must be implemented by subclasses
...
+method will *not* trigger the element to update.
   * `render()`: Returns a `lit-html` TemplateResult (e.g. <code>html\`Hello ${world}\`</code>)
-  to render element DOM. Setting properties in `render()` does not trigger an update.
-  * `firstRendered()`: Called after the DOM is rendered the first time.
-  Setting properties in `firstRendered()` does trigger an update.
+  to render element DOM. Setting properties inside this method will *not* trigger
+  the element to update.
+* `firstUpdated()`: Called after the element is updated the first time.
+  Setting properties inside this method will trigger the element to update.
+* `updated()`: Called whenever the element is updated. Setting properties inside
```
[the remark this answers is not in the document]
sorvell: Done.

src/lib/updating-element.ts
```diff
@@ -424,23 +424,33 @@ export abstract class UpdatingElement extends HTMLElement {
...
   /**
-   * Triggers an invalidation and records an old value for the specified
-   * property to be presented in the `changedProperties` argument to `update`.
-   * When manually creating a property setter, this method should be called to
-   * trigger an invalidation that honors any of the property options specified
-   * for the given property.
+   * Requests an update which is processed asynchronously. This should
+   * only be called when an element should update based on some state not triggered
+   * by setting a property or when manually implementing a property
+   * setter. In this case, pass the property `name` and `oldValue` to ensure that any
```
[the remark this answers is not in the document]
sorvell: Fixed.

src/lib/updating-element.ts
sorvell: Removed.

src/lib/updating-element.ts
sorvell: Added.

src/lib/updating-element.ts

src/lib/updating-element.ts

src/lib/updating-element.ts
sorvell: Added.

src/lit-element.ts
sorvell: Already mentioned, but made more similar.

sorvell:
Tests failing on old Chrome versions due to https://github.com/Polymer/lit-html/issues/462.

PR #167, opened 2018-08-31
Clarify a few things in the documentation

### LINKED ISSUE #166: Clarify what 'property value is set directly to the attribute' means
> If no toAttribute function is provided and reflect is set to true, the property value is set directly to the attribute.

I assume this will call the `toString()` by default.

Testing with 
`$0.setAttribute("hi", new Object)` yields `[object Object]`
and 
`$0.setAttribute("hi", true)` yields `true`

But it might also throw (ie. if I make a mistake like below):

image

`$0.setAttribute("hi", new Object({ toString: () => "hi" }))` becomes `hi` (string)

But apart from that, you do have special handling of Boolean, which should probably be mentioned here as well:

```
    const toAttribute = options.type === Boolean ? toBooleanAttribute :
        (options.type && (options.type as AttributeSerializer).toAttribute || String);
```

kenchris:
Fixes #166

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,31 +23,32 @@ and renders declaratively using `lit-html`.
     be performed when a property is set, for example validation. Call `requestUpdate(name, oldValue)`
     in the setter to trigger an update and use any configured property options.
 
-    Properties can be given an options argument which is an object that describes how to
+    Properties can be given an `options` argument which is an object that describes how to
     process the property. This can be done either in the `@property({...})` decorator or in the
     object returned from the `properties` getter, e.g. `static get properties { return { foo: {...} }`.
 
     Property options include:
 
     * `attribute`: Indicates how and whether the property becomes an observed attribute.
-    If the value is `false`, the property is not added to `observedAttributes`.
-    If true or absent, the lowercased property name is observed (e.g. `fooBar` becomes `foobar`).
+    If the value is `false`, the property is not added to the static `observedAttributes` getter.
+    If `true` or absent, the lowercased property name is observed (e.g. `fooBar` becomes `foobar`).
     If a string, the string value is observed (e.g `attribute: 'foo-bar'`).
     * `type`: Indicates how to serialize and deserialize the attribute to/from a property.
-    If this value is a function, it is used to deserialize the attribute value
-    a the property value. If it's an object, it can have keys for `fromAttribute` and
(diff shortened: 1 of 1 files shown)


kenchris:
@sorvell rebased

README.md
```diff
@@ -23,31 +23,35 @@ and renders declaratively using `lit-html`.
...
-    If `true`, when the property is set, the attribute is set using the
-    attribute name determined according to the rules for the `attribute`
-    property option and the value of the property serialized using the rules from
-    the `type` property option.
-    * `hasChanged`: A function that indicates if a property should be considered
-    changed when it is set. The function should take the `newValue` and `oldValue`
-    and return `true` if an update should be requested.
+    function is provided and `reflect` is set to `true`, the attribute value is
+    set directly using the property value, thus using the property value's associated
+    `toString` method. In the case that is missing, serialization will throw.
```
jrunning: I don't know if it's even worth mentioning, but this isn't _quite_ correct. This is [what's happening in the source](https://github.com/Polymer/lit-element/blob/master/src/lib/updating-element.ts#L283-L285):

```js
const toAttribute = options.type === Boolean ? toBooleanAttribute :
    (options.type && (options.type as AttributeSerializer).toAttribute || String);
return toAttribute(value);
```

In other words, if `toAttribute` isn't supplied, it doesn't call `value.toString()`; it calls `String(value)`. This is a distinction mostly without meaning, since [`String` calls `arg.toString`](http://www.ecma-international.org/ecma-262/6.0/#sec-string-constructor-string-value). The only scenario I can think of in which the behavior would be different is if you redefined `window.String`.

kenchris:
Ping @sorvell @justinfagnani

README.md
justinfagnani: I think we can just say that `type` defaults to `String`

README.md
```diff
@@ -23,31 +23,35 @@ and renders declaratively using `lit-html`.
...
+    * `reflect`: Indicates whether setting this property will be reflected in the
```
justinfagnani: nit:

> whether setting this property will be reflected

That reads a little weird to me. I like the original wording, but maybe with the detail added here:

"whether the property should reflect to its associated attribute (as determined by the `attribute` option)"

README.md
kenchris: But that is not really the case as it only uses type if it had a toAttribute property, or else it uses string

PR #184, opened 2018-09-12
Export decorators

sorvell:


src/lib/updating-element.ts
```diff
@@ -31,6 +31,8 @@ interface AttributeSerializer<T = any> {
 
 type AttributeType<T = any> = AttributeSerializer<T>|((value: string) => T);
 
+export * from './decorators.js';
```
justinfagnani: This creates a cycle. Exporting from lit-element.js should break it.

PR #187, opened 2018-09-13
Remove Polymer custom typings

### LINKED ISSUE #186: custom typings for polymer can be removed
Since `LitElement` is not based on polymer property bindings anymore it is save to remove these typings, I guess.

abdonrd:
### Reference Issue

Fixes #186


hastebrot:
Hi @abdonrd. I think there is also a reference to the typings in `package.json`

abdonrd:
@hastebrot what typings do you mean?

I only find:

https://github.com/Polymer/lit-element/blob/b8c53be92fa867d280fde1714d365dc19917910e/package.json#L40

But this is about the lit-element itself.

hastebrot:
Oh sorry. You're right.

PR #203, opened 2018-09-17
Remove custom_typings

### LINKED ISSUE #186: custom typings for polymer can be removed

sorvell:
Fixes #186

Supersedes and closes #187


web-padawan:
Same as #187 😃

PR #206, opened 2018-09-18
fixed decorators to return a property descriptor to work with babel d…

### LINKED ISSUE #205: lit decorators doesnt work with babel legacy mode decorators
babel decorators (in legacy mode) expect the property descriptor to be returned rather than be defined within the decorator, TS also supports this feature.

#### Steps to Reproduce
checkout https://github.com/mzeiher/lit-element-test

npm install 
npm run build
open index.html

#### Expected Results
both elements are updated (one decorated with TS the other with babel decorators in legacy mode)

#### Actual Results
only the TS version works

### Browsers Affected
- [ x] Chrome
- [ x] Firefox
- [ x] Edge
- [ x] Safari 11
- [ x] Safari 10
- [ x] IE 11

### Versions
- lit-element: v0.6.1
- webcomponents: v2.1.3

mzeiher:
…ecorators

### Reference Issue
Fixes #205

### DIFF
--- a/src/lib/decorators.ts
+++ b/src/lib/decorators.ts
@@ -21,6 +21,11 @@ export type Constructor<T> = {
   new (...args: unknown[]): T
 };
 
+export type FixedPropertyDecorator = (target: Object, propertyKey: string | symbol, descriptor?: BabelPropertyDescriptor) => any;
+export interface BabelPropertyDescriptor extends PropertyDescriptor {
+    initializer: () => any;
+}
+
 /**
  * Class decorator factory that defines the decorated class as a custom element.
  *
@@ -55,9 +60,9 @@ export const customElement = (tagName: keyof HTMLElementTagNameMap) =>
  * corresponding attribute value. A `PropertyDeclaration` may optionally be
  * supplied to configure property features.
  */
-export const property = (options?: PropertyDeclaration) => (proto: Object,
-                                                            name: string) => {
-  (proto.constructor as typeof UpdatingElement).createProperty(name, options);
+export const property = (options?: PropertyDeclaration):FixedPropertyDecorator => (proto: Object,
+                                                            name: string | symbol, descriptor?:BabelPropertyDescriptor) => {
+  return (proto.constructor as typeof UpdatingElement).createProperty(name, options, descriptor);
 };
 
 /**
--- a/src/lib/decorators.ts
+++ b/src/lib/decorators.ts
@@ -60,8 +60,8 @@ export const customElement = (tagName: keyof HTMLElementTagNameMap) =>
  * corresponding attribute value. A `PropertyDeclaration` may optionally be
  * supplied to configure property features.
  */
-export const property = (options?: PropertyDeclaration):FixedPropertyDecorator => (proto: Object,
-                                                            name: string | symbol, descriptor?:BabelPropertyDescriptor) => {
+export const property = (options?: PropertyDeclaration): FixedPropertyDecorator => (proto: Object,
+                                                            name: string | symbol, descriptor?: BabelPropertyDescriptor) => {
   return (proto.constructor as typeof UpdatingElement).createProperty(name, options, descriptor);
 };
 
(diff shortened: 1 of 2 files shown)


mzeiher:
I signed it!

aadamsx:
Can't add decorators to our classes right now using Bable7.  Does this fix that issue?

src/lib/decorators.ts
```diff
@@ -55,9 +60,9 @@ export const customElement = (tagName: keyof HTMLElementTagNameMap) =>
  * corresponding attribute value. A `PropertyDeclaration` may optionally be
  * supplied to configure property features.
  */
-export const property = (options?: PropertyDeclaration) => (proto: Object,
-                                                            name: string) => {
-  (proto.constructor as typeof UpdatingElement).createProperty(name, options);
+export const property = (options?: PropertyDeclaration): FixedPropertyDecorator => (proto: Object,
```
43081j: this should return a `PropertyDescriptor | null` i think

43081j: to satisfy typescript (temporarily) and be consistent with how other decorators are written, this should probably return `any`.

src/lib/updating-element.ts
```diff
@@ -187,19 +190,29 @@ export abstract class UpdatingElement extends HTMLElement {
...
-      return;
+      return null;
     }
+
     const key = typeof name === 'symbol' ? Symbol() : `__${name}`;
-    Object.defineProperty(this.prototype, name, {
-      get() { return this[key]; },
-      set(value) {
+
+    let defaultInitializer: null | (() => any) = descriptor ? descriptor!.initializer || null : null;
```
43081j: This whole handling of the default value seems just a little messy but i can see why...

The default value **really** needs to be in the constructor but babel annoyingly doesn't put it there...

mzeiher:
@43081j I found away to support all, stage-0 babel, stage-0 typescript and stage-2 babel decorators while experimenting in my own decorator project: https://github.com/mzeiher/ce-decorators/blob/master/src/stage2decorators.ts, https://github.com/mzeiher/ce-decorators/blob/master/src/prop.ts, should in include such a thing in the PR to be future proof?

43081j:
I'll have a read soon.

One thing to keep in mind is that the legacy decorators babel produces are neither stage 0 nor stage 2, they are an incorrectly patched stage-0 by the looks of it:

https://github.com/babel/babel/blob/5150f5f5003b25d89f8b5041543412839cfde5fd/packages/babel-helpers/src/helpers.js#L999

They make use of `PropertyDescriptor#initializer` which doesn't seem to be a thing. New decorators (stage-2) do have an `initializer` but **not on the descriptor** (rather on the "class element descriptor", a new interface introduced by stage-2 decorators which wraps the property descriptor).

mzeiher:
hehe that's the fun when using experimental features :)
that's the how the implementation works at the moment (still in development):
Stage-0 and Stage-0 babel case:
member:
if decorating a member which is a "value" property:
define "hidden" value-property with undefined as value (for ts and initializer for babel) on prototype (es extra in stage-2 case)
return get-set property with a delegate to the created "hidden" value property and then handles reflection/sync/re-render
if  decorating a member which has get/set (case e.g. @prop() get whatever() {} or when another decorator already patched the property):
return a new get-set property which delegates to the previous get/set (in this case there is no initializer in babel case)  and then handles reflection/sync/re-render

the same is done for stage-2 decorators, this way multiple decorators are possible which manipulates the property every decorator delegates to get/set to the previous get/set

for method decorators it's just the get/set case is used but the delegate goes to the value of the previous PropertyDescriptor which holds the original function or a overidden one from another decorator (value = function(...parameter) { original.value.apply(this, [...parameter] }

the delegation logic must be implemented from every decorator which patches a member or member-function so the one implementing the decorator must take care of that (if someone mixes the decorators created by you and for example own logging decorators), I'll add an example and a test for it...

mzeiher:
cosing... implemented by the lit-element team

PR #209, opened 2018-09-19
Fix customElement decorator

### LINKED ISSUE #207: customElement decorator bug
### Description

If I import a file which use customElement decorator serveral times,there will be a "Failed to execute 'define' on 'CustomElementRegistry': this name has already been used with this registry" error
#### Live Demo

There is a example in my project's demo (https://github.com/fsx950223/chartjs-web-components/tree/master/demo).
#### Steps to Reproduce

1. Define element in base.js
 ``` js
import {LitElement,html,property,customElement} from '@polymer/lit-element'
@customElement('base-chart' as keyof HTMLElementTagNameMap)
export default class BaseChart extends LitElement{}
```
2. Import base.js in a file
``` js
import './base'
```
3. Import base.js in the other file
``` js
import './base'
```
4. They have different entries and use webpack bundle them
#### Expected Results

js exec normally
#### Actual Results

Failed to execute 'define' on 'CustomElementRegistry': this name has already been used with this registry
### Browsers Affected

- [*] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari 11
- [ ] Safari 10
- [ ] IE 11

### Versions

- lit-element: v0.6.1
- webcomponents: vX.X.X

fsx950223:
### Reference Issue

Fixes #207

### DIFF
--- a/src/lib/decorators.ts
+++ b/src/lib/decorators.ts
@@ -42,7 +42,9 @@ export type Constructor<T> = {
  */
 export const customElement = (tagName: keyof HTMLElementTagNameMap) =>
     (clazz: Constructor<HTMLElement>) => {
-      window.customElements.define(tagName, clazz);
+      if (!window.customElements.get(tagName)) {
+        window.customElements.define(tagName, clazz);
+      }
       // Cast as any because TS doesn't recognize the return type as being a
       // subtype of the decorated class when clazz is typed as
       // `Constructor<HTMLElement>` for some reason. `Constructor<HTMLElement>`


LarsDenBakker:
You are catching the error and throwing a new one. So there is no functional change?

fsx950223:
> You are catching the error and throwing a new one. So there is no functional change?

Yes

mkay581:
The idea is that it _should_ break your javascript because the same custom element shouldn't be defined more than once. If you want to avoid this error, you should use a build tool that bundles your code so that the custom element definition only happens once in your bundled file, if the build library you're using doesn't already do it for you.

PR #213, opened 2018-09-19
Add disconnectedCallback stub.

### LINKED ISSUE #212: super.disconnectedCallback() is no longer present
### Description
`super.disconnectedCallback()` having been removed from `LitElement` puts pressure on the element developer to make a clear decision as to how to address the possible presence of this method at various extension and mixin levels. Further, it makes for the possibility that there would need to be a future breaking change in `LitElement` where core functionality needed in this callback.

#### Live Demo
https://glitch.com/edit/#!/five-latex?path=index.html:39:50

#### Steps to Reproduce
1. Create an element
2. Add a `disconnectedCallback` method.
3. Call `super.disconnectedCallback()` in it.
4. Disconnect the element and see the error.

#### Expected Results
No error. Or clear documentation of the approach to this.

#### Actual Results
Uncaught TypeError: (intermediate value).disconnectedCallback is not a function
    at HTMLElement.disconnectedCallback ((index):25)
    at (index):36

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari 11
- [x] Safari 10
- [x] IE 11

### Versions

- lit-element: vX.X.X
- webcomponents: vX.X.X

Westbrook:
Fixes #212

Disconnected stub allows for feature additions in the disconnect phase of the element lifecycle as needed without it being a breaking change, while making it possible to apply `super.disconnectedCallback()` in all extensions.

### DIFF
--- a/src/lib/updating-element.ts
+++ b/src/lib/updating-element.ts
@@ -397,6 +397,13 @@ export abstract class UpdatingElement extends HTMLElement {
   }
 
   /**
+   * Allows for `super.disconnectedCallback()` in extensions while
+   * reserving the positibility of making non-breaking feature additions
+   * when disconnecting at some point in the future.
+   */
+  disconnectedCallback() {}
+
+  /**
    * Synchronizes property values when attributes change.
    */
   attributeChangedCallback(name: string, old: string, value: string) {


Westbrook:
Upon reflection and review of https://github.com/Polymer/polymer/blob/master/lib/mixins/properties-mixin.js, I wonder if this would be better applied in https://github.com/Polymer/lit-element/blob/master/src/lit-element.ts with the same extension protection syntax as properites-mixin.js did?

mkay581:
@Westbrook this is a very nice catch. I was wondering about this the other day.

PR #216, opened 2018-09-20
Small changes on README: Adding boolean attribute declarative syntax, removing some useless ""

LostInBrittany:
Small improvements to the README


LostInBrittany:
I know, it's rather bothering, but that's a plugin problem, let's issue there ;)

Or is we need to use `""` for the moment, let's use them in all the examples :)

In any case, I like the idea of clarifying that both ways work and linking to specs as this works like HTML.

katejeffreys:
Obsoleted by recent changes to README

PR #225, opened 2018-09-27
removed property from listed import in examples as that is not valid …

btopro:
Implementing the doc from the doc page leads to:

The requested module '../../@polymer/lit-element/lit-element.js' does not provide an export named 'property'

Remove property from exports in the docs until it's available (if it's planned in the future)


web-padawan:
Duplicate of #220 and please note that `property` is a decorator.

btopro:
oops mybad, so it is

Issue #237, opened 2018-10-03
Add an `eventListenerOptions` decorator


justinfagnani:
https://github.com/Polymer/lit-html/pull/531 adds support for declarative event listener options. We should support this with a decorator:

```ts
class MyElement extends LitElement {
  render() {
    return html`<button @click=${this.onClick}>Click Me</button>`;
  }

  @eventListenerOptions({capture: true})
  onClick(e) {
    console.log('clicked');
  }
}
```

We should also document that without decorators, users can simply set an attribute on the method directly:

```ts
MyElement.prototype.capture = true;
```

justinfagnani:
@jolleekin the override will need its own options. We're not going to read from up the prototype chain.

aadamsx:
@justinfagnani what do you mean by this?

> We're not going to read from up the prototype chain.

frankiefu:
Fixed by https://github.com/Polymer/lit-element/pull/244

PR #244, opened 2018-10-05
Events

### LINKED ISSUE #236: Call lit-html render() with `this` as the eventContext
https://github.com/Polymer/lit-html/pull/523 added support for an `eventContext` option to `render()`. If we set it to `this`, then event listeners will not have to be bound to have the correct `this` reference when invoked.

justinfagnani:
Fixes #236 and #237 

This lets users add methods as declarative event handlers without having to bind the method or use an arrow function. The decorator allows options to be set on the method for capture, passing and once.

```ts
  render() {
    return html`<button @click=${this.onClick}>Click Me</button>`;
  }

  @eventOptions({capture: true})
  onClick(e) {
    console.log('clicked');
  }
}
```

This PR relies on https://github.com/Polymer/lit-html/pull/531 and tests will fail till that is merged and released, though tests pass locally.


justinfagnani:
cc @katejeffreys and @arthurevans for docs.

README.md
```diff
@@ -64,7 +64,7 @@ and renders declaratively using `lit-html`.
     * expression: ``` html`<div>${disabled ? 'Off' : 'On'}</div>` ```
     * property: ``` html`<x-foo .bar="${bar}"></x-foo>` ```
     * attribute: ``` html`<div class="${color} special"></div>` ```
-    * event handler: ``` html`<button @click="${(e) => this._clickHandler(e)}"></button>` ```
+    * event handler: ``` html`<button @click="${this._clickHandler}"></button>` ```
```
kevinpschaaf: Worth mentioning (here or elsewhere) that listener is bound to element (since that's not standard `lit-html`)?

Also, worth showing a non-decorator pattern for options?

aadamsx: Please show/add as many explanations and examples as possible on things like this! :)

thepassle:
Seconded

README.md
justinfagnani: The README is about to be replaced with real docs.... It already didn't say how to set up event handlers correctly, so I think this at least isn't any worse.

justinfagnani:
@kevinpschaaf PTAL.

I didn't update the README yet because honestly I'm not sure where to put it, the README's far too long already, and it's about to be replaced.

kevinpschaaf:
Hah, oh well.  I was going to suggest to modify the "declarative rendering" section along these lines:

* event handler: ``` html`<button @click="${this._clickHandler}"></button>` ```, where the value is either
  * a function (to be called with the element as context), or
  * an object with a `handleEvent` property (called with the element as context), and
  * in either case, any [additional addEventListener options](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Parameters) such as `capture: true` may be included on the function or object to control the event listener

PR #251, opened 2018-10-06
Fix import path in minimal example

bgotink:
Import paths must be absolute URLs or start with `/`, `./` or `../`.

Note that the minimal example still doesn't work because lit-element imports from `'lit-html'`, which won't work, but at least the code example looks right now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ into the element.
 ```html
   <script src="node_modules/@webcomponents/webcomponentsjs/webcomponents-bundle.js"></script>
   <script type="module">
-    import {LitElement, html} from '@polymer/lit-element';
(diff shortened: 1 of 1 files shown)


keanulee:
You're correct that package names don't work in the browser. However, we've decided to vend our libraries using package names for convenience and to align with the Node.js community. For more details, see [this blog post](https://www.polymer-project.org/blog/2018-02-26-3.0-preview-paths-and-names).

Many build tools, including webpack/rollup/[polymer build/polymer serve](https://www.polymer-project.org/3.0/toolbox/build-for-production#transforms), as well as services like unpkg.com (with the `?module` query string), will handle package names. The code samples in the README assumes  this.

PR #255, opened 2018-10-08
add missing import `repeat` directive

SerkanSipahi:


keanulee:
The `repeat` directive isn't used in this example - `string.repeat` is being used.

PR #259, opened 2018-10-11
Added reference to lit-html guide for how to create templates

ErikHellman:
I was trying to find the documentation for the lit-html API for how to write templates but had some problems finding it. I suspect I'm not alone, so I added a sentence and a link to the lit-html guide to make it easier to find.


keanulee:
Thanks

PR #277, opened 2018-10-25
Updated README.ms  regarding hasChanged and property setters

ErikHellman:
I realised I was using hasChanged in the wrong way and thought that a more explicit documentation on both setter and the property option would be useful

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,9 @@ for additional information on how to create templates for lit-element.
     * With a static `properties` getter.
     * By manually writing getters and setters. This can be useful if tasks should
     be performed when a property is set, for example validation. Call `requestUpdate(name, oldValue)`
-    in the setter to trigger an update and use any configured property options.
+    in the setter to trigger an update and use any configured property options. If you want to perform 
+    validation of the data before it is set, see how to use `hasChanged` as described in the property 
+    options below.
 
     Properties can be given an `options` argument which is an object that describes how to
     process the property. This can be done either in the `@property({...})` decorator or in the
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -50,7 +52,9 @@ for additional information on how to create templates for lit-element.
     in the absence of the attribute.
     * `hasChanged`: A function that indicates whether a property should be considered
     changed when it is set and thus result in an update. The function should take the
-    `newValue` and `oldValue` and return `true` if an update should be requested.
+    `newValue` and `oldValue` and return `true` if an update should be requested. This 
+    is mostly useful to perform validation before setting the value. To perform an action
```
kenchris: Please check advanced section

> A property is set (e.g. element.foo = 5).
> If the property's hasChanged(value, oldValue) returns false, the element does not update. If it returns true, requestUpdate() is called to schedule an update.

hasChanged is for when you don't manually do your own properties (but define them via the static properties getter).

It is basically a /bring your own dirty checker/ for whether to call requestUpdate or not.

Now, if you do your own property manually (you should still define them via the static properties getter for the serialization and attribute reflection), then hasChanged is not called and you have to manually decide whether to call requestUpdate or not, given your validation code, which might not even change the property value.

```js
set age(value) { 
  if (value < 0 || value > 120 ) { return; // invalid age, don't update property and don't re-render }
  this.requestUpdate("age");
}
```

kenchris:
It's README.md not .*ms*

README.md
ErikHellman: Ah, got it. I'll think of how to phrase this better and will write a new PR :)

ErikHellman:
Will make a new PR when I got a better text to suggest. :)

Issue #291, opened 2018-11-05
customElement decorator should use type string for `tagName`


sorvell:
This is currently `keyof HTMLElementTagNameMap` but this doesn't work nicely in TypeScript right now and the result is you have to cast to `any`. Instead we should just use `string` for now.

jhpratt:
Seconding this. I just ran into this issue while testing out lit element on an existing component.

jhpratt:
@nrdobie That's because that long list _is_ what `keyof HTMLElementTagNameMap` is equivalent to. That's not a wrong setting at all; lit-element currently requires you to manually extend the interface. I don't think that should be necessary, and switching to `string` would suffice (rather than casting to `any`).

legowerewolf:
Just ran into this, would be a nice fix. @nrdobie how would changing to `string` change the behavior?

nrdobie:
Functionally it won't but it doesn't require the HTML tag list to be
updated for TypeScript. I know I'm splitting hairs and I don't really have
a strong opinion. I can union this with string and get the best of both.

vladnicula:
Can someone explain the reasoning for having keyof there? I don 't fully understand why this wasn't a string in the first place.

sorvell:
Fixed via https://github.com/Polymer/lit-element/pull/292.

PR #292, opened 2018-11-05
Update decorator types

### LINKED ISSUE #288: property() decorator should work with Symbol-named properties
Like so:

```ts
const protectedField = Symbol();
  @property()
  [protectedField] = 0;
}
```

sorvell:
Fixes #288 and #291


jhpratt:
@kevinpschaaf Any plans to merge this?

PR #303, opened 2018-11-12
Allow @query and @queryAll to memoize the results

jolleekin:
This pull request adds a second parameter to `@query` and `@queryAll`, allowing them to memorize the results for better performance.

Examples
``` ts
@query(selector: string, memoize = false)

@query('#child')
child!: Element;

@query('#child', true)
child!: Element;

/***********************************************/

@queryAll(selector: string, memoize = false)

@queryAll('.item')
items!: NodeListOf<Element>;

@queryAll('.item', true)
items!: NodeListOf<Element>;
 
```

### DIFF
--- a/src/lib/decorators.ts
+++ b/src/lib/decorators.ts
@@ -63,6 +63,17 @@ export const property = (options?: PropertyDeclaration) => (proto: Object,
 /**
  * A property decorator that converts a class property into a getter that
  * executes a querySelector on the element's renderRoot.
+ *
+ * __Usage__
+ * ``` ts
+ * @query(selector: string, memoize = false)
+ *
+ * @query('#child')
+ * child!: Element;
+ *
+ * @query('#child', true)
+ * child!: Element;
+ * ```
  */
 export const query = _query((target: NodeSelector, selector: string) =>
                                 target.querySelector(selector));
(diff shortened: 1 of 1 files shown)


jolleekin:
Any plan to merge this pull request? All of my use cases only require `@query` to be executed once.

CaptainCodeman:
Instead of adding a parameter specifically to memoize this one case, you should be able to use a generic memoization decorator that could then be re-used for anything.

justinfagnani:
I think this is a nice option to include because it can be a decent perf win in some cases and we don't document a separate thing.

src/lib/decorators.ts
```diff
@@ -63,13 +63,35 @@ export const property = (options?: PropertyDeclaration) => (proto: Object,
...
 export const query = _query((target: NodeSelector, selector: string) =>
                                 target.querySelector(selector));
 
 /**
  * A property decorator that converts a class property into a getter
  * that executes a querySelectorAll on the element's renderRoot.
+ *
+ * __Usage__
+ * ``` ts
+ * @queryAll(selector: string, memoize = false)
```
justinfagnani: This decorator isn't decorating anything

src/lib/decorators.ts
```diff
@@ -63,13 +63,35 @@ export const property = (options?: PropertyDeclaration) => (proto: Object,
 /**
  * A property decorator that converts a class property into a getter that
  * executes a querySelector on the element's renderRoot.
+ *
+ * __Usage__
```
justinfagnani: We don't have `__Usage__` in any other jsdocs here. We should use the `@example` tag: http://usejsdoc.org/tags-example.html

justinfagnani:
@jolleekin any reason you closed this?

jolleekin:
I think we should use @CaptainCodeman 's approach. In fact, that's what I'm doing now.

PR #306, opened 2018-11-13
Docs redesign

keanulee:
https://lit-element.polymer-project.org


docs/tools/develop.md
```diff
@@ -0,0 +1,79 @@
...
+
+    To view the served project in the example above, open http://127.0.0.1:8081.
+
+<a id="ide">
+
+### [Text editor](#ide)
+
+You will need a text editor that does HTML syntax highlighting inside JavaScript template literals. 
+
+* [VSCode](https://code.visualstudio.com/#alt-downloads) with [https://marketplace.visualstudio.com/items?itemName=bierner.lit-html#qna](Matt Bierner's lit-html plugin)
```
Donskelle: Matt Bierner lit-html plugin needs to be swapped with url

Donskelle: Ohh i think its an issue now 😟

keanulee: Fixed in #308 and deployed. Thanks for pointing it out!

PR #307, opened 2018-11-14
Fully support Closure compiler including ADVANCED_OPTIMIZATIONS

rickcnagy:
Currently, compiling lit-element with the Closure compiler results in a
clean build but a number of hard JS errors due to a few issues:
- static methods being collapsed (i.e. removed entirely)
- Object.hasOwnPropertyName() looking for static properties that have
been rewritten by the compiler
- static property initializers not being run when @exported

The last issue is a compiler bug being tracked internally, but all three
need to be worked around before this will compile and run cleanly.

This includes all of the necessary fixes, including addressing the last
issue by moving the Object.hasOwnPropertyName('_classProperties') check
to _finalize(). This is necessary to ensure that _classProperties is
defined before it's iterated in observedProperties on an element that
has no properties.

Another option for making _finalized and _classProperties public would
be to move to checking this._finalized and this._classProperties and not
needing to access them by name - I thought that was out of scope for
this but would be happy to change to that style if reviewers prefer.

### DIFF
--- a/src/lib/updating-element.ts
+++ b/src/lib/updating-element.ts
@@ -132,23 +132,32 @@ export abstract class UpdatingElement extends HTMLElement {
   /**
    * Maps attribute names to properties; for example `foobar` attribute
    * to `fooBar` property.
+   * @nocollapse
    */
   private static _attributeToPropertyMap: AttributeMap = new Map();
 
   /**
-   * Marks class as having finished creating properties.
+   * Marks class as having finished creating properties. This is public only to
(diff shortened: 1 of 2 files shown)


rictic:
The most surprising change is the static property initializers. Apparently static fields don't consistently inherit after optimization in closure compiler

sorvell:
We're going to need to slightly change the way `_classProperties` is done to ensure that `decorators` work correctly. This is a little tricky to explain in review, so going to merge this to a branch and make the changes from there.

Issue #309, opened 2018-11-15
LitElement.render has return type void, why not TemplateResult?


mikeleonard:
### Description
In typescript you can currently extend LitElement with a class whose render method has a different type signature. For example this doesn't produce any typechecking errors:

```typescript
export class MyElement extends LitElement {
    render() {
        return "something other than TemplateResult";
    }
}
```

#### Expected Results

I would expect to see the following ts error:
```typescript
[ts]
Property 'render' in type 'MyElement' is not assignable to the same property in base type 'LitElement'.
  Type '() => string' is not assignable to type '() => TemplateResult'.
    Type 'string' is not assignable to type 'TemplateResult'. [2416]
(method) MyElement.render(): string
```

#### Proposal

In the LitElement class, change `protected render(): void` to `protected render(): TemplateResult`
```typescript
export declare class LitElement extends UpdatingElement {
    /**
     * Invoked on each update to perform rendering tasks. This method must return
     * a lit-html TemplateResult. Setting properties inside this method will *not*
     * trigger the element to update.
     * @returns {TemplateResult} Must return a lit-html TemplateResult.
     */
    protected render(): TemplateResult;
}
```

The type signature was previously `abstract protected render(): TemplateResult` and was changed to `protected render(): void` in this pull request: https://github.com/Polymer/lit-element/pull/228/files. 

This was done to change it from an abstract class to a normal class in order to make the mixin pattern work better. I agree with getting rid of abstract, but I'm not sure why the return type was changed to `void` - could it be changed back to TemplateResult? 

If there's a reason for having the return type as `void` then no problem at all, just wanted to ask the question before you hit 1.0.

mikeleonard:
Also the `updateComplete` promise returned by `requestUpdate` has the type `Promise<unknown>`... however the documentation says it returns a promise of a boolean. 

It would be I think useful for typescript users to change this to boolean. Thoughts?
```javascript
     /**
     * Returns a Promise that resolves when the element has completed updating.
     * The Promise value is a boolean that is `true` if the element completed the
     * update without triggering another update. The Promise result is `false` if
     * a property was set inside `updated()`. This getter can be implemented to
     * await additional state. For example, it is sometimes useful to await a
     * rendered element before fulfilling this Promise. To do this, first await
     * `super.updateComplete` then any subsequent state.
     *
     * @returns {Promise} The Promise returns a boolean that indicates if the
     * update resolved without triggering another update.
     */
[...]

PR #316, opened 2018-11-27
Explicitly annotate render's return type

rictic:
This way, an overriding class will get a type error if their render function returns something that isn't a TemplateResult.

### DIFF
--- a/src/lit-element.ts
+++ b/src/lit-element.ts
@@ -51,7 +51,6 @@ export class LitElement extends UpdatingElement {
    * Invoked on each update to perform rendering tasks. This method must return
    * a lit-html TemplateResult. Setting properties inside this method will *not*
    * trigger the element to update.
-   * @returns {TemplateResult} Must return a lit-html TemplateResult.
    */
   protected render(): TemplateResult|void {}
 }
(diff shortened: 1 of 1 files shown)


web-padawan:
Fixes #309

src/lit-element.ts
```diff
@@ -53,5 +53,5 @@ export class LitElement extends UpdatingElement {
    * trigger the element to update.
    * @returns {TemplateResult} Must return a lit-html TemplateResult.
    */
-  protected render() {}
+  protected render(): TemplateResult|void {}
```
justinfagnani: I'd remove the `|void`. render() is supposed to be side-effect free, so `void` wouldn't make any sense.

justinfagnani: Shouldn't we remove the type annotation from the comment?

rictic: Yes, except that there's a non-abstract implementation of it right here, and that version returns `void`

rictic: +1

justinfagnani: I wish we could implement an abstract method just so it won't throw...

evmar: Why not make it abstract?  If that's not an option, what should it return in the cases where people extend this class but fail to implement render()?  A final option is to just `throw 'notreached'` in the impl and drop the void.

PR #326, opened 2018-11-30
Allow passive event option to be set

### LINKED ISSUE #325: @eventOptions decorator doesn't allow passive to be set
The current `@eventOptions` decorator doesn't allow passive to be set so the following will fail to compile.

```
@eventOptions({passive: true})
private _handleInteraction(e: Event) {
    this.mdcFoundation.handleInteraction(e);
}
```

#### Live Demo
Attempt to build the [material-components-web-components](https://github.com/material-components/material-components-web-components) project.

#### Steps to Reproduce
Follow instructions in above repo to build locally

#### Expected Results
Code compiles

#### Actual Results
Code does not compile with errors such as:

```
packages/tab-scroller/src/mwc-tab-scroller.ts:57:18 - error TS2345: Argument of type '{ passive: boolean; }' is not assignable to parameter of type 'EventListenerOptions'.
  Object literal may only specify known properties, and 'passive' does not exist in type 'EventListenerOptions'.

57   @eventOptions({passive: true})
```

### Browsers Affected
Not applicable

### Versions
- lit-element: v0.14.0

CaptainCodeman:
The passive option is defined on the `AddEventListenerOptions` interface, not `EventListenerOptions`

Closes #325

### DIFF
--- a/src/lib/decorators.ts
+++ b/src/lib/decorators.ts
@@ -116,7 +116,7 @@ function _query<T>(queryFn: (target: NodeSelector, selector: string) => T) {
  *       }
  *     }
  */
-export const eventOptions = (options: EventListenerOptions) =>
+export const eventOptions = (options: AddEventListenerOptions) =>
     (proto: any, name: string) => {
       // This comment is here to fix a disagreement between formatter and linter
       Object.assign(proto[name], options);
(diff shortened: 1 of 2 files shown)


sorvell:
Thanks for this PR! Unfortunately, we missed it before adding #357. This PR has a better fix and we've incorporated that into #357 which we'll merge instead of this only because it has a couple additional tests.

PR #328, opened 2018-12-03
use AddEventListenerOptions for event decorator

43081j:
As you can see [here](https://github.com/Polymer/lit-html/blob/3172fbd32fd6b08cee1b3d3cea2505b4a46c1998/src/lib/parts.ts#L418), it seems we should really be using `AddEventListenerOptions` in our decorator (i know this isn't necessarily of the same type, though).

`EventListenerOptions` doesn't contain `passive` and what not, so this doesn't seem correct as it currently is.

### DIFF
--- a/src/lib/decorators.ts
+++ b/src/lib/decorators.ts
@@ -116,7 +116,7 @@ function _query<T>(queryFn: (target: NodeSelector, selector: string) => T) {
  *       }
  *     }
  */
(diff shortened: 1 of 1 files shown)


CaptainCodeman:
Duplicate of #325 ?

43081j:
Yup I clearly didn't search hard enough. 

Good work on fixing it 👍

PR #344, opened 2018-12-07
[docs] Add scripts for managing docs

keanulee:
The idea is to add a package.json for all the product docs sites so that they can all use the same commands:

```
npm run build
npm run serve
npm run deploy
```

And hide the jekyll/gulp stuff behind this.

Related: https://github.com/Polymer/lit-html/pull/685


keanulee:
webcomponentsjs/package.json is an artifact of checking in the output of `polymer build` (for the code preview tool). I guess since we're not using GH pages anymore, we don't need to check in that output.

docs/package.json
```diff
@@ -1,23 +1,10 @@
 {
-  "name": "lit-element-docs",
-  "version": "1.0.0",
-  "description": "lit-element documentation",
-  "main": "index.js",
+  "private": true,
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "build": "polymer build && jekyll build",
+    "serve": "dev_appserver.py _site/app.yaml",
```
arthurevans: We should add a unique port # for each site, like `--admin_port 8100 --port=9000 _site`  ...

Not only does this let you run more than one at one time, but it prevents the service workers from stomping on each other. (Possibly it's just polymer-library that has a SW at this point? But everytime I switch projects I have to work to clear the previous one out of the cache.)

keanulee: Is [using `-- --args`](https://stackoverflow.com/questions/11580961/sending-command-line-arguments-to-npm-script) satisfactory?

```
npm run serve -- --admin_port 8100 --port=9000
```

arthurevans: I'd rather bake it into the files, rather than have to remember to do it every time I launch a site. Do the other sites use service worker? If not, we could just change the port for the library site and leave everything else as the default.

arthurevans: Happy to do this myself as a follow-on PR, too.

keanulee: Only polymer-library-docs has SW. I'll let you pick a port and add that after. I think you would only need `--port`, not `--admin_port`

arthurevans: Yep. Only need unique admin ports if you want to run multiple instances of dev_appserver at the same time (why it can't auto-pick open ports, I don't know, but so it goes).

docs/package.json
```diff
@@ -1,23 +1,10 @@
 {
-  "name": "lit-element-docs",
-  "version": "1.0.0",
-  "description": "lit-element documentation",
-  "main": "index.js",
+  "private": true,
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "build": "polymer build && jekyll build",
```
katejeffreys: any problem with adding `npm install` to the build command?

```
"build": "npm install && polymer build && jekyll build",
```

keanulee: I don't think that's a good idea - `npm i` is only run once, then `npm run build` after local doc modification. Also, running `npm i` from a package.json script seems like a bad idea (you could imagine what happens if you put `npm i` in `postinstall`...)

katejeffreys: I get it :) Thanks!

PR #350, opened 2018-12-11
[docs] Generate API docs

### LINKED ISSUE #339: [docs] Generate API docs

katejeffreys:
Staged https://20181211t022319-dot-polymer-lit-element.appspot.com/api/index.html

`typedoc --readme none --mode modules --theme minimal --plugin none --target ES6 --ignoreCompilerErrors --exclude '{**/*test*,**/node_modules/**,**/test/**}' --excludeNotExported --excludePrivate --out ./docs/api ./src/lit-element.ts ./src/lib/*.ts`

fixes #339


arthurevans:
Can we mark the imported render method private? Currently, the lit-html "render" method shows up as a property and the LitElement render method shows up as a method. 

https://20181211t022319-dot-polymer-lit-element.appspot.com/api/classes/_lit_element_.litelement.html#render

Having two methods with the same name that do completely different things, and have different signatures is hella confusing. It doesn't help that the lit-html `render` method shows up first in the docs.

arthurevans:
Also "Constructor" looks really weird here: https://20181211t022319-dot-polymer-lit-element.appspot.com/api/modules/_lib_decorators_.html#constructor 

... and on the same page, "query" and "queryAll" show up as variables with really weird signatures, as opposed to the other decorators that look more or less normal.

UpdatingElement docs include the license header in place of module docs.

Currently, we include all of the HTMLElement docs as inherited, so you can either turn off inherited in LitElement (and not see stuff from UpdatingElement) or turn it on, and see an absolute flood of props, etc. Wonder if we could treat these as external somehow.

PR #356, opened 2018-12-11
introduce default type serializers

### LINKED ISSUE #264: Discussion: supported serializers out of the box
I recently wrote [this blog post](https://43081j.com/2018/10/lit-element-types) about the fact that we don't support `type: Object` and `type: Array` like Polymer did. I also made [this repository](https://github.com/43081j/lit-element-serializers).

This is a thing people assume quite often, [even in the recent docs PR](https://github.com/Polymer/lit-element/blob/e205f51b67bda647001248afa43545b28c8135a4/docs/projects/try/expressions/custom-element.js#L6).

So I think we need to make a decision before 1.0, one of the following two choices:

* Ship serializers for `Object` and `Array` (and possibly require that consumers import them rather than supporting them by default)
* Somehow make it clearer that `type` is _not_ a type constructor but rather a type serializer (at this point we can't really rename `type` but maybe we could document it _very well_)

I'm happy to PR this if we go with the former.

Right now, people almost always assume because they see examples use `type: Number`, `type: String`, etc, they should then use `type: Date`, `type: Object`, `type: Array` and so on. This won't work as expected, but people rarely bump into this realisation as they pass their data via properties rather than attributes.

We do want to keep the core as light as we can, so if we do introduce this, it probably belongs in an extra module we must import ourselves.

cc @justinfagnani

43081j:
Fixes #264.

Maybe this is a solution?

* The old `type` option has been renamed to `serializer`
* `type` is now one of the primitive constructors (`Boolean`, `String`, `Number`, `Object` or `Array`) if set
* `AttributeSerializer` has been renamed to `ComplexAttributeSerializer`
* `AttributeType` has been renamed to `AttributeSerializer`
* `AttributeType` is a type alias for the basic primitive constructors
* Deserialization and serialization checks for a default serializer if none is specified and a type has been

This would reduce some confusion, especially for those coming from Polymer, as the type system will be _almost_ the same (there is no support for `Date`). Additionally, much of the content out there about lit mistakenly uses `type: Object` often, so this alone tells me people are assuming that is the correct syntax already.

Examples:

```typescript
@property({ type: Object })
@property({ serializer: (val) => customSerializer(val) })
@property({
  serializer: {
    toAttribute(val) { return `${val}`; },
    fromAttribute(val) { return JSON.parse(val); }
  }
});
```

When using something like `type: Object`, the following occurs:

```js
el.myProp = { a: 5 }; // myprop='{"a":5}'
el.setAttribute('myprop', '{"a":6}'); // myProp === 6
```

If you have better ideas or other plans, im happy to close, just figured i'd push this since I had it.

cc @sorvell

### DIFF
--- a/src/lib/updating-element.ts
+++ b/src/lib/updating-element.ts
@@ -355,7 +355,7 @@ export abstract class UpdatingElement extends HTMLElement {
     const type = options && options.type;
     let serializer = options && options.serializer;
 
-    if (type !== undefined) {
+    if (serializer === undefined && type !== undefined) {
       serializer = getDefaultSerializer(type);
     }
 
@@ -385,7 +385,7 @@ export abstract class UpdatingElement extends HTMLElement {
     const type = options && options.type;
     let serializer = options && options.serializer;
 
-    if (type !== undefined) {
+    if (serializer === undefined && type !== undefined) {
       serializer = getDefaultSerializer(type);
     }
 
--- a/src/lib/updating-element.ts
+++ b/src/lib/updating-element.ts
@@ -15,7 +15,7 @@
 /**
  * Converts property values to and from attribute values.
  */
-export interface AttributeSerializer<T = any> {
+export interface ComplexAttributeSerializer<T = any> {
 
   /**
    * Deserializing function called to convert an attribute value to a property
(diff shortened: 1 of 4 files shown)


src/lib/updating-element.ts
```diff
@@ -266,14 +353,19 @@ export abstract class UpdatingElement extends HTMLElement {
   private static _propertyValueFromAttribute(value: string,
                                              options?: PropertyDeclaration) {
     const type = options && options.type;
-    if (type === undefined) {
+    let serializer = options && options.serializer;
+
+    if (type !== undefined) {
```
admwx7: This requires omitting `type` if you want to run a custom serializer.  Would it make more sense to instead drop in the a default serializer based on type only if you didn't explicitly provider one?

sorvell:
Thanks very much for the PR, it inspired the changes in #369. There were just a few more changes desired than seemed feasible to describe in a review so we incorporated your ideas into the other PR instead.

The serializers/deserializers there are almost exactly the same. One difference is that Polymer allows `Object` based properties to be non-objects which seems a little loser than we want here.

Other than that, instead of `serializer` (I realized I suggested that), we went with `converter` and decided to pass `type` to it. This way `type` is really just a hint for the converter. 

We welcome your feedback on #369.

PR #357, opened 2018-12-12
Adds passive/once to `EventListenerOptions` type

### LINKED ISSUE #325: @eventOptions decorator doesn't allow passive to be set

sorvell:
Fixes #325.

Despite this supposedly being in TypeScirpt (https://github.com/Microsoft/TypeScript/issues/9548), it's not, so we augment the type here.

Also adds tests for `once` and `passive` options.


CaptainCodeman:
The options are defined on the `AddEventListenerOptions` interface, not `EventListenerOptions`.

See: https://github.com/Polymer/lit-element/pull/326

src/lib/decorators.ts
```diff
@@ -90,6 +90,16 @@ function _query<T>(queryFn: (target: NodeSelector, selector: string) => T) {
   };
 }
 
+// TODO(sorvell): augment TypeScript's EventListenerOptions which does
+// not include these options despite
+// https://github.com/Microsoft/TypeScript/issues/9548
+declare global {
+  interface EventListenerOptions {
```
CaptainCodeman: Aren't these already defined in Typescript here?

https://github.com/Microsoft/TypeScript/blob/master/lib/lib.dom.d.ts#L33

PR #359, opened 2018-12-12
Ensure user accessor is never overridden

### LINKED ISSUE #286: getter/setter pairs should be available in extensions
### Description
When extending a class with getter/setter controlled property management, these effects should persist in child extensions.

#### Live Demo
https://glitch.com/edit/#!/nervous-columnist?path=index.html:24:0

If you open you console, you should see a log of `parent` and one of `child` but the second log is missing.

#### Steps to Reproduce
1. Create an element that includes the `mood` attribute to property relationship.
2. Use a getter/setter pair to extend the response to changing the value of `mood`.
3. Create a new element with a class that extends the class of the previous element.
4. Attempt to use the getter/setter functionality....

#### Expected Results
Getter/Setter works.

#### Actual Results
Getter/Setter is ignored... 😢 

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari 11
- [x] Safari 10
- [x] IE 11

### Versions
 - "@polymer/lit-element": "^0.6.0",
 - "polymer-cli": "next",
 - "@webcomponents/webcomponentsjs": "latest"

sorvell:
Fixes #286.

* When a property is declared in `properties` or via the `property` decorator, we will _always_ generate an accessor on the current class that enqueues an update via `requestUpdate`

* However, if an accessor already existed on the class prototype chain, that accessor will be effectively super'ed to for both `set` and `get` (while still ensuring an update is enqueued); this is achieved by capturing the existing accessor and calling them from the generated accessor

* This also removes the optimization of taking the property options from the closure where the property is defined. This allows subclasses to redefine property options without needing to redefine the accessor.

* Finally, also removes the private `_requestPropertyUpdate` which would be superfluous with this change.


src/test/lit-element_test.ts
```diff
@@ -1243,26 +1310,238 @@ suite('LitElement', () => {
...
+          bar : {type : String, noAccessor : true}
+        };
+      }
+      set foo(value: string|null) { this.setAttribute('foo', value as string); }
+      get foo() { return this.getAttribute('foo') || 'defaultFoo'; }
+      set bar(value: string|null) { this.setAttribute('bar', value as string); }
+      get bar() { return this.getAttribute('bar') || 'defaultBar'; }
+      attributeChangedCallback(name: string, old: null|string,
+                               _value: null|string) {
+        this.requestUpdate(name, old);
```
sorvell: Let's call `super.attributeChangedCallback` here because otherwise an extension would not be compatible with `LitElement`'s reflect system and it might be confusing if someone looked at this as an example.

src/test/lit-element_test.ts
```diff
@@ -1243,26 +1310,238 @@ suite('LitElement', () => {
...
+    el.foo = 'foo3';
+    await el.updateComplete;
+    assert.equal(el.foo, 'foo3');
+    assert.equal(el.getAttribute('foo'), 'foo3');
+    assert.equal(el.getAttribute('bar'), 'bar');
+    assert.equal(el.shadowRoot!.textContent, 'foo3-bar');
+    assert.equal(el._updateCount, 3);
+  });
+
+  test('attributeChangedCallback based rendering', async () => {
```
sorvell: Suggest duplicating this test without `noAccessor: true` and not overriding `attributeChangedCallback`.

src/test/lit-element_test.ts
```diff
@@ -1243,26 +1310,238 @@ suite('LitElement', () => {
...
+    assert.equal(sup._oldFoo, 10);
+    assert.equal(sup._supSet, 4);
+    assert.equal(sup.shadowRoot!.textContent, '5');
+
+    const sub = new Sub();
+    container.appendChild(sub);
+    await sub.updateComplete;
+    assert.equal(sub.foo, 0);
+    assert.equal(sub._oldFoo, undefined);
+    assert.equal(sub._subSet, 1);
```
sorvell: Also test `sub._supSet` in this assert and the ones below.

src/test/lit-element_test.ts
kevinpschaaf: Done

src/test/lit-element_test.ts
kevinpschaaf: Done

src/test/lit-element_test.ts
kevinpschaaf: Done

kevinpschaaf:
Won't merge until Sauce service incident is resolved and tests pass.

PR #360, opened 2018-12-12
Redesign

yuinchien:
Updated latest footer, slack url.

Edited content for better layout.


docs/css/main.css
```diff
@@ -17,32 +17,62 @@ h1 {
...
+ol li,
+ul li {
+  padding-left: 12px;
+}
+
 article p {
   font-family: 'Roboto', 'Noto', sans-serif;
   font-size: 16px;
   line-height: 26px;
+  max-width: 720px;
```
frankiefu: Should we set `max-width`, `font-family`, and `font-size` on `article` instead of `article p`?

docs/css/theme.css
```diff
@@ -1,5 +1,6 @@
 /* Theme colors */
-a {
+a,
+h1 {
   color: #00A3E6;
```
katejeffreys: `#00A3E6` against `#FFFFFF` background is probably hard for some users to read because of the low contrast ratio. Could we consider some options for more accessible theme colors?

docs/index.md
```diff
@@ -11,36 +11,69 @@ layout: default
...
-## Fast, lightweight web components
+<div class="responsive-row">
 
-LitElement is a simple base class for creating fast, lightweight web components that work in any web page with any framework.
+<div style="flex:1;">
+<h2>
+Fast, lightweight web components
+</h2>
+<p>
+LitElement is a simple base class for creating fast and lightweight web components that work in any web page with any framework.
```
katejeffreys: The flow of this sentence seems strange after the comma was swapped for 'and'.

docs/index.md
```diff
@@ -11,36 +11,69 @@ layout: default
...
+</p>
+</div>
 
-For rendering, LitElement uses [lit-html](https://lit-html.polymer-project.org/)–a fast HTML templating library. To build an app out of LitElement components, check out [PWA Starter Kit](https://pwa-starter-kit.polymer-project.org/).
+<div style="flex:1;">
+<h2>
+Using lit-html
+</h2>
+<p>
+For rendering, LitElement uses [lit-html](https://lit-html.polymer-project.org/) – a fast HTML templating library. To build an app out of LitElement components, check out [PWA Starter Kit](https://pwa-starter-kit.polymer-project.org/).
```
katejeffreys: Google style guide says no spaces around the dash https://developers.google.com/style/dashes

Looks like I need to make it an em dash (—) instead of an en dash (-) tho

docs/index.md
```diff
@@ -86,44 +119,43 @@ LitElement uses [lit-html](https://github.com/Polymer/lit-html) to define and re
...
+
 <section>
 <div class="wrapper">
-<div class="responsive-row center">
-<div style="max-width: 600px">
-
-LitElement is brought to you by developers on the Google Chrome team with the input of web developers at organizations big and small around the world. 
 
+<h1 class="title">Browser Compatibility</h1>
+<h2 class="description">PWA Starter Kit works in all major browsers (Chrome, Firefox, IE, Edge, Safari, and Opera). </h2>
```
katejeffreys: `<h2 class="description">PWA Starter Kit works in all major browsers`

Should be `LitElement` instead of `PWA Starter Kit`

docs/_config.yml
```diff
@@ -1,8 +1,8 @@
 name: LitElement
-description: A simple base class for creating fast, lightweight web components
+description: A simple base class for creating fast and lightweight web components
```
katejeffreys: This sentence feels weird to me after switching the comma for "and".

docs/css/main.css
katejeffreys: I think there is some inconsistent width introduced here, e.g. lists and stackblitz embeds are wider than the article.

docs/_config.yml
yuinchien: The reason that I tried to replace comma with "and" is for better text flow/layout. The original blurb has huge gap and results in unbalanced text flow. 
screen shot 2018-12-13 at 10 13 28 am

Can we try tweak the blurb to make the text flow more even like this? 
screen shot 2018-12-13 at 10 13 54 am

docs/css/theme.css
yuinchien: Thanks for the feedback. Personally the color contrast is not a problem for me, but i understand some people might have issue. The blue should work with the color palette which is designed to be vibrant/energetic. I updated to the color to Material Blue 500 #2196f3, which should work well with others, and improve contrast. 

screen shot 2018-12-13 at 11 28 15 am

screen shot 2018-12-13 at 11 29 02 am

docs/_config.yml
```diff
@@ -1,8 +1,8 @@
 name: LitElement
-description: A simple base class for creating fast, lightweight web components
+description: A simple base class for creating fast, <br/> lightweight web components
```
keanulee: No tags here - description is also used in `<meta name="description" content="{{ site.description }}">`

yuinchien:
@katejeffreys TOC has ordered list nested inside unordered list
screen shot 2018-12-13 at 4 37 08 pm

yuinchien:
@frankiefu will propagate main.css!

PR #367, opened 2018-12-13
Use lit-html 1.0 rc

azakus:


package.json
```diff
@@ -40,7 +40,7 @@
   },
   "typings": "lit-element.d.ts",
   "dependencies": {
-    "lit-html": "^0.14.0"
+    "lit-html": "^1.0.0-rc.1"
```
justinfagnani: Since 1.0 isn't breaking from 0.14, can you include both in the version range? Then if someone else is still on 0.14 they're less likely to get duplicate packages.

azakus: OK

azakus: Hmm, the [semver calculator](https://semver.npmjs.com/) seems to suggest that isn't possible with RCs

azakus: Testing with a package.json confirms that it won't work.

PR #368, opened 2018-12-13
Adds `performUpdate` and defers `update` until connection

### LINKED ISSUE #258: Multiple `update`s for an element may run during parsing depending on script type
### Description
If an element definition is loaded synchronously with a legacy script, `update` may be called multiple times for any main-document elements that follow the script in the page: one for any defaults set in the `constructor`, plus one for each `attributeChangedCallback`.

If an element definition is loaded using defer semantics (e.g. `type="module"` or `defer`), `update` is only called once for such elements.

The assumption that one microtask is enough to collect the constructor defaults and overrides via attributes appears to be insufficient in the first case above.

#### Live Demo
Note console logging:
* Expected (using module script): [demo](https://hushed-algebra.glitch.me/index.html), [code](https://glitch.com/edit/#!/hushed-algebra?path=index.html:8:0)
  * Update is called once
* Unexpected (using bundled legacy script): [demo](https://hushed-algebra.glitch.me/webpack.html), [code](https://glitch.com/edit/#!/hushed-algebra?path=webpack.html:7:4)
  * Update is called 4 times (ctor + 3x attributes)

It appears we need to add back the optimization (used in Polymer 2) to not flush property changes (e.g. the initial `_validate` -> `update` -> `render`) until the element has been connected, to ensure efficient startup of the element and avoid spamming the update/render lifecycles.

Applying the following stopgap `DeferUntilConnected` mixin demonstrates the expected improvements, but ideally `changedProperties` should accumulate until the actual `update`, which this does not attempt: [code](https://glitch.com/edit/#!/honey-mile?path=index.js:4:0), [demo](https://honey-mile.glitch.me/webpack.html)

### Browsers Affected

- [x] Chrome
- [ ] Firefox
- [ ] Edge
- [x] Safari 12
- [x] Safari 11
- [ ] IE 11

### Versions
- lit-element: v0.6.2

### LINKED ISSUE #290: Make `validate` and `invalidate` protected
These are useful extension points for scheduling rendering so they should be protected.

sorvell:
Fixes #290. The previous private `_validate` method has been made protected and named `performUpdate`. It's all now possible to return a promise from this method and doing so will block the update's completion until the promise is resolved. This allows control over the timing of update and integration with schedulers.

Fixes #258. Updates are now deferred until element connection. This ensures that microtask checkpoints that may occur during element construction do not generate errors or cause extra update cycles.


src/lib/updating-element.ts
```diff
@@ -468,69 +480,83 @@ export abstract class UpdatingElement extends HTMLElement {
...
-   * Validates the element by updating it.
-   */
-  private _validate() {
+   * Performs an element update.
+   *
+   * You can override this method to change the timing of updates. For instance,
+   * to schedule updates to occur just before the next frame:
+   *
+   * ```
+   * protected async scheduleUpdate(): Promise<unknown> {
```
kenchris: shouldn't this be performUpdate() ?

src/lib/updating-element.ts
```diff
@@ -468,69 +480,83 @@ export abstract class UpdatingElement extends HTMLElement {
...
+    this._updateState = this._updateState | STATE_UPDATE_REQUESTED;
+    let resolve: (r: boolean) => void;
+    const previousUpdatePromise = this._updatePromise;
+    this._updatePromise = new Promise((res) => resolve = res);
+    // Ensure any previous update has resolved before updating.
+    // This `await` also ensures that property changes are batched.
+    await previousUpdatePromise;
+    // Make sure the element has connected before updating.
+    if (!this._hasConnected) {
+      await new Promise((res) => this._hasConnectedResolver = res);
```
kevinpschaaf: It seems unfortunate to need to introduce another promise to synchronize a pending update with connected, but I get that it is needed to keep the `updateComplete` promise pending to allow someone to `await el.updateComplete` before it is connected.  I don't have a better idea, other than saving the `_updatePromise` resolver and letting connected's call to `requestUpdate` use the same promise.  Not sure that's any better.

Would it be too weird to just document that the pending update system isn't ready until first connected, so `updateComplete` shouldn't be awaited until the element is connected?  That would allow you to just bail in `requestUpdate` (since you'll always come back in when connected).

If you don't like either of those suggestions this code is fine, not that big of a deal.  Just somehow feels more complicated than it should be and wanted to brainstorm a little.

src/lib/updating-element.ts
```diff
@@ -468,69 +480,83 @@ export abstract class UpdatingElement extends HTMLElement {
...
+          this._reflectingProperties.set(name, options);
+        }
+        // abort the request if the property should not be considered changed.
+      } else {
+        shouldRequestUpdate = false;
+      }
     }
-    return this._invalidate();
+    if (!this._hasRequestedUpdate && shouldRequestUpdate) {
+      this._setupUpdate();
```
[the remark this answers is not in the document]

src/lib/updating-element.ts
```diff
@@ -468,69 +480,83 @@ export abstract class UpdatingElement extends HTMLElement {
...
+    // Make sure the element has connected before updating.
+    if (!this._hasConnected) {
+      await new Promise((res) => this._hasConnectedResolver = res);
     }
-    // track old value when changing.
-    if (!this._changedProperties.has(name)) {
-      this._changedProperties.set(name, oldValue);
+    // Allow `performUpdate` to be asynchronous to enable scheduling of updates.
+    const result = this.performUpdate();
+    if (result != null &&
```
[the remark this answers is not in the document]

src/lib/updating-element.ts
sorvell: Acknowledged, but think it makes sense to keep as is.

PR #369, opened 2018-12-13
Changes property options to support `converter`

### LINKED ISSUE #264: Discussion: supported serializers out of the box

sorvell:
Fixes #264

Changes `type` to be only a hint to the `converter` option which has the previous `type` functionality, an object with `toAttribute` and `fromAttribute` or just a function which is `fromAttribute`. In addition to the `value` these functions now also get the property's `type`.

Also provides a default converter that supports `Boolean`, `String`, `Number`, `Object`, and `Array` out of the box.

In addition, numbers and strings now become `null` if their reflected attribute is removed.


README.md
```diff
@@ -35,11 +35,12 @@ for additional information on how to create templates for lit-element.
...
     If a string, the string value is observed (e.g `attribute: 'foo-bar'`).
-    * `type`: Indicates how to serialize and deserialize the attribute to/from a property.
+    * `converter`: Indicates how to serialize and deserialize the attribute to/from a property.
     The value can be a function used for both serialization and deserialization, or it can
     be an object with individual functions via the optional keys, `fromAttribute` and `toAttribute`.
-    `type` defaults to the `String` constructor, and so does the `toAttribute` and `fromAttribute`
-    keys.
+    * `type`: Indicates the type of the property. This is used only as a hint for the
+    `converter` to determine how to serialize and deserialize the attribute
+    to/from a property.
```
kevinpschaaf: Add: 
> Default converters are provided for `String`, `Number`, ...

src/lib/updating-element.ts
```diff
@@ -90,9 +99,33 @@ type AttributeMap = Map<string, PropertyKey>;
...
-const toBooleanAttribute = (value: string) => value ? '' : null;
+export const defaultConverter: ComplexAttributeConverter = {
+
+  toAttribute(value: any, type?: any) {
+    switch (type) {
+    case Boolean:
+      return value ? '' : null;
+    case Object:
+    case Array:
+      return JSON.stringify(value);
```
kevinpschaaf: Based on experience with this being a footgun for users, consider anything better?

src/lib/updating-element.ts
```diff
@@ -260,21 +294,16 @@ export abstract class UpdatingElement extends HTMLElement {
...
+   * `converter` or `converter.fromAttribute` property option.
    */
   private static _propertyValueFromAttribute(value: string,
                                              options?: PropertyDeclaration) {
     const type = options && options.type;
-    if (type === undefined) {
-      return value;
-    }
-    // Note: special case `Boolean` so users can use it as a `type`.
+    const converter = options && options.converter || defaultConverter;
```
kevinpschaaf: Shouldn't need a null check since there's a `defaultPropertyDeclaration`?

README.md
sorvell: Added to `converter` section.

README.md
```diff
@@ -35,11 +35,16 @@ for additional information on how to create templates for lit-element.
...
     be an object with individual functions via the optional keys, `fromAttribute` and `toAttribute`.
-    `type` defaults to the `String` constructor, and so does the `toAttribute` and `fromAttribute`
-    keys.
+    A default `converter` is used if none is provided; it supports
+    `Boolean`, `String`, `Number`, `Object`, and `Array`.
+    * `type`: Indicates the type of the property. This is used only as a hint for the
+    `converter` to determine how to serialize and deserialize the attribute
+    to/from a property. Note, when a property changes and the converter is used
+    to update the attribute, the property is never updated again as a result of
+    the attribute changing, and visa versa.
```
vicb:
```suggestion
    the attribute changing, and vice versa.
```

src/lib/updating-element.ts
```diff
@@ -15,27 +15,28 @@
 /**
  * Converts property values to and from attribute values.
  */
-export interface AttributeSerializer<T = any> {
+export interface ComplexAttributeConverter<Type = any, TypeHint = any> {
 
   /**
    * Deserializing function called to convert an attribute value to a property
```
vicb: IMO `Deserializing` and `Deserializing` is a confusing terminology.
Is it still needed after this change ?

README.md
vicb: `docs/_guide/properties.md` should also be updated, see #370

src/lib/updating-element.ts
sorvell: Added this #375 to ensure this is well documented.

src/lib/updating-element.ts
```diff
@@ -490,12 +523,13 @@ export abstract class UpdatingElement extends HTMLElement {
...
-    // track old value when changing.
+    // Track old value when changing.
     if (!this._changedProperties.has(name)) {
       this._changedProperties.set(name, oldValue);
     }
-    // add to reflecting properties set
-    if (options.reflect === true) {
+    // Add to reflecting properties set if `reflect` is true and the property
+    // is not reflecting from the attribute
+    if (options.reflect === true &&  !(this._updateState & STATE_IS_REFLECTING)) {
```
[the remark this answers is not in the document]

src/lib/updating-element.ts
```diff
@@ -416,41 +446,44 @@ export abstract class UpdatingElement extends HTMLElement {
       name: PropertyKey, value: unknown,
       options: PropertyDeclaration = defaultPropertyDeclaration) {
     const ctor = (this.constructor as typeof UpdatingElement);
-    const attrValue = ctor._propertyValueToAttribute(value, options);
-    if (attrValue !== undefined) {
```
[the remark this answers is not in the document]
sorvell: Done and added test.

src/lib/updating-element.ts

README.md
sorvell: Done.

sorvell: Added this #375 to ensure this is well documented.

carsonpowers:
> In addition, numbers and strings now become null if their reflected attribute is removed.

Shouldn't numbers and strings become `undefined` if they are removed? Changing them to `null` means that defaults set by the component will now be bypassed:

```
function func(foo='bar'){
  console.log(foo);
}

func(null); // null
func(undefined); // 'bar'
```

PR #370, opened 2018-12-14
[Docs] Fix property type - function case

vicb:
The serialization (`.toAttribute`) is handled by `String` at https://github.com/Polymer/lit-element/blob/4b20f80bc3145b0a0f7421f23d9a9f9e3d43e922/src/lib/updating-element.ts#L293-L298

### DIFF
--- a/docs/_guide/properties.md
+++ b/docs/_guide/properties.md
@@ -118,10 +118,10 @@ By default, LitElement uses the `String` constructor to serialize and deserializ
 
 `type` can be: 
 
-* A function that performs both serialization and deserialization:
+* A function that performs the deserialization (i.e. returns the property value from the attribute value). The `String` constructor is used to reflect the property value to the attribute.
 
   ```js
-  propName: { type: someFunction }
+  propName: { type: deserializerFunction }
   ```
 
 * An object with two function properties, `fromAttribute` and `toAttribute`. `fromAttribute` performs deserialization, and `toAttribute` performs serialization:


katejeffreys:
@kenchris @justinfagnani Could you comment on this issue please?

My understanding from what @vicb points out is that this happens:

```js
const MyObject = { 
  toAttribute: /* ... custom toAttribute */ 
  fromAttribute: /* ... custom fromAttribute */ 
};
const myFunction = /* ... myFunction */ ;
//...
// property declarations:
{ x: { type: Number } }  // serializer is String constructor, deserializer is Number constructor
{ y: { type: MyObject } } // serializer is custom toAttribute, deserializer is custom fromAttribute
{ z: { type: myFunction } } // serializer is String constructor, deserializer is myFunction
```

We also have:

```js
// don't do this
{ w: { type: Object } } 
// Serializer is String constructor
// Deserializer is Object constructor, produces something weird e.g. { 0: "{", 1: "a", 2: ":", 3: "'", ... } 
```

Just wanted to check with you both because of this commit: https://github.com/Polymer/lit-element/commit/b4c5f517655fae936293ae4d6c57350378751716

katejeffreys:
@sorvell  fyi

katejeffreys:
Just saw this https://github.com/Polymer/lit-element/pull/369 which probably changes things

vicb:
@katejeffreys yep it should change. I added [a note](https://github.com/Polymer/lit-element/pull/369/files#r241927423) there because it should have been updated.

Feel free to close this PR (maybe after #369 is updated ?)

kevinpschaaf:
@vicb @katejeffreys I just added some more info regarding the change in #369 to https://github.com/Polymer/lit-element/issues/375#issuecomment-447528189

vicb:
lg

PR #374, opened 2018-12-15
Make compatible with closure advanced compilation

sorvell:
* uses `JSCompiler_renameProperty` where literal names are needed instead of using `@export`
* The private `_classProperties` is no longer created by default. Instead `_ensureClassProperties` was added and is called to make sure that the specific class has its own storage for class properties.
* Adds a test that ensure a subclass cannot alter a superclass. The condition tested here would only fail after merging #359.
* optimization: ensures `LitElement` is marked as `finalized` so that it does not ever try to finalize.


katejeffreys:
Does this need docs changes?

src/lib/updating-element.ts
```diff
@@ -233,25 +244,40 @@ export abstract class UpdatingElement extends HTMLElement {
...
-   * or uses a strict identity check to determine whether or not to request
-   * an update.
+   * Ensures the private `_classProperties` property metadata is created.
+   * In addition to `_finalize` this is also called in `createProperty` to
+   * ensure the `@property` decorator can add property metadata.
    */
-  static createProperty(name: PropertyKey,
-                        options:
-                            PropertyDeclaration = defaultPropertyDeclaration) {
+  /** @nocollapse */
```
justinfagnani: I'm surprised we need nocollapse here. Is it because of `this` access? If so, seems like an unsafe optimization by Closure.

src/lib/updating-element.ts
```diff
@@ -200,29 +208,32 @@ export abstract class UpdatingElement extends HTMLElement {
...
    * Marks class as having finished creating properties.
    */
-  private static _finalized = true;
+  protected static finalized = true;
 
   /**
    * Memoized list of all class properties, including any superclass properties.
    */
-  private static _classProperties: PropertyDeclarationMap = new Map();
+  private static _classProperties?: PropertyDeclarationMap;
```
justinfagnani: why remove the initializer?

src/lit-element.ts
```diff
@@ -22,12 +22,19 @@ export {html, svg} from 'lit-html/lit-html';
 
 export class LitElement extends UpdatingElement {
 
+  /**
+   * Ensure this class is marked as `finalized` as an optimization ensuring
+   * it will not needlessly try to `finalize`.
+   */
+  protected static finalized = true;
```
justinfagnani: Shouldn't we initialize UpdatingElement.finalized to true too then?

kevinpschaaf:
@katejeffreys No, it's just making the type annotations and constructions compatible with closure; no real user-facing considerations there.

src/lib/updating-element.ts
kevinpschaaf: `UpdatingElement` itself will never set anything into its `_classProperties` (since it never has `properties`); the set is created lazily during finalization on actual user subclasses

kevinpschaaf: Can add a comment to that effect.

src/lib/updating-element.ts
```diff
@@ -200,29 +208,32 @@ export abstract class UpdatingElement extends HTMLElement {
   /**
    * Maps attribute names to properties; for example `foobar` attribute
    * to `fooBar` property.
+   * @nocollapse
```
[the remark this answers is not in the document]
kevinpschaaf: No, not related to `hasOwnProperty` (that is dealt with by making sure we use `JSCompiler_renameProperty` anywhere we use a property by name.  It's working around ES6 compiler bugs -- and definitely deserve a comment.

I believe this is right:
* any static method needs @nocollapse to not be removed altogether
* _private_ static fields need @nocollapse to get their initializers (if they're not `private` they are fine)

@rictic Does the above sound correct?

That said, similar to Steve we removed the initializer for `_classProperties`, I'm not entirely sure why `_attributeToPropertyMap` and `properties` need initial values on `UpdatingElement` either... will take a look.

src/lib/updating-element.ts
kevinpschaaf: Done

src/lib/updating-element.ts
kevinpschaaf: See above, all static methods need @nocollapse currently (closure bug workaround).

PR #377, opened 2018-12-16
fix links for next steps

atomikolex:
### Reference Issue
Fixing broken links on main page


katejeffreys:
This was fixed in a thing that was merged this morning I think :) thank you

PR #380, opened 2018-12-18
Properties docs

### LINKED ISSUE #135: Default value of properties location
Please make it more clear that default values do not go in the properties block but rather in the constructor. This question has come up in multiple locations #114, slack, and twitter

### LINKED ISSUE #318: [docs] Clarify docs on hasChanged & update lifecycle (e.g. multiple property changes)
According to the [documentation](https://lit-element.polymer-project.org/docs/properties#haschanged) if an observed property's hasChanged() returns true, LitElement starts an element update; returning false will result in no updates. But this isn't always the case, if you have multiple properties with multiple hasChanged methods, and some return true, others return false. LitElement will still update properties whose hasChanged() return false.

`
import {LitElement, html} from '@polymer/lit-element';

	static get properties() {
		return {

			myProp: {
				type: String,
				attribute: 'my-prop',
				reflect: true,
				hasChanged(newValue, oldValue) {
					return true;
				}
			},

			someProp: {
				type: String,
				hasChanged(newValue, oldValue) {
					return false;
				}
			}
		};
	}

	constructor() {
		super();
		this.myProp = 'myProp';
		this.someProp = 'someProp';
	}

	attributeChangedCallback(name, oldValue, newValue) {
		super.attributeChangedCallback(name, oldValue, newValue);
	}

	changeAttributes() {
		let randomString = Math.floor(Math.random()*100).toString();
		this.setAttribute('my-prop', 'my-prop ' + randomString);
		this.setAttribute('someprop', 'someprop ' + randomString);
		this.requestUpdate();
	}

	changeProperties() {
		let randomString = Math.floor(Math.random()*100).toString();
		this.myProp = 'myProp ' + randomString;
		this.someProp = 'someProp ' + randomString;
	}

	render() {
		return html
<p>${this.myProp}</p>
<p>${this.someProp}</p>
<button @click="${this.changeAttributes}">Change Attributes</button>
<button @click="${this.changeProperties}">Change Properties</button>
;
	}

	updated() {
		console.log(this);
	}
}

customElements.define('my-element', MyElement);
`

#### Expected Results

Properties whose hashChanged() returning false should not be updated.

#### Actual Results

As long as there's one property's hasChanged() returns true, everything gets updated.

### Browsers Affected

- [*] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari 11
- [ ] Safari 10
- [ ] IE 11

### Versions

- "@polymer/lit-element": "latest",
- "@webcomponents/webcomponentsjs": "^2.2.0",
- "lit-html": "latest"

### LINKED ISSUE #354: [docs] Review Properties page for technical accuracy
Review the content at https://lit-element.polymer-project.org/guide/properties for technical accuracy only. 

Markdown source: https://github.com/Polymer/lit-element/blob/master/docs/_guide/properties.md

* [x] Code embeds
* [x] check latest changes on serialization/types/etc
* [x] double check `blah: String` in samples, change to `blah: { type: String }`

### LINKED ISSUE #372: [docs] Property accessors
See https://github.com/Polymer/lit-element/issues/286

* [x] ensure users understand that inclusion in properties is required to get LitElement's value-added features for properties (e.g. adding them to observedAttributes, pre-upgrade instance property handling, etc

* [x] `noAccessor`

```
  /**
   * Indicates whether an accessor will be created for this property. By
   * default, an accessor will be generated for this property that requests an
   * update when set. If this flag is `true`, no accessor will be created, and
   * it will be the user's responsibility to call
   * `this.requestUpdate(propertyName, oldValue)` to request an update when
   * the property changes.
   */
  noAccessor?: boolean;
```

### LINKED ISSUE #375: [docs] Document changes to property declarations in #369
Please refer to the changes noted in the [README](https://github.com/Polymer/lit-element/pull/369/files#diff-04c6e90faac2675aa89e2176d2eec7d8R38) and [CHANGELOG](https://github.com/Polymer/lit-element/pull/369/files#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR13).

We should also add a special note that although it's possible to `reflect: true` a `type: Object|Array` property, this should only be done in exceptional situations since the attribute becomes that JSON version of the property. If the property can have a non-trivial value, this will result in poor performance since the attribute can become a very long string.

katejeffreys:
fixes #135
fixes #372
fixes #354
fixes #375 
fixes #318

https://20181218t132131-dot-polymer-lit-element.appspot.com/guide/properties

^^^updated with feedback


katejeffreys:
@kevinpschaaf @keanulee @sorvell 

Updated properties docs are staged here: https://20181218t010822-dot-polymer-lit-element.appspot.com/guide/properties

Could you (or anyone) please comment on a few questions? 

* When should the user set `noAccessor: true`?
* If I specify `noAccessor` & don't create my own accessors, what should happen?
* If I don't specify `noAccessor` & do create my own accesors, what should happen? Would there be a situation where you want to let LitElement create the generated ones & also create your own? When/why?
* Should the user care/be able to find out what LitElement has called their generated property accessors? If so how can they do that?

docs/_guide/properties.md
```diff
@@ -28,158 +78,232 @@ static get properties() {
...
-  ```js
-  propName: { type: {
-    toAttribute: serializerFunction,
-    fromAttribute: deserializerFunction
-  }}
-  ```
+  * To **reflect** an attribute (set an attribute from a property), the property value must be converted to a string.
 
-By default, `type`, `fromAttribute` and `toAttribute` are the `String` constructor. 
+LitElement defines a **converter** for every declared property. A converter is an object that contains two functions:
```
kevinpschaaf: Maybe "LitElement _accepts_ a **converter** for every declared property"?

Also, a `converter` can be an object with two functions, or just a function (with `fromAttribute` semantics).

docs/_guide/properties.md
```diff
@@ -28,158 +78,232 @@ static get properties() {
...
+* If `toAttribute` returns `null`, the attribute is removed. 
+* If `toAttribute` returns `undefined`, the attribute is not changed.
 
-One option is to use `JSON.parse` and `JSON.stringify`:
+|**Type**|**`fromAttribute` returns**|**`toAttribute` returns**|
+|---|---|---|
+| `String` | Attribute value | Property value |
+| `Number` | Number(attributeValue) | Property value |
+| `Boolean` | `true` if non-`null`, `false` otherwise | `''` if truthy, `null` otherwise |
+| `Array` | `JSON.parse(attributeValue)` | `null` if the property value is `null`;  `JSON.stringify(propertyValue)` otherwise |
```
kevinpschaaf: Actually, fo objects and arrays `toAttribute` returns "`null` (attribute is removed) if the property value is `null` _or_ `undefined`".

docs/_guide/properties.md
```diff
@@ -192,37 +316,102 @@ myProp: { attribute: 'my-prop' }
...
+static get properties() { return { myProp: { type: String } }; }
+set myProp(newValue) { 
+  ...
+  this.requestUpdate('foo', oldVal);
+}
+get myProp() { ... }
 ```
 
-### Configure reflection to attributes
+Call `requestUpdate('propertyName', oldValue)` when implementing a custom property setter to ensure that changes to the property will trigger updates when required. Pass the name and old value of the property to `requestUpdate` so that any property options can be correctly applied.
```
kevinpschaaf: So with the recent change, manually call `requestUpdate` is _only_ required in the `noAccessor` case when we don't generate an accessor for them (should be rare), so I think this can move down to that section.  It's obviously only required if they want the property to trigger an update.

Unless `noAccessor` is true, If a user writes an accessor with the same name as a declared property, a generated accessor is still created that calls `requestUpdate`, but it will "super" to the user accessor... similar to this:

```js
// The original user accessor is saved off (`userAccessor`, with `get` and
// `set` functions), and then the user accessor is overwritten with the generated
// accessor, which "supers" to the saved accessor; looks roughly like this:
set foo(value) {
  const oldValue = this.foo;
  this.requestUpdate('foo' oldValue);
  userAccessor.set.call(this, value);
}
get foo() {
  return userAccessor.get.call(this);
}
```
I think 👆is probably too much implementation detail to put in code, not sure.  But thought it would be useful to understand what we do under the covers to make sure it aligns with the docs.

docs/_guide/properties.md
katejeffreys: Could you expand on that point? I somehow thought the property value being undefined wouldn't affect the attribute. Will try to find where I got that impression (could have been code comments or comments on another issue)

docs/_guide/properties.md
katejeffreys: Thanks, that helps a lot!

docs/_guide/properties.md
katejeffreys: Er never mind that, i guess it would be bad to `JSON.stringify(undefined)`...

PR #392, opened 2018-12-20
Api docs readme

katejeffreys:
Staged here https://20181219t164450-dot-polymer-lit-element.appspot.com


katejeffreys:
Going to separate docs changes from changes to inline comments in source code

PR #398, opened 2018-12-20
Add API readme and regenerate API docs

katejeffreys:


docs/api-readme.md
```diff
@@ -0,0 +1,35 @@
...
+      return html`<p>your template here</p>`;
+  }
+}
+
+customElements.define('my-element', MyElement);
+```
+
+Inherits API functionality from [updating-element](#updating-element).
+
+### updating-element
```
justinfagnani: I think we should write closer to the module you'd need to import, like `lib/updating-element.js`.

We should also show the import statement so it's really clear.

docs/api-readme.md
```diff
@@ -0,0 +1,35 @@
...
+
+[updating-element](/api/modules/_lib_updating_element_.html)
+
+Provides API functionality for LitElement. Manages properties and attributes; Performs element updates.
```
justinfagnani: "Provides API functionality for LitElement." isn't very clear to me, especially why it is separate from LitElement. I'd try something like:

"Custom Element base class that supports declaring observable properties, reflecting attribute to properties, and the core update lifecycle methods"

docs/api-readme.md
```diff
@@ -0,0 +1,35 @@
+# LitElement API Documentation
+
+## Install @polymer/lit-element
+
+```
+npm install --save @polymer/lit-element
```
justinfagnani: This is where I'd like to include the version once we release 1.0: `npm install --save @polymer/lit-element@^1.0.0`

docs/api-readme.md
```diff
@@ -0,0 +1,35 @@
...
+```
+
+## Modules
+
+### lit-element
+
+[lit-element](/api/modules/_lit_element_.html)
```
justinfagnani: The "lit-element" header, and then the "lit-element" link make the page look visually repetitive
screen shot 2018-12-21 at 10 32 05 am

Can we either put the link in the header, or add some text to the link like "lit-element API"? Same below.

docs/api-docs-theme/layouts/default.hbs
```diff
@@ -0,0 +1,69 @@
+<!doctype html>
+<html class="default no-js">
+<head>
```
justinfagnani: is this file copied from somewhere else, or was it generated?

docs/api-readme.md
katejeffreys: Not quite sure what you mean. Where do you want to put the import statement, and where should `updating-element` be changed to `lib/updating-element`?

docs/api-readme.md
katejeffreys: The whole api docs output is visually and otherwise repetitive so it's in theme ;) jk, I'll fix it

docs/api-docs-theme/layouts/default.hbs
katejeffreys: Copied from the default theme per typedoc docs.

docs/api-readme.md
katejeffreys: Why is it separate from LitElement?

docs/api-readme.md
justinfagnani: I think this should be `### lib/updating-element.js`

And right below we should have an import example like you do for lit-element:

```js
import { UpdatingElement } from '@polymer/lib/updating-element.js';
```

Except for the main module, modules don't really have names, they have paths. "updating-element" makes it look like a module name, that you could maybe import with `import {...} from 'updating-element';` or maybe `import {...} from '@polymer/updating-element';`... if you're new to imports, npm scoped packages, etc.

docs/api-readme.md
justinfagnani: Because all those things don't have any dependency on lit-html, so they could be used with another templating system, or none at all. We'll probably split it out into a separate package at some point.

docs/api-readme.md
```diff
@@ -0,0 +1,31 @@
+# LitElement API Documentation
+
+## Install @polymer/lit-element
+
+```
+npm install --save @polymer/lit-element@^1.0.0
```
justinfagnani: Sorry, my previous comment was "once we release 1.0" and just a note. We can't do this quite yet: https://jubianchi.github.io/semver-check/#/^1.0.0/1.0.0-rc.1

docs/api-readme.md
katejeffreys: Should the import statement be 

```js
import { UpdatingElement } from '@polymer/lit-element/lib/updating-element.js';
```

?

docs/api-readme.md
katejeffreys: cool... will that be obvious to the audience?

justinfagnani: I think it would be more obvious if LitElement was described as an extension of UpdatingElement that ties it into lit-html by letting users define a render() method that's called at the right time in update().

katejeffreys:
@justinfagnani Updated and staged here: https://20181221t121611-dot-polymer-lit-element.appspot.com

docs/api-readme.md
blikblum: > o they could be used with another templating system

For the record:

https://github.com/blikblum/snabbdom-element (uses snabbdom virtual dom)
https://github.com/blikblum/raw-element (no templating at all)

katejeffreys:
@justinfagnani LMK if you want any more changes, otherwise I'll merge soonish

PR #399, opened 2018-12-21
WIP: Generate React components

justinfagnani:
Do not merge. Mostly a prototype.

Tests don't run until I do some work load React a different way...


justinfagnani:
@rictic adding you for eyes on TypeScript. It looks like I have the component type reasonable now, but getting the types annotated while passing in the React module as an argument was messier than I'd like.

The generated type is `ComponentClass<ElementClass, void>` which should mean that any property defined in the prototype chain should be a valid prop, and there's no state. I think I need to update the `setProperty` function to not just look in `_classProperties` but anything defined on the prototype so we can set regular, non-UpdatingElement, properties. That would make it not UpdatingElement-specific anymore, so we'd have to see if there are other things to do that would justify this being in the lit-element package, vs something generic.

justinfagnani:
We'll get to this again soon, but closing this since it's not active right now.

PR #400, opened 2018-12-22
WIP: Adds api to help measure elements

### LINKED ISSUE #365: Provide a synchronous update option
### Description

The use of the promise micro-task to batch property updates into a single render is usually useful but can cause issues because components no longer behave like components are expected to and always have - updating the DOM when properties are changed. After setting a property, there is no immediate change to the DOM, so elements that don't know that they have to wait for the promise break. An example of this is the virtual-scroller which assumes it can calculate the size of the just-updated element but at that point it can't, so the size of elements are initially wrong, things overlap and then snap into place.

There is some trickery that can be done to wait on the promise, schedule a rAF to trigger the layout recalc but nothing seems to work 100% correctly and at some point, your container component really shouldn't have to know to treat lit-element components differently and it may not even contain the lit-element based components directly - they may be inside other elements, and then which promise(s) do you wait on? The parent or the child(ren) at the bottom? Imagine a scenario where some component is updated at some point in future to utilize lit-element and happens to be used inside some card inside a list, suddenly things start to break and it's unclear why.

I would prefer elements behave like elements always have and how native elements do - when you set something, they reflect that change in the DOM when you expect. If there is any optimization required where an element shouldn't render something until it has all it's properties / data provided then it can manage that delay itself so there is only a single render but it would still be synchronous when the last property is provided so would work as expected.

The sierpinski triangle type demos of lots of things updating smoothly are interesting but are not representative of the issues I'm typically working on when developing an app. Right now I'm finding that using lit-element based components creates too much extra work and I end up replacing them with vanilla HTMLElement based ones to make things work.

#### Live Demo
I haven't made one yet, it tends to show up in more complex apps, not simple demos, but I'll try and extract the pieces to show the issue when I get chance.

#### Steps to Reproduce
Use a complex lit-element based component inside something that expects to measure and position it.

#### Expected Results
Components update and layout correctly

#### Actual Results
Components overlap because measurement doesn't reflect current property state

### Browsers Affected

- [X] Chrome
- [X] Firefox
- [X] Edge
- [X] Safari 11
- [-] Safari 10
- [-] IE 11

### Versions
All

sorvell:
Work in progress, do not merge.

Fixes #365.

* `updateSubtreeComplete` promise: resolves when entire element subtree is updated: wait and then measure.
* `flushUpdate()`: call to flush any pending elements on the element or its subtree: call and then measure.
* `LitElement.flushUpdates()`: call to flush any pending updates for all LitElements: call and then measure element containing updating elements.
* `patch-layout.js`: module which patches native DOM properties so that they flush any pending updates therefore always returning expected results. (load to work with 3rd party libraries that expect to be able to measure synchronously.

### DIFF
--- a/demo/lit-element.html
+++ b/demo/lit-element.html
@@ -34,22 +34,12 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       static get properties() {
         return {
           nug: {},
-          zot: {},
           foo: {},
           bar: {},
           whales: {type: Number},
           fooBar: {converter: {fromAttribute: parseInt, toAttribute: (value) => value + '-attr'}, reflect: true}
         }
       }
-
-      // a custom getter/setter can be created to customize property processing
-      get zot() { return this.getAttribute('zot'); }
-
-      set zot(value) {
-        this.setAttribute('zot', value);
-        this.invalidate();
-      }
-
 
       constructor() {
         super();
--- a/src/lib/patch-layout.ts
+++ b/src/lib/patch-layout.ts
@@ -1,0 +1,97 @@
+/**
+ * @license
+ * Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
+ * This code may only be used under the BSD style license found at
+ * http://polymer.github.io/LICENSE.txt
+ * The complete set of authors may be found at
+ * http://polymer.github.io/AUTHORS.txt
+ * The complete set of contributors may be found at
+ * http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at
+ * http://polymer.github.io/PATENTS.txt
+ */
+
+import {elementsPendingUpdate, UpdatingElement} from './updating-element.js';
+
+/**
+ * **EXPERIMENTAL**
+ * Loading this module redefines properties that provoke a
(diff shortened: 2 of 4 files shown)


anoblet:
Awesome!

T-Knott-Mesh:
Woot woot !

hyperpress:
Awesome! Thank you.

mercmobily:
This was a while ago... but we could really do with this.

justinfagnani:
Requests for this have come up again on Slack. I think we should consider some combination of mixin + ref + controller design, possibly in labs. Closing this on the old repo though.

mercmobily:
Sorry for the interruption... just my humble opinion. With This one closed,
and #365 closed as well, the information -- and code -- that were in those
tickets will be essentially lost. Especially sorvell's comments about
possible strategies, as well as his WIP code. Isn't it worthwhile to make a
new ticket referencing at least #365 and #400, so that there is a starting
point?

justinfagnani:
The information isn't lost, it's still here. We didn't have a currently opened issue as it was, and so if we wanted to track it we already would have to reopen #365 or open a new issue, which I think we should do in lit/lit

mercmobily:
Would you mind me opening one up referencing these two issues and the
important content-filled comments to give a quick overview?

justinfagnani:
@mercmobily not at all!

PR #401, opened 2018-12-22
Adds `static get styles()`

### LINKED ISSUE #391: Consider adding a specific styling API to better integrate with constructible stylesheets
[Constructible stylesheets](https://wicg.github.io/construct-stylesheets/) are soon coming to the web platform. They should provide a significant performance improvement for styling Shadow DOM!

The intention of the API is to allow sharing of stylesheets between `shadowRoots` that use the same styling, and this is a perfect fit for custom elements. It does strongly discourage modifying styles per instance since the stylesheet would be shared across all elements of a given type. 

Because of this, it may not make sense for users to put non-dynamic `<style>` elements into the `render` method since it looks like these styles can change. Although this is not the case when `shady-render` is used, that's a polyfill specific limitation rather than part of the core design.

Perhaps a `static get style()` could be added which is how users declare element styling.

If so, we either need to integrate this feature with `shady-render` or stop using it.

We should probably consider if this is worth doing prior to a 1.0 release.

sorvell:
Fixes #391.

This should be implemented to return an array of styles defined using the `css` tag function. These styles will be applied with `adoptedStyleSheets` where available and shimmed when not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ This takes advantage of [`adoptedStyleSheets`](https://wicg.github.io/construct-
 * Added the `performUpdate` method to allow control of update timing ([#290](https://github.com/Polymer/lit-element/issues/290)).
 * Updates deferred until first connection ([#258](https://github.com/Polymer/lit-element/issues/258)).
 ### Changed
+* [Breaking] The `createRenderRoot` method has moved from `UpdatingElement` to `LitElement`. Therefore, `UpdatingElement` no longer creates a `shadowRoot` by default ([#391](https://github.com/Polymer/lit-element/issues/391)).
 * [Breaking] Changes property options to add `converter`. This option works the same as the previous `type` option except that the `converter` methods now also get `type` as the second argument. This effectively changes `type` to be a hint for the `converter`. A default `converter` is used if none is provided and it now supports `Boolean`, `String`, `Number`, `Object`, and `Array` ([#264](https://github.com/Polymer/lit-element/issues/264)).
 * [Breaking] Numbers and strings now become null if their reflected attribute is removed (https://github.com/Polymer/lit-element/issues/264)).
 * [Breaking] Previously, when an attribute changed as a result of a reflecting property changing, the property was prevented from mutating again as can happen when a custom
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 
 ### Added
+* Added `static get styles()` to allow defining element styling separate from `render` method.
+This takes advantage of [`adoptedStyleSheets`](https://wicg.github.io/construct-stylesheets/#using-constructed-stylesheets) when possible ([#391](https://github.com/Polymer/lit-element/issues/391)).
 * Added the `performUpdate` method to allow control of update timing ([#290](https://github.com/Polymer/lit-element/issues/290)).
 * Updates deferred until first connection ([#258](https://github.com/Polymer/lit-element/issues/258)).
 ### Changed
--- a/src/lib/css-tag.ts
+++ b/src/lib/css-tag.ts
@@ -38,19 +38,19 @@ const cssLiteralValue = (value: CSSLiteral) => {
 };
 
 export type CSSStyleSheetOrCssText = {
-  cssText?: CSSLiteral,
+  cssText: CSSLiteral,
   styleSheet?: ConstructableStyleSheet}
 ;
 
 export const css = (strings: TemplateStringsArray, ...values: CSSLiteral[]): CSSStyleSheetOrCssText => {
   const cssText = values.reduce((acc, v, idx) =>
       acc + cssLiteralValue(v) + strings[idx + 1], strings[0]);
-  const result: CSSStyleSheetOrCssText = {};
+  const result: CSSStyleSheetOrCssText = {
+    cssText: new CSSLiteral(cssText)
+  };
   if (supportsAdoptedStyleSheets) {
     result.styleSheet = new CSSStyleSheet() as ConstructableStyleSheet;
     result.styleSheet.replaceSync(cssText);
-  } else {
-    result.cssText = new CSSLiteral(cssText);
   }
   return result;
 };
--- a/src/lib/css-tag.ts
+++ b/src/lib/css-tag.ts
@@ -1,14 +1,16 @@
 /**
 @license
 Copyright (c) 2019 The Polymer Project Authors. All rights reserved.
-This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
-The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
-The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
-Code distributed by Google as part of the polymer project is also
-subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+This code may only be used under the BSD style license found at
+http://polymer.github.io/LICENSE.txt The complete set of authors may be found at
+http://polymer.github.io/AUTHORS.txt The complete set of contributors may be
+found at http://polymer.github.io/CONTRIBUTORS.txt Code distributed by Google as
+part of the polymer project is also subject to an additional IP rights grant
+found at http://polymer.github.io/PATENTS.txt
 */
 
-export const supportsAdoptingStyleSheets = ('adoptedStyleSheets' in Document.prototype);
+export const supportsAdoptingStyleSheets =
+    ('adoptedStyleSheets' in Document.prototype);
 
 export class CSSResult {
 
@@ -16,15 +18,14 @@ export class CSSResult {
 
   readonly cssText: string;
 
-  constructor(cssText: string) {
-    this.cssText = cssText;
-  }
+  constructor(cssText: string) { this.cssText = cssText; }
 
-  // Note, this is a getter so that it's lazy. In practice, this means stylesheets
-  // are not created until the first element instance is made.
+  // Note, this is a getter so that it's lazy. In practice, this means
+  // stylesheets are not created until the first element instance is made.
   get styleSheet(): CSSStyleSheet|null {
     if (this._styleSheet === undefined) {
-      // Note, if `adoptedStyleSheets` is supported then we assume CSSStyleSheet is constructable.
+      // Note, if `adoptedStyleSheets` is supported then we assume CSSStyleSheet
+      // is constructable.
       if (supportsAdoptingStyleSheets) {
         this._styleSheet = new CSSStyleSheet();
         this._styleSheet.replaceSync(this.cssText);
@@ -41,13 +42,15 @@ const textFromCSSResult = (value: CSSResult) => {
     return value.cssText;
   } else {
     throw new Error(
-        `Value passed to 'css' function must be a 'css' function result: ${value}.`
-    );
+        `Value passed to 'css' function must be a 'css' function result: ${
+            value}.`);
   }
 };
 
-export const css = (strings: TemplateStringsArray, ...values: CSSResult[]): CSSResult => {
-  const cssText = values.reduce((acc, v, idx) =>
-    acc + textFromCSSResult(v) + strings[idx + 1], strings[0]);
-  return new CSSResult(cssText);
-};
+export const css =
+    (strings: TemplateStringsArray, ...values: CSSResult[]): CSSResult => {
+      const cssText = values.reduce(
+          (acc, v, idx) => acc + textFromCSSResult(v) + strings[idx + 1],
+          strings[0]);
+      return new CSSResult(cssText);
+    };
--- a/src/lib/css-tag.ts
+++ b/src/lib/css-tag.ts
@@ -1,28 +1,40 @@
 /**
 @license
-Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
+Copyright (c) 2019 The Polymer Project Authors. All rights reserved.
 This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
 The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
 The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
 Code distributed by Google as part of the polymer project is also
 subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
 */
-export const supportsAdoptedStyleSheets = ('adoptedStyleSheets' in Document.prototype);
+const supportsConstructableStyleSheets = (() => {
+  let constructable = true;
+  try {
+    new CSSStyleSheet();
+  } catch {
+    constructable = false;
+  }
+  return constructable;
+})();
+
+export const supportsAdoptingStyleSheets = supportsConstructableStyleSheets &&
+    ('adoptedStyleSheets' in Document.prototype);
 
 export class CSSResult {
 
   _styleSheet?: CSSStyleSheet|null;
 
-  constructor(public readonly cssText: string) {
+  readonly cssText: string;
+
+  constructor(cssText: string) {
+    this.cssText = cssText;
   }
 
   // Note, this is a getter so that it's lazy. In practice, this means stylesheets
   // are not created until the first element instance is made.
   get styleSheet(): CSSStyleSheet|null {
     if (this._styleSheet === undefined) {
-      // Note, assume that if adoptedStyleSheets is supported,
-      // constructable stylesheets are as well.
-      if (supportsAdoptedStyleSheets) {
+      if (supportsAdoptingStyleSheets) {
         this._styleSheet = new CSSStyleSheet();
         this._styleSheet.replaceSync(this.cssText);
       } else {
--- a/src/lib/css-tag.ts
+++ b/src/lib/css-tag.ts
@@ -42,7 +42,7 @@ export type CSSStyleSheetOrCssText = {
   styleSheet?: ConstructableStyleSheet}
 ;
 
-export const css = (strings: string[], ...values: CSSLiteral[]): CSSStyleSheetOrCssText => {
+export const css = (strings: TemplateStringsArray, ...values: CSSLiteral[]): CSSStyleSheetOrCssText => {
   const cssText = values.reduce((acc, v, idx) =>
       acc + cssLiteralValue(v) + strings[idx + 1], strings[0]);
   const result: CSSStyleSheetOrCssText = {};
@@ -55,7 +55,7 @@ export const css = (strings: string[], ...values: CSSLiteral[]): CSSStyleSheetOr
   return result;
 };
 
-export const cssLiteral = (strings: string[], ...values: any[]) => {
+export const cssLiteral = (strings: TemplateStringsArray, ...values: any[]) => {
   return new CSSLiteral(values.reduce((acc, v, idx) =>
       acc + cssLiteralValue(v) + strings[idx + 1], strings[0]));
 };
--- a/src/lib/css-tag.ts
+++ b/src/lib/css-tag.ts
@@ -9,7 +9,12 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 */
 export const supportsAdoptedStyleSheets = ('adoptedStyleSheets' in Document.prototype);
 
-class LiteralString {
+interface ConstructableStyleSheet extends CSSStyleSheet {
+  replaceSync(cssText: string): void;
+  replace(cssText: string): Promise<unknown>;
+}
+
+class CSSLiteral {
 
   value: string;
 
@@ -22,31 +27,35 @@ class LiteralString {
   }
 }
 
-const literalValue = (value: LiteralString) => {
-  if (value instanceof LiteralString) {
+const cssLiteralValue = (value: CSSLiteral) => {
+  if (value instanceof CSSLiteral) {
     return value.value;
   } else {
     throw new Error(
-        `non-literal value passed to 'css' function: ${value}`
+        `Non-literal value passed to 'css' function: ${value}.`
     );
   }
 };
 
-export const css = (strings: string[], ...values: any[]) => {
+export type CSSStyleSheetOrCssText = {
+  cssText?: CSSLiteral,
+  styleSheet?: ConstructableStyleSheet}
+;
+
+export const css = (strings: string[], ...values: CSSLiteral[]): CSSStyleSheetOrCssText => {
   const cssText = values.reduce((acc, v, idx) =>
-      acc + literalValue(v) + strings[idx + 1], strings[0]);
-  let result;
+      acc + cssLiteralValue(v) + strings[idx + 1], strings[0]);
+  const result: CSSStyleSheetOrCssText = {};
   if (supportsAdoptedStyleSheets) {
-    result = new CSSStyleSheet();
-    // TODO(sorvell): fix type.
-    (result as any).replaceSync(cssText);
+    result.styleSheet = new CSSStyleSheet() as ConstructableStyleSheet;
+    result.styleSheet.replaceSync(cssText);
   } else {
-    result = new LiteralString(cssText);
+    result.cssText = new CSSLiteral(cssText);
   }
   return result;
 };
 
 export const cssLiteral = (strings: string[], ...values: any[]) => {
-  return new LiteralString(values.reduce((acc, v, idx) =>
-      acc + literalValue(v) + strings[idx + 1], strings[0]));
+  return new CSSLiteral(values.reduce((acc, v, idx) =>
+      acc + cssLiteralValue(v) + strings[idx + 1], strings[0]));
 };
--- a/src/lib/css-tag.ts
+++ b/src/lib/css-tag.ts
@@ -9,53 +9,42 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 */
 export const supportsAdoptedStyleSheets = ('adoptedStyleSheets' in Document.prototype);
 
-interface ConstructableStyleSheet extends CSSStyleSheet {
-  replaceSync(cssText: string): void;
-  replace(cssText: string): Promise<unknown>;
-}
+export class CSSResult {
 
-class CSSLiteral {
+  _styleSheet?: CSSStyleSheet|null;
 
-  value: string;
-
-  constructor(value: string) {
-    this.value = value.toString();
+  constructor(public readonly cssText: string) {
   }
 
-  toString() {
-    return this.value;
+  // Note, this is a getter so that it's lazy. In practice, this means stylesheets
+  // are not created until the first element instance is made.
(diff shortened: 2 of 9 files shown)


daKmoR:
Questions: why is it an array? and not just 
```js
return css`
  h2 { color: orange; }
`;
```

using only a css tagged template literal you can do more advanced stuff like duplicate css.

e.g.
```js
// somewhere
const headerColor = css`h2 { color: orange; }`;

// myBase.js
static get styles() {
  return css`
     ${headerColor}
     .footer { border-top: 1px solid green; }
  `;
}

// myEl.js
static get styles() {
  return css`
     ${super.styles}
     ${headerColor}
     .some { color: yellow; }
  `;
}
```

**Output of myEl css**
```css
h2 { color: orange; }
.footer { border-top: 1px solid green; }
h2 { color: orange; }
.some { color: yellow; }
```

**Output of myEl css WITH dedupe**
```css
h2 { color: orange; }
.footer { border-top: 1px solid green; }
.some { color: yellow; }
```

I am telling this as we have been using a BaseClass that supports style deduping for a while now and this is a really powerful feature and it makes the live of the css parser/shadycss just simpler by not needing to process as much css.

In this simple example it's not much but really if you have huge css blocks then deduping css improves performance a lot.

So I am really interested what the reasoning for using an array are.

PS: lit-css already supports this see https://github.com/lit-styles/lit-styles/blob/master/packages/lit-css/test/lit-css.spec.js#L29

bashmish:
I'm happy to see that you folks take a step forward to support this. Would be nice if you take [lit-css](https://github.com/lit-styles/lit-styles/blob/master/packages/lit-css/README.md) as an inspiration for your implementation. I personally always thought this should be part of the LitElement core. I do understand that tagged templates are not designed for such a use case or at least this use case with deduping is uncommon, but so far there were some users happy with it, for example, @web-padawan who prototypes new [vaadin components](https://github.com/web-padawan/lit-components/blob/master/packages/button-lumo/lit-button-lumo.js).

bashmish:
Also some ideas behind such a design I presented in this talk a few weeks ago https://youtu.be/wNKyEtLRTe4?t=2373

Between the lines explain the motivation for lit-css and why the getter for styles in LitElement is static (we agree here anyway which is a good sign).

src/lib/updating-element.ts
```diff
@@ -215,6 +217,14 @@ export abstract class UpdatingElement extends HTMLElement {
 
   static properties: PropertyDeclarations = {};
 
+  /**
+   * Array of styles to apply to the element. The styles should be defined
+   * using the `css` tag function.
+   */
+  static get styles() {
```
justinfagnani: Add type of `Array<CSSStyleSheet|LiteralString>`

src/lib/css-tag.ts
```diff
@@ -0,0 +1,52 @@
...
+Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
+This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+Code distributed by Google as part of the polymer project is also
+subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+*/
+export const supportsAdoptedStyleSheets = ('adoptedStyleSheets' in Document.prototype);
+
+class LiteralString {
```
justinfagnani: I would name this `CSSLiteral`

src/lib/css-tag.ts
```diff
@@ -0,0 +1,52 @@
...
+  if (value instanceof LiteralString) {
+    return value.value;
+  } else {
+    throw new Error(
+        `non-literal value passed to 'css' function: ${value}`
+    );
+  }
+};
+
+export const css = (strings: string[], ...values: any[]) => {
```
justinfagnani: I think we should consider having this return an object with an optional `.styleSheet` and `.cssText` so that there's a consistent return type for the `css` tag and so the static style property has a single item type.

This would change below to just:

```ts
   const cssText = values.reduce((acc, v, idx) =>
       acc + literalValue(v) + strings[idx + 1], strings[0]);
   return new CSSResult(cssText);
```

src/lib/css-tag.ts
```diff
@@ -0,0 +1,52 @@
...
+    return this.value;
+  }
+}
+
+const literalValue = (value: LiteralString) => {
+  if (value instanceof LiteralString) {
+    return value.value;
+  } else {
+    throw new Error(
+        `non-literal value passed to 'css' function: ${value}`
```
justinfagnani: It'd be nice to allow user-provided values in safe contexts with context-aware auto-escaping. Not sure how far we can go purely client-side. We may be able to allow directive-like functions that return a sentinel value that parses correctly for a particular context, then replace it, lit-html style. Soy does interesting things here at compile time. Maybe @mikesamuel has ideas.

src/lib/updating-element.ts
```diff
@@ -462,7 +472,25 @@ export abstract class UpdatingElement extends HTMLElement {
...
+    } else if (supportsAdoptedStyleSheets) {
+      // TODO(sorvell): fix type
+      (shadowRoot as any).adoptedStyleSheets = styles;
+    } else {
+      this.requestUpdate();
+      this._updatePromise.then(() => {
+        styles.forEach((s) => {
+          const style = document.createElement('style');
+          style.textContent = s;
+          shadowRoot.appendChild(style);
```
justinfagnani: to match Constructible StyleSheet ordering these have to come first in the ShadowRoot. Is that happening because you're directly waiting `_updatePromise`? Comment if so.

src/lib/updating-element.ts
```diff
@@ -462,7 +472,25 @@ export abstract class UpdatingElement extends HTMLElement {
    * @returns {Element|DocumentFragment} Returns a node into which to render.
    */
   protected createRenderRoot(): Element|ShadowRoot {
-    return this.attachShadow({mode : 'open'});
+    const shadowRoot = this.attachShadow({mode : 'open'});
+    const styles = (this.constructor as typeof UpdatingElement).styles;
```
justinfagnani: I (weakly) feel like all this should be in LitElement, since it does the actual rendering.

src/lib/updating-element.ts
```diff
@@ -462,7 +472,25 @@ export abstract class UpdatingElement extends HTMLElement {
...
+    if (window.ShadyCSS !== undefined && !(window.ShadyCSS as any).nativeShadow) {
```
justinfagnani: comment the three cases a little

justinfagnani:
@daKmoR styles is an array because elements may be sharing styles, and in a Constrctible StyleSheet supporting browser the `css` tag will return an actual StyleSheet object. We don't want to (and can't easily in all cases) merge these into each other with interpolations because `adoptedStyleSheets` is an array - we want to pass a reference to each StyleSheet object individually.

A modified version of your example:

```ts
const sharedStyles = css`
  h2 { color: orange; }
`;

class Base extends LitElement {
  static get styles() {
    return [
      sharedStyles,
      css`
        .footer { border-top: 1px solid green; }
      `
    ];
  }
}

class El extends LitElement {
  static get styles() {
    return [
      ...super.styles,
      // you probably know this is included in the parent, so this is dubious
      // we can still dedupe the arrays before setting `adoptedStyleSheets` but that
      // can change the cascade
      sharedStyles,
      css`
        .some { color: yellow; }
      `
    ];
  }
}
```

b-strauss:
Is `createRenderRoot` the right place for this code?

Having a custom implementation for `createRenderRoot` for example [here](https://github.com/material-components/material-components-web-components/blob/fbae6bedd5d8fcf5657e0acb2764a970eec2003c/packages/button/src/mwc-button.ts#L48-L50) with custom options for `attachShadow` will require you to duplicate all the code.

daKmoR:
>    // you probably know this is included in the parent, so this is dubious

unfortunately you can not know that always... for example if you are a mixin.

So sometimes you have shared stylings for two mixins but each needs it to work on it's own. However you can also apply both mixins and then you will get dublicate css.

```js
const srOnlyClass = css`.sr-only { ... }`;

const DescriptionMixin = ... // applies srOnlyClass so
class foo extends DescriptionMixin(BaseElement) // works
const DetailsMixin = ... // applies srOnlyClass so
class foo extends DetailsMixin(BaseElement) // works

// now this will apply srOnlyClass 2 times
class bar extends DescriptionMixin(DetailsMixin(BaseElement)) {}
```

we have this actually actually quite often in our codebase that there is shared stylings.

> // we can still dedupe the arrays before setting `adoptedStyleSheets` but that
   // can change the cascade

true so for us we always import the (parent) stylings at the top... which worked really nicely for us.


So overall arrays should work as well... I do find it a little bit more cumbersome but if it's more future prove then it's worth it.

But deduping should still be done... and for a css tagged template it was nicely part of the `css` tag itself... for array we either have to do `return dedupeStyles([...super.styles, myStyleB])` or do "magic" with the array... 

PS: I am off work till next year so the above example I made up... if need be I could see if I can look up some of our real examples

LarsDenBakker:
I really like this direction! 

I think it's important to consider how to compose style modules:

```javascript
const typographyStyles = css`...`;
const buttonStyles = css`...`;

// a standalone style module
const styleModuleA = css`...`;
// a style module that composes with another
const styleModuleB = [typographyStyles, css`...`];

  static get styles() {
    return [
      typographyStyles,
      buttonStyles,
      styleModuleA,
      ...styleModuleB,
    ];
  }
}
```

I think this creates two challenges:
1. Deduplication: Like @daKmoR mentioned you don't want to include the same styles twice, and you don't want to need to know within `MyElement` how `styleModuleB` is constructed internally.
2. Composition: You don't want to have to know in `MyElement` which style modules are arrays, and it exposes an extra implementation detail. Perhaps the same approach as Polymer 1 behaviors can be used here, where the styles array is flattened before composition. On modern browsers Array.prototype.flat is enough.

src/lib/updating-element.ts
sorvell: Added specific type for this and used it here.

src/lib/css-tag.ts
sorvell: Done.

src/lib/css-tag.ts
sorvell: Done.

src/lib/updating-element.ts
sorvell: Yup, comment added. Note, they come last in the `shadowRoot`.

src/lib/updating-element.ts
sorvell: It seems like the `shadowRoot` rendering (creation + styles) should go together, but I think it's reasonable that this go into `LitElement` and not `UpdatingElement`. Let's discuss separately.

src/lib/updating-element.ts

src/lib/updating-element.ts
justinfagnani: Don't they have to come first in case there are `<style>` tags in the rest of the shadow content?

src/lib/css-tag.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export const css = (strings: TemplateStringsArray, ...values: CSSLiteral[]): CSSStyleSheetOrCssText => {
+  const cssText = values.reduce((acc, v, idx) =>
+      acc + cssLiteralValue(v) + strings[idx + 1], strings[0]);
+  const result: CSSStyleSheetOrCssText = {};
+  if (supportsAdoptedStyleSheets) {
+    result.styleSheet = new CSSStyleSheet() as ConstructableStyleSheet;
+    result.styleSheet.replaceSync(cssText);
+  } else {
+    result.cssText = new CSSLiteral(cssText);
```
justinfagnani: you can always assign this field

justinfagnani:
> Not only that, but just disabling shadowDom by returning this in createRenderRoot, as recommended in documentation, will break styles.

That's just going to be true of any styles. If you have styles that assume ShadowDOM scoping, they're going to do the wrong thing if you don't render to a ShadowRoot. In this case at least nothing will happen with the styles unless you do something yourself, which is probably better than them being applied globally by default.

src/lib/css-tag.ts
```diff
@@ -0,0 +1,61 @@
...
+export const supportsAdoptedStyleSheets = ('adoptedStyleSheets' in Document.prototype);
+
+interface ConstructableStyleSheet extends CSSStyleSheet {
+  replaceSync(cssText: string): void;
+  replace(cssText: string): Promise<unknown>;
+}
+
+class CSSLiteral {
+
+  value: string;
```
justinfagnani: `readonly`

src/lib/css-tag.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  constructor(value: string) {
+    this.value = value.toString();
```
justinfagnani: Why the `.toString()`? If `value` can be something other than a string, update the type.

src/lib/updating-element.ts
```diff
@@ -462,7 +483,48 @@ export abstract class UpdatingElement extends HTMLElement {
...
+    this.createRenderRootStyles(shadowRoot);
+    return shadowRoot;
+  }
+
+  /**
+   * Applies styling to the element shadowRoot using the `static get styles`
+   * property. Styling will apply using `shadowRoot.adoptedStyleSheets` where
+   * available and will fallback otherwise.
+   */
+  protected createRenderRootStyles(shadowRoot: ShadowRoot) {
```
justinfagnani: The name is too generic, since this doesn't work with any render root, but only ShadowRoots. How about `adoptStyles` or `applyStyles`?

src/lib/updating-element.ts
```diff
@@ -462,7 +483,48 @@ export abstract class UpdatingElement extends HTMLElement {
...
+    let styles = (this.constructor as typeof UpdatingElement).styles;
+    // de-duplicate styles preserving the last item in the list.
```
justinfagnani: Add why this is important

src/lib/updating-element.ts
```diff
@@ -462,7 +483,48 @@ export abstract class UpdatingElement extends HTMLElement {
...
+    const stylesSet = new Set();
```
justinfagnani: Not sure if this is better, but it gets rid of the deletes (which involve hashes) in exchange for a reverse:

```js
const stylesSet = new Set();
for (const i = styles.length - 1; i >= 0; i==) {
  s.add(styles[i]);
}
styles = [...styleSet].reverse();
```

or:

```js
styles = [...new Set([...styles].reverse())].reverse();
```

src/lib/updating-element.ts
justinfagnani: Ok, I see: https://wicg.github.io/construct-stylesheets/#using-constructed-stylesheets Let's include a link to the spec

src/lib/updating-element.ts
```diff
@@ -462,7 +483,48 @@ export abstract class UpdatingElement extends HTMLElement {
...
-    return this.attachShadow({mode : 'open'});
+    const shadowRoot = this.attachShadow({mode : 'open'});
+    this.createRenderRootStyles(shadowRoot);
+    return shadowRoot;
+  }
+
+  /**
+   * Applies styling to the element shadowRoot using the `static get styles`
+   * property. Styling will apply using `shadowRoot.adoptedStyleSheets` where
+   * available and will fallback otherwise.
```
justinfagnani: Let's describe the fallback. Specifically that it appends `<style>` elements to the ShadowRoot after the first render. This will be important for other rendering libraries.

src/lib/updating-element.ts
```diff
@@ -11,6 +11,19 @@
  * subject to an additional IP rights grant found at
  * http://polymer.github.io/PATENTS.txt
  */
+import {supportsAdoptedStyleSheets, CSSStyleSheetOrCssText} from './css-tag.js';
+export * from './css-tag.js';
+
+// Augment existing types with bleeding edge API.
+declare global {
```
justinfagnani: I think we may want these here: https://github.com/Polymer/lit-element/blob/master/src/env.d.ts

src/lib/css-tag.ts
```diff
@@ -0,0 +1,61 @@
...
+Copyright (c) 2017 The Polymer Project Authors. All rights reserved.
+This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+Code distributed by Google as part of the polymer project is also
+subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+*/
+export const supportsAdoptedStyleSheets = ('adoptedStyleSheets' in Document.prototype);
+
+interface ConstructableStyleSheet extends CSSStyleSheet {
```
justinfagnani: According to https://github.com/w3c/csswg-drafts/issues/3433 these additions are going to be merged into `CSSStyleSheet`, so add them globally as with `adoptedStyleSheets`.

src/lib/css-tag.ts
```diff
@@ -0,0 +1,61 @@
...
+  if (value instanceof CSSLiteral) {
+    return value.value;
+  } else {
+    throw new Error(
+        `Non-literal value passed to 'css' function: ${value}.`
+    );
+  }
+};
+
+export type CSSStyleSheetOrCssText = {
```
justinfagnani: I don't think `Or` is the best naming anymore. It's really specific to this operation, so something like `CSSResult` seems fine.

src/lib/updating-element.ts
```diff
@@ -462,7 +483,48 @@ export abstract class UpdatingElement extends HTMLElement {
...
+    return shadowRoot;
+  }
+
+  /**
+   * Applies styling to the element shadowRoot using the `static get styles`
+   * property. Styling will apply using `shadowRoot.adoptedStyleSheets` where
+   * available and will fallback otherwise.
+   */
+  protected createRenderRootStyles(shadowRoot: ShadowRoot) {
+    let styles = (this.constructor as typeof UpdatingElement).styles;
```
blikblum: Check if styles is undefined and return to avoid extra works on elements that don't declare styles?

src/lib/css-tag.ts
sorvell: Done.

src/lib/css-tag.ts
sorvell: Removed.

src/lib/updating-element.ts
sorvell: Changed to `adoptStyles`

src/lib/updating-element.ts
sorvell: Added note, for performance.

src/lib/updating-element.ts
sorvell: Used `reduceRight` + `reverse`

src/lib/updating-element.ts
sorvell: Added.

src/lib/updating-element.ts
sorvell: Added types to `env.d.ts`

src/lib/css-tag.ts
sorvell: Done.

src/lib/css-tag.ts
sorvell: Changed.

src/lib/updating-element.ts
sorvell: Thanks, done.

src/lib/updating-element.ts
```diff
@@ -415,6 +444,12 @@ export abstract class UpdatingElement extends HTMLElement {
    */
   protected initialize() {
     this.renderRoot = this.createRenderRoot();
+    // Note, tf renderRoot is not a shadowRoot, styles would/could apply to the
```
kenchris: "tf" should be "if"

src/lib/css-tag.ts
```diff
@@ -0,0 +1,50 @@
...
+  constructor(public readonly cssText: string) {
+  }
+
+  // Note, this is a getter so that it's lazy. In practice, this means stylesheets
+  // are not created until the first element instance is made.
+  get styleSheet(): CSSStyleSheet|null {
+    if (this._styleSheet === undefined) {
+      // Note, assume that if adoptedStyleSheets is supported,
+      // constructable stylesheets are as well.
+      if (supportsAdoptedStyleSheets) {
```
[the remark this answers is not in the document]
sorvell: Changed to `supportsAdoptingStyleSheets` and added explicit check for construct-able stylesheet support.

src/lib/updating-element.ts
```diff
@@ -465,6 +500,42 @@ export abstract class UpdatingElement extends HTMLElement {
...
+   * Applies styling to the element shadowRoot using the `static get styles`
+   * property. Styling will apply using `shadowRoot.adoptedStyleSheets` where
+   * available and will fallback otherwise. When Shadow DOM is polyfilled,
+   * ShadyCSS scopes styles and adds them to the document. When Shadow DOM
+   * is available but `adoptedStyleSheets` is not, styles are appended to the
+   * end of the `shadowRoot` to [mimic spec behavior](https://wicg.github.io/construct-stylesheets/#using-constructed-stylesheets).
+   */
+  protected adoptStyles() {
+    const styles = (this.constructor as typeof UpdatingElement)._uniqueStyles;
+    if (!styles.length) {
```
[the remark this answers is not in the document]

src/lib/css-tag.ts
```diff
@@ -0,0 +1,50 @@
...
+Code distributed by Google as part of the polymer project is also
+subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+*/
+export const supportsAdoptedStyleSheets = ('adoptedStyleSheets' in Document.prototype);
+
+export class CSSResult {
+
+  _styleSheet?: CSSStyleSheet|null;
+
+  constructor(public readonly cssText: string) {
```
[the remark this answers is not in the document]
sorvell: Done.

src/lib/updating-element.ts

src/lib/updating-element.ts
```diff
@@ -415,6 +444,12 @@ export abstract class UpdatingElement extends HTMLElement {
    */
   protected initialize() {
     this.renderRoot = this.createRenderRoot();
+    // Note, if renderRoot is not a shadowRoot, styles would/could apply to the
+    // element's getRootNode(). While this could be done, we're choosing not to
+    // support this now since it would require different logic around de-duping.
+    if (this.renderRoot instanceof window.ShadowRoot) {
```
blikblum: Will fail in a env that does not supports shadow dom and not is pollyfied. Such env would be supported at all?

src/lib/updating-element.ts
```diff
@@ -215,6 +217,33 @@ export abstract class UpdatingElement extends HTMLElement {
...
+
+  private static get _uniqueStyles(): CSSResult[] {
+    if (this._styles === undefined) {
+      const styles = this.styles;
+      // As a performance optimization to avoid duplicated styling that can
+      // occur especially when composing via subclassing, de-duplicate styles
+      // preserving the last item in the list. The last item is kept to
+      // try to preserve cascade order with the assumption that it's most
+      // important that last added styles override previous styles.
+      const styleSet = styles.reduceRight((set, s) => {
```
kevinpschaaf: can just be `styles.reduceRight((set, s) => set.add(s), new Set())`

(set.add returns the set)

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 
 ### Added
+* Added `static get styles()` to allow defining element styling separate from `render` method.
+This takes advantage of [`adoptedStyleSheets`](https://wicg.github.io/construct-stylesheets/#using-constructed-stylesheets) when possible ([#391](https://github.com/Polymer/lit-element/issues/391)).
 * Added the `performUpdate` method to allow control of update timing ([#290](https://github.com/Polymer/lit-element/issues/290)).
 * Updates deferred until first connection ([#258](https://github.com/Polymer/lit-element/issues/258)).
 ### Changed
```
kevinpschaaf: Add `[breaking]` change notice about moving `createRenderRoot` and related functionality from `UpdatingElement` to `LitElement`

src/lit-element.ts
```diff
@@ -31,6 +33,112 @@ export class LitElement extends UpdatingElement {
...
+  protected adoptStyles() {
+    const styles = (this.constructor as typeof LitElement)._uniqueStyles;
+    if (styles.length === 0) {
+      return;
+    }
+    // There are three separate cases here based on Shadow DOM support.
+    // (1) shadowRoot polyfilled: use ShadyCSS
+    // (2) shadowRoot.adoptedStyleSheets available: use it.
+    // (3) shadowRoot.adoptedStyleSheets polyfilled: append styles after rendering.
+    if (window.ShadyCSS !== undefined && !window.ShadyCSS.nativeShadow) {
```
kevinpschaaf: Is it worth moving more of the prep steps to the static one-time spot (`_uniqueStyles`)?  Basically, move the `map` calls out of instance-time work, and possibly also create a template of styles to clone for the final case?

src/lib/updating-element.ts
sorvell: Thanks, fixed.

src/lib/css-tag.ts
```diff
@@ -0,0 +1,62 @@
...
+  try {
+    new CSSStyleSheet();
+  } catch {
+    constructable = false;
+  }
+  return constructable;
+})();
+
+export const supportsAdoptingStyleSheets = supportsConstructableStyleSheets &&
+    ('adoptedStyleSheets' in Document.prototype);
```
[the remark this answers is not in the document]
sorvell: Done.

src/lib/updating-element.ts
```diff
@@ -465,6 +500,42 @@ export abstract class UpdatingElement extends HTMLElement {
...
+    if (window.ShadyCSS !== undefined && !window.ShadyCSS.nativeShadow) {
+      window.ShadyCSS.prepareAdoptedCssText(styles.map((s) => s.cssText), this.localName);
+    } else if (supportsAdoptingStyleSheets) {
+      (this.renderRoot as ShadowRoot).adoptedStyleSheets = styles.map((s) => s.styleSheet!);
+    } else {
+      // Ensure styling comes after rendering so styles are *after* all other rendered content.
+      // This matches the spec'd behavior of `adoptedStyleSheets`.
+      // Ensure an updated is requested and then render styling directly after the update.
+      this.requestUpdate();
+      this._updatePromise.then(() => {
```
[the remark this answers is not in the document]
sorvell: Since `part` is cached off the container, avoiding this. However, moved this case to `update` to avoid the fancy waiting here.

src/lib/updating-element.ts

CHANGELOG.md
sorvell: Done.

motss:
`<style>` tags no longer showing in DevTools on Chrome 73. Is that expected behavior with native support of Constructible Stylesheet?

PR #422, opened 2019-01-09
Support new decorators proposal as implemented in Babel 7

### LINKED ISSUE #156: Make decorators work with new decorator spec
Babel 7 has an implementation of the new spec. We should be able to make decorators detect how they're being called to work with either TypeScript or Babel 7.

justinfagnani:
* Moves @property tests to decorators_test.ts
* Adds a Babel build step to compile decorators_test.ts to decorators-babel_test.js
* Updates decorators to detect if they’re called in legacy or modern decorator mode

Fixes #156


justinfagnani:
cc @katejeffreys with this change we should stop referring to the decorators as "TypeScript decorators". I'll add commits to change what I can to this PR. You can push to this branch too.

justinfagnani:
Chrome 41 is failing due to https://github.com/babel/babel/issues/9234. IE11 appears to be different. Manually running though the other browsers to see what's up.

src/lib/decorators.ts
```diff
@@ -41,24 +41,73 @@ export type Constructor<T> = {
...
+export const property = (options?: PropertyDeclaration) =>
+    (protoOrDescriptor: Object|ClassElement, name?: PropertyKey): any => {
+      if (name !== undefined) {
+        // Legacy decorator
+        (protoOrDescriptor.constructor as typeof UpdatingElement)
+            .createProperty(name!, options);
+        return;
+      }
+      const element = protoOrDescriptor as ClassElement;
+      console.assert(element.kind === 'field');
```
rictic: Is console.assert what we want here? It only logs, it does not throw or otherwise stop execution.

src/lib/decorators.ts
```diff
@@ -41,24 +41,73 @@ export type Constructor<T> = {
...
 export const customElement = (tagName: string) =>
-    (clazz: Constructor<HTMLElement>) => {
-      window.customElements.define(tagName, clazz);
-      // Cast as any because TS doesn't recognize the return type as being a
-      // subtype of the decorated class when clazz is typed as
-      // `Constructor<HTMLElement>` for some reason. `Constructor<HTMLElement>`
-      // is helpful to make sure the decorator is applied to elements however.
-      return clazz as any;
+    (classOrDescriptor: Constructor<HTMLElement>|ClassDescriptor) => {
+      if (typeof classOrDescriptor === 'function') {
```
rictic: Can we separate out the legacy decorators from the standard ones in separate functions?

It'd make the code easier to read, validate, and type check. Can still export only one version, like:

```
export const customElement = (tagName: string) =>
  (classOrDescriptor: Constructor<HTMLElement>|ClassDescriptor) => {
    if (typeof classOrDescriptor === 'function') return experimentalTsCustomElement(classOrDescriptor);
    return standardCustomElement(classOrDescriptor);
  );
```

justinfagnani:
@rictic 
> decorators-babel_test.js looks like it wasn't uploaded? I don't see it in the diffs

This is a file generated by the build:babel-test npm script.

src/lib/decorators.ts
```diff
@@ -41,24 +41,73 @@ export type Constructor<T> = {
...
+        return;
+      }
+      const element = protoOrDescriptor as ClassElement;
+      console.assert(element.kind === 'field');
+      // createProperty() takes care of defining the property, but we still must
+      // return some kind of descriptor, so return a descriptor for an unused
+      // prototype field. The finisher calls createProperty().
+      return {
+        kind : 'field',
+        key : Symbol(),
```
sorvell: It seems unfortunate that we create these fake properties, but I guess this is the only way we can get the behavior we want.

blikblum: This version works: https://github.com/Polymer/lit-element/issues/205#issuecomment-427719522

AFAICS the difference is that the setter will not be called in instance initialization

src/env.d.ts
```diff
@@ -11,3 +11,19 @@ interface Window {
...
+interface ClassDescriptor {
+  kind: 'class';
+  elements: ClassElement[];
+}
+
+// From the TC39 Decorators proposal
+interface ClassElement {
+  kind: 'field'|'method';
+  key: PropertyKey;
+  placement: 'static'|'prototype'|'own';
```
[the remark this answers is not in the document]
kevinpschaaf: Done.

src/env.d.ts
```diff
@@ -11,3 +11,19 @@ interface Window {
   ShadyCSS?: ShadyCSS;
   ShadyDOM?: ShadyDOM;
 }
+
+// From the TC39 Decorators proposal
+interface ClassDescriptor {
+  kind: 'class';
+  elements: ClassElement[];
```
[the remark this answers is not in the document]
kevinpschaaf: Done

src/lib/decorators.ts
kevinpschaaf: Removed

src/lib/decorators.ts
kevinpschaaf: Done

src/lib/decorators.ts
```diff
@@ -117,7 +178,18 @@ function _query<T>(queryFn: (target: NodeSelector, selector: string) => T) {
...
 export const eventOptions = (options: AddEventListenerOptions) =>
-    (proto: any, name: string) => {
-      // This comment is here to fix a disagreement between formatter and linter
-      Object.assign(proto[name], options);
+    (protoOrDescriptor: any, name?: string) => {
+      if (name !== undefined) {
+        // Legacy decorator
+        Object.assign(protoOrDescriptor[name], options);
+      } else {
+        return {
```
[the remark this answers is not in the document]
kevinpschaaf: Pass

src/env.d.ts
```diff
@@ -11,3 +11,19 @@ interface Window {
...
+}
+
+// From the TC39 Decorators proposal
+interface ClassElement {
```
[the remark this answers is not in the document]
kevinpschaaf: Done

PR #427, opened 2019-01-11
Fix references to @polymer/lit-element in docs and README

azakus:


docs/_includes/projects/old/clone/manifest.json
```diff
@@ -8,7 +8,7 @@
     "README.md"
   ],
   "dependencies": {
-    "@polymer/lit-element": "0.5.2",
+    "lit-element": "0.5.2",
```
sorvell: This change isn't right. Why is there an "old" folder at all?

docs/_includes/projects/old/declare/manifest.json
```diff
@@ -8,7 +8,7 @@
     "README.md"
   ],
   "dependencies": {
-    "@polymer/lit-element": "0.5.2",
+    "lit-element": "0.5.2",
```
sorvell: Same as above and a few more cases below.

docs/_includes/projects/old/clone/manifest.json
sorvell: Removed.

docs/_includes/projects/old/declare/manifest.json
sorvell: Removed.

Issue #428, opened 2019-01-11
[bug] Cannot find name 'ClassDescriptor' (v0.7.0)


lacolaco:
### Description

When I use decorators of LitElement, TypeScript compiler emits errors.

```
ERROR in [at-loader] ./node_modules/@polymer/lit-element/lib/decorators.d.ts:37:105
    TS2304: Cannot find name 'ClassDescriptor'.

ERROR in [at-loader] ./node_modules/@polymer/lit-element/lib/decorators.d.ts:43:118
    TS2304: Cannot find name 'ClassElement'.

ERROR in [at-loader] ./node_modules/@polymer/lit-element/lib/decorators.d.ts:48:80
    TS2304: Cannot find name 'ClassElement'.

ERROR in [at-loader] ./node_modules/@polymer/lit-element/lib/decorators.d.ts:53:83
    TS2304: Cannot find name 'ClassElement'.
```

It looks a bug. `env.d.ts` is only accessible from LitElement source code.
https://github.com/Polymer/lit-element/blob/master/src/env.d.ts

#### Live Demo

https://stackblitz.com/edit/lit-element-example?file=index.js

https://glitch.com/edit/#!/hello-lit-element?path=index.html:1:0

#### Steps to Reproduce

1. Install v0.7.0
2. Use `@property()` in TypeScript 
3. Execute `tsc`

#### Expected Results

No errors.

#### Actual Results
Error is thrown

### Browsers Affected

- [ ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari 11
- [ ] Safari 10
- [ ] IE 11

### Versions

- lit-element: v0.7.0
- webcomponents: vX.X.X

motss:
FWIW, there is also another error.

```js
error TS2304: Cannod find name 'Constructor'.
```

PR #431, opened 2019-01-11
Fix decorator types causing compiler errors for TypeScript users.

justinfagnani:
Fixes: #428 

Verified by pushing 2.0.0-dev.ff4fefd to npm and building PWA Starter Kit


abdonrd:
Great! Thanks @justinfagnani!

Issue #429, opened 2019-01-11
Style modules API


LarsDenBakker:
Now that there is a new style system, I think it's good to consider how to create shared style modules. In Polymer we used to do this:

```html
<!-- file: style-module-a.html -->
<dom-module id="style-module-a">
  <template>
    <style>/* ... */</style>
  </template>
</dom-module>

<!-- file: style-module-b.html -->
<dom-module id="style-module-b">
  <template>
    <style include="style-module-a">/* ... */</style>
  </template>
</dom-module>
```

This way you could use `style-module-b` without knowing that it also composes `style-module-a`.

With the new `LitElement` style system, it would translate to this:

```javascript
// file: style-module-a.js
export const styleModuleA = css`
  ...
`;

// file: style-module-b.js
import { styleModuleA } from './style-module-a.js';

export const styleModuleB = [
  styleModuleA,
  css`
    ...
  `;
];
```

This is a bit awkward, as the consumer of `style-module-b` now has to know that it is an array of styles and has to spread it:

```javascript
class MyElement extends LitElement {
  static get styles() {
    return [
      ..styleModuleB,
      styleModuleC
    ];
  }
}
```

But this also means that `styleModuleB` could never change back to not being an array, and `styleModuleC` can never change to being an array of styles. 

We could solve this by making style modules always an array, even if there is only one entry. Another solution would be if `LitElement` flattened the styles array so that you don't need to worry about this.

sorvell:
The reason that `static get styles` must return an array to make composition easier.

I think it's reasonable to flatten the array of styles and we'd accept a PR for that.

justinfagnani:
Why do we need to support this when we have array spread syntax? It's prefer to not add code for things the language already does for us.

justinfagnani:
One thing that would make flattening worth it to me is if we changed the `styles` property to allow for a single CssResult. Then merging in super-class styles would always be `[super.styles, css`...`]` and writing styles would be simpler for the single-stylesheet case.

hyperpress:
I agree with @justinfagnani last comment. That provides a good relief solution for those who really want/need a single stylesheet method.

LarsDenBakker:
Yes definitely make sense to support single `CSSResult` values.

PR #437, opened 2019-01-11
Flatten styles array

LarsDenBakker:
Fixes https://github.com/Polymer/lit-element/issues/429

Flattens styles array before deduplication. 

Recursive types are a bit funky, I had to use an interface as instructed here: https://github.com/unional/typescript-guidelines/blob/master/pages/advance-types/recursive-types.md

I'm using Array.prototype.flat when supported, otherwise there is a minimal implementation. 

I had to enable esnext typings for Array.prototype.flat support.

### DIFF
--- a/src/lit-element.ts
+++ b/src/lit-element.ts
@@ -22,14 +22,14 @@ export {html, svg, TemplateResult, SVGTemplateResult} from 'lit-html/lit-html';
 import {supportsAdoptingStyleSheets, CSSResult} from './lib/css-tag.js';
 export * from './lib/css-tag.js';
 
-export interface LitElementStyles extends Array<CSSResult | CSSResult[] | LitElementStyles> {}
+export interface CSSResultArray extends Array<CSSResult | CSSResultArray> {}
 
 /**
  * Minimal implementation of Array.prototype.flat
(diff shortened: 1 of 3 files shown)


justinfagnani:
See my comments on #429

src/lit-element.ts
```diff
@@ -22,6 +22,28 @@ export {html, svg, TemplateResult, SVGTemplateResult} from 'lit-html/lit-html';
 import {supportsAdoptingStyleSheets, CSSResult} from './lib/css-tag.js';
 export * from './lib/css-tag.js';
 
+export interface LitElementStyles extends Array<CSSResult | CSSResult[] | LitElementStyles> {}
```
sorvell: Let's also allow this to be a single `CSSResult`. This way `static get styles` can be a single `CSSResult` and then any form of composition (sharing some "style modules" or some super class styles) is similar, e.g. `static get styles() { return [super.styles, sharedStyles]; }`

If we do this, then `flattenStyles` needs to handle the case where the input is a non-array and we'll obviously need to test that.

LarsDenBakker: Good suggestion, added it in :)

PR #443, opened 2019-01-11
Rebuild API docs

katejeffreys:
Staged: https://20190111t151039-dot-polymer-lit-element.appspot.com


katejeffreys:
I added the install command when I added the import statement because installing and importing generally go together, y/n? Maybe people wouldn't expect or need install info in API docs tho.. @arthurevans WDYT?

PR #454, opened 2019-01-17
Removes wrapping of properties in favor of always respecting `noAccessor`

### LINKED ISSUE #450: Extending does some "magic" for properties which reduces control
### Description
Why do we have this magic of "auto" merging `static get properties()` while extending? Why not use a simple ` ...super.properties`?

Doing this "magic" brings the following downsides:

1) When applying mixins that each have the same property it will execute requestUpdate for each definition.  
Use-case would be Mixins that work independently but can also be combined.
```js
const MixinA = superclass => class MixinA extends superclass {
  static get properties() { return { name: { type: String }, }; }
}

const MixinB = superclass => class MixinB extends superclass {
  static get properties() { return { name: { type: String }, }; }
}

class CustomGreeting extends MixinA(LitElement) {
// =>will execute 1 time

class CustomGreeting extends MixinA(MixinB(LitElement)) { 
// =>will execute 2 times

class CustomGreeting extends MixinA(MixinB(LitElement)) { 
  static get properties() { return { name: { type: String }, }; }
}
// =>will execute 3 times
```

2) When extending an element and using `...super.properties`  it will execute requestUpdate for each definition.  
Use-case would be changing the property type on the extending element.
```js
class MyBase extends LitElement {
  static get properties() { return { name: { type: String }, }; }
}

class ExtendGreeting extends MyBase {
  static get properties() { return { 
      ...super.properties,
      name: { type: Number},
};}
```

#### Live Demo
https://stackblitz.com/edit/lit-element-request-update-bug?file=index.html

#### Expected Results
One call to requestUpdate for changing properties.

#### Actual Results
Multiple calls to requestUpdate for each `static get properties` with the same name in the "extending tree".

### Browsers Affected
- [x] Chrome

### Versions
- lit-element: v2.0.0-rc.2

sorvell:
Fixes #450. This removes all wrapping of existing accessors. This changes makes property generation simpler and more straightforward. If an accessor should not be generated, for example when there is a user defined accessor, users should set the `noAccessor` flag to true.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 * Fix jsfiddle reference.
 
 ### Changed
-* [Breaking] Property accessors are no longer wrapped when they already exist. Instead the `noAccessor` flag should be set when a user defined accessor is created. ([#450](https://github.com/Polymer/lit-element/issues/450)).
+* [Breaking] Property accessors are no longer wrapped when they already exist. Instead the `noAccessor` flag should be set when a user-defined accessor exists on the prototype (and in this case, user-defined accessors must call `requestUpdate` themselves). ([#450](https://github.com/Polymer/lit-element/issues/450)).
 
 ## [2.0.0-rc.2] - 2019-01-11
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 * Fix jsfiddle reference.
 
+### Changed
+* [Breaking] Property accessors are no longer wrapped when they already exist. Instead the `noAccessor` flag should be set when a user defined accessor is created. ([#450](https://github.com/Polymer/lit-element/issues/450)).
+
 ## [2.0.0-rc.2] - 2019-01-11
 ### Fixed
 * Fix references to `@polymer/lit-element` in README and docs ([#427](https://github.com/Polymer/lit-element/pull/427)).
--- a/src/lib/decorators.ts
+++ b/src/lib/decorators.ts
@@ -91,9 +91,21 @@ export const customElement = (tagName: string) => (
 
 const standardProperty =
     (options: PropertyDeclaration, element: ClassElement) => {
-      // createProperty() takes care of defining the property, but we still must
-      // return some kind of descriptor, so return a descriptor for an unused
-      // prototype field. The finisher calls createProperty().
+      // When decorating an accessor, pass it through and add property metadata.
+      // Note, the `hasOwnProperty` check in `createProperty` ensures we don't
+      // stomp over the user's accessor.
+      if (element.kind === 'method' && element.descriptor &&
+          !('value' in element.descriptor)) {
+        return {
+          ...element,
+          finisher(clazz: typeof UpdatingElement) {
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -18,6 +18,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 * Fix jsfiddle reference.
 
+### Changed
+* [Breaking] Property accessors are no longer wrapped when they already exist. Instead the `noAccessor` flag should be set when a user defined accessor is created. ([#450](https://github.com/Polymer/lit-element/issues/450)).
```
kevinpschaaf:
```suggestion
* [Breaking] Property accessors are no longer wrapped when they already exist. Instead the `noAccessor` flag should be set when a user-defined accessor exists on the prototype (and in this case, user-defined accessors must call `requestUpdate` themselves). ([#450](https://github.com/Polymer/lit-element/issues/450)).
```

justinfagnani: I think you can remove the "`noAccessor` flag should be set" part

src/lib/decorators.ts
```diff
@@ -91,9 +91,21 @@ export const customElement = (tagName: string) => (
...
+      // stomp over the user's accessor.
+      if (element.kind === 'method' && element.descriptor &&
+          !('value' in element.descriptor)) {
+        return {
+          ...element,
+          finisher(clazz: typeof UpdatingElement) {
+            clazz.createProperty(element.key, options);
+          }
+        };
+        // createProperty() takes care of defining the property, but we still
```
justinfagnani: move comment to below the `else`

src/lib/updating-element.ts
```diff
@@ -288,35 +270,18 @@ export abstract class UpdatingElement extends HTMLElement {
...
-          set(value: any) {
-            const oldValue = this[name];
-            set.call(this, value);
-            this.requestUpdate(name, oldValue);
-          },
-          configurable : true,
-          enumerable : true
-        };
-      } else {
+    if (!options.noAccessor && !this.prototype.hasOwnProperty(name)) {
```
justinfagnani: I think an early return would read nicer here

src/lib/updating-element.ts
```diff
@@ -288,35 +270,18 @@ export abstract class UpdatingElement extends HTMLElement {
...
         const key = typeof name === 'symbol' ? Symbol() : `__${name}`;
-        desc = {
-          get() { return this[key]; },
-          set(value: any) {
-            const oldValue = this[name];
-            this[key] = value;
+      const desc = {
```
justinfagnani: you can inline this into the defineProperty call now

src/lib/updating-element.ts
kevinpschaaf:
```suggestion
    // Do not generate an accessor if the prototype already has one, since
    // it would be lost otherwise and that would never be the user's intention;
    // Instead, we expect users to call `requestUpdate` themselves from
    // user-defined accessors. Note that if the super has an accessor we will
    // still overwrite it
    if (!options.noAccessor && !this.prototype.hasOwnProperty(name)) {
```

CHANGELOG.md
sorvell: Added content @kevinpschaaf suggested.

src/lib/updating-element.ts

src/lib/updating-element.ts

src/lib/decorators.ts

PR #456, opened 2019-01-17
Removes default for `static get styles`

### LINKED ISSUE #452: Can't use static property for styles in TypeScript
TypeScript compiles properties to assignment (for now). The means that a subclass of a class that has `static get styles() {...}` can't use TypeScript static properties to set styles. The ergonomics of static properties are nice enough that we should allow for them with our base classes:

```ts
  static styles = css`
    :host {
      display: block;
    }
  `;
```

I suggest we remove the styles getter from LitElement, since it only returns an empty array. If we _really_ needed to keep it, we could just do:

```ts
LitElement.styles = [];
```

after the class definition.

sorvell:
Fixes #452.

In addition...
* Fixes type settings so `Array.flat` works.
* Optimization: ensures `this.styles` is accessed only once per definition. Since this getter generates CSSResults, it should not be run superfluously.

### DIFF
--- a/src/lit-element.ts
+++ b/src/lit-element.ts
@@ -66,14 +66,15 @@ export class LitElement extends UpdatingElement {
    * Array of styles to apply to the element. The styles should be defined
    * using the `css` tag function.
    */
-  static get styles(): CSSResult | CSSResultArray { return []; }
+  static styles?: CSSResult | CSSResultArray;
 
   private static _styles: CSSResult[]|undefined;
 
   private static get _uniqueStyles(): CSSResult[] {
     if (this._styles === undefined) {
-      if (Array.isArray(this.styles)) {
-        const styles = flattenStyles(this.styles);
+      const userStyles = this.styles;
+      if (Array.isArray(userStyles)) {
+        const styles = flattenStyles(userStyles);
         // As a performance optimization to avoid duplicated styling that can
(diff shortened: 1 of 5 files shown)


src/lit-element.ts
```diff
@@ -66,14 +66,15 @@ export class LitElement extends UpdatingElement {
    * Array of styles to apply to the element. The styles should be defined
    * using the `css` tag function.
    */
-  static get styles(): CSSResult | CSSResultArray { return []; }
+  static styles?: CSSResult | CSSResultArray;
 
   private static _styles: CSSResult[]|undefined;
 
   private static get _uniqueStyles(): CSSResult[] {
     if (this._styles === undefined) {
```
kevinpschaaf: Needs to be `hasOwnProperty` for subclassing to work.

sorvell: Done.

src/lit-element.ts
```diff
@@ -66,14 +66,15 @@ export class LitElement extends UpdatingElement {
...
-      if (Array.isArray(this.styles)) {
-        const styles = flattenStyles(this.styles);
+      const userStyles = this.styles;
+      if (Array.isArray(userStyles)) {
+        const styles = flattenStyles(userStyles);
```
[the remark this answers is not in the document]
sorvell: Done.

Issue #353, opened 2018-12-11
[docs] Review CSS Styles page for technical accuracy


katejeffreys:
Review the content at https://lit-element.polymer-project.org/guide/styles for technical accuracy only.

Markdown source: https://github.com/Polymer/lit-element/blob/master/docs/_guide/styles.md

* [ ] fix `:host` vs custom element tag
* [ ] fix/remove info on JS expressions in `<style></style>`

frankiefu:
https://lit-element.polymer-project.org/guide/styles#use-the-custom-element-tag-as-a-selector
`The :host CSS pseudo-class has higher specificity than the element’s type selector` is incorrect. I think it should be the other way around. See this [jsbin](https://jsbin.com/dujugun/edit?html,output)

keanulee:
"Use JavaScript expressions in LitElement style blocks"

`<style>${this.myStyles}</style>` is considered an anti-pattern since the browser has to re-parse the entire style tag if the textContent changes. Suggested pattern would be to make `this.myStyles` a complete style tag instead if you really want to call out styling with JS expression (which IMO could be skipped).

johnthad:
@keanulee What if `style` is a constant? As in
```
  const style = `font-family: ${label.font.family};...
  this.checkbox = html`
    <span aria-checked="${this.ariaChecked}">
      <label style="${style}">...
...
render() {
  this.setAttribute('role', 'checkbox');
  return html`
    ${HostStyle} ${this.checkbox}
  `;
}
```
If `style` never changes, it shouldn't have to rerender. Yes?

keanulee:
Looking at your snippet, there's a limitation on what actually works with JS expressions in style which might make the above code not behave as expected.

- When using LitElement with Shady DOM (the shim for when there's no native Shadow DOM), the `<style>` tags returned from `render()` are only processed _once per element class_ (not instance). So you can't have multiple instances of the same element with different text content of `<style>`. Use custom properties to style instances instead (e.g. `--element-color: red`).

Using constants for theme styling across elements is fine, but it's still recommended to make the expression a complete `<style>` tag. That way, the shared style tag will only be parsed once, even if multiple element classes it. For example:

```js
// 2 style tags with different text content, `partialStyle` parsed twice.
// element-one
html`<style>
  color: blue;
  ${partialStyle}
</style>`;

// element-two
html`<style>
  color: red;
  ${partialStyle}
</style>`;
```

```js
// `completeStyle` has same text content, only parsed once.
[...]

PR #458, opened 2019-01-18
Fix documentation of :host and :host() pseudo-classes

djlauk:
I had some trouble getting the attribute `hidden` to work, because the examples in the docs are faulty.

* I fixed the code snippets / examples.
* I changed the explanatory introduction text, to point out that there are two pseudo-classes, how they are different, and which one to use when.

### Reference Issue
Fixes part of #353.


justinfagnani:
Thanks @djlauk!

PR #464, opened 2019-01-18
[docs] Add a search box

keanulee:
Staged at https://20190118t155041-dot-polymer-lit-element.appspot.com. Should be pretty simple to apply the same changes to the other product sites.


frankiefu:
The API page looks a bit off with the new change:
https://20190118t155041-dot-polymer-lit-element.appspot.com/api/index.html

keanulee:
Fixed, though now the UI is somewhat confusing since the TypeDoc search box is now visible (previously present but behind the top nav).

frankiefu:
Wondering if we should just keep hiding the TypeDoc toolbar.

keanulee:
The bad part about hiding the TypeDoc toolbar is that TypeDoc search is actually still accessible by keydown events on body - i.e. you can start typing characters anywhere on the page and the search autocomplete shows.

frankiefu:
Wow I didn't know it does search autocomplete on keydown. I saw there is an [typedoc issue](https://github.com/TypeStrong/typedoc/issues/812) talks about how to get rid of keydown on the body element but not sure how easy it is to override that in custom theme. I guess for now the simplest way would be just go with what you have in the PR, which is already better than what currently deployed in production.

katejeffreys:
@frankiefu 

>not sure how easy it is to override that in custom theme

My understanding is that TypeDoc theme uses the default files unless they are overridden, so this would just involve adding that file with whatever modification required.

However, I tried this following the solution in the issue you linked, and it did not work. Maybe I also need to hide the TypeDoc toolbar? I took a look but couldn't figure out how we are hiding/unhiding it. Here's the branch

https://github.com/Polymer/lit-element/tree/searchbox-disable-keydown/docs/_api/theme

keanulee:
Updated to match lit-html docs - see https://github.com/Polymer/lit-html/pull/753#issuecomment-458319358

PR #468, opened 2019-01-22
Fix inherit  from constructor chain

### LINKED ISSUE #467: `styles` are not read from the constructor chain
#### Live Demo
https://stackblitz.com/edit/lit-element-request-update-bug-lspgdq?file=test.js

#### Steps to Reproduce
1. Create function `MyStyleMixin` with `static get styles` defined. For instance:
```js
const MyStyleMixin = (superclass => class MyStyleMixin extends superclass {
  static get styles() { return [
    css`
      div {
        border: 2px solid blue;
      }`
    ];
  }
});
```
2. Create class `MyStyledElement` that applies `MyStyleMixin` and extends `LitElement`. 
Define a `render` function that are styled by `MyStyleMixin`. Define it as custom element `my-styled-element`.
For instance :
```js
class MyStyledElement extends MyStyleMixin(LitElement) {
  render() {
    return html`<div>Styles should be applied here</div>`;
  }  
}
customElements.define('my-styled-element', MyStyledElement);
```
3. Instantiate `<my-styled-element></my-styled-element>` on a page.

#### Expected Results
`<my-styled-element>` has a div with a blue 2px border.

#### Actual Results
`<my-styled-element>` has no styling at all.

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari 11
- [x] Safari 10
- [x] IE 11

### Versions

- lit-element: v2.0.0-rc.3
- webcomponents: v2.2.3

tlouisse:
### Reference Issue
Fixes #467

Instead of checking 'current' constructor level only, this checks the whole constructor chain for `static get styles`. In this way, it never misses `styles` defined on constructors never 'ran' before (like mixins): https://stackblitz.com/edit/lit-element-request-update-bug-lspgdq?file=test.js

This could however mean that this._styles is created redundantly for constructor layers that should have the same outcome as their parents(and therefore should be able tor reuse `this._styles`). A code example that addresses this problem can be found here:
https://github.com/tlouisse/lit-element/blob/fix/getStylesFromConstructorChainAlternative/src/lit-element.ts#L92
But I'm not sure if this is more performant and outweighs the benefits of less code. 

Let me know what you think :)

### DIFF
--- a/src/lit-element.ts
+++ b/src/lit-element.ts
@@ -73,8 +73,8 @@ export class LitElement extends UpdatingElement {
   private static get _uniqueStyles(): CSSResult[] {
     if (!this.hasOwnProperty(JSCompiler_renameProperty('_styles', this))) {
       // Inherit styles from superclass if none have been set.
-      if (!this.hasOwnProperty(JSCompiler_renameProperty('styles', this))) {
-        this._styles = this._styles !== undefined ? this._styles : [];
+      if (!this[JSCompiler_renameProperty('styles', this)]) {
+        this._styles = [];
       } else {
         // Take care not to call `this.styles` multiple times since this generates
         // new CSSResults each time.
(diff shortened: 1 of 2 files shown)


tlouisse:
Closing, since already fixed via https://github.com/Polymer/lit-element/pull/470

Issue #469, opened 2019-01-22
Custom property Accessors listing issues with ts-lint


Christian24:
### Description

The docs show how to use custom accessors. However, upon trying I found out that typing  `static get properties()`is not possible in TypeScript, because `PropertyDeclarations` is not exported. Also since RC3 `requestUpdate`has to be called inside of the set accessor. TS-Lint always fails here, because requestUpdate returns a Promise that is never waited for and according to https://github.com/Microsoft/TypeScript/issues/14982 cannot be waited for anyway in accessors.

It would also be great to have an example in the docs how to achieve something similar with the`@property` decorator. I have figured out that putting the decorator on the getter works fine (see child2.ts in the stackblitz), but is it intended to be that way? I kinda like the idea of doing something similar to `attribute`, where you can overwrite the name of the property yourself, so the decorator could be put on the private class variable (that likely also has the initial value). 
#### Live Demo

child.ts shows the implementation with `@property`:
https://stackblitz.com/edit/lit-element-hello-world-85lc6q?file=child.ts

child2.ts shows the get properties implementation:
https://stackblitz.com/edit/lit-element-hello-world-85lc6q?file=child2.ts

#### Expected Results
No linting issues when using custom accessors. Documentation on how to implement custom accessors with `@property` decorator.

#### Actual Results
Two linting issues with custom accessors. 

### Versions

- lit-element: v2.0.0-rc3
- webcomponents: vX.X.X

Christian24:
I had another look at the source code: I think it would be an acceptable solution to not return `updateComplete` at the end of `requestUpdate` as the return value is never used in the source code and `updateComplete` is available as a get accessor anyway.  I made this adjustment in my own fork already, will submit a PR soon.

sorvell:
The return value is there merely for convenience, and this could be removed. That said, it seems pretty unfortunate that TypeScript is complaining about this.

@justinfagnani Any thoughts?

daKmoR:
I think this is quite convenient as well
```js
await el.requestUpdate();
// would become
el.requestUpdate();
await el.updateComplete;
```

or a little bigger example:
```js
el.requestUpdate().then(() => {
  const text = el.querySelector('div > p > span').innerText;
});

// would become

el.requestUpdate();
await el.updateComplete.then(() =>
  const text = this.querySelector('div > p > span').innerText;
});
```

if possible it would be nice to keep it 🤗 

 if it's really that big of a problem I think we probably could survive without it 😛

PR #472, opened 2019-01-23
Do not return updateComplete at the end of requestUpdate

Christian24:
This addresses an issue mentioned in #469. When implementing custom accessors, TS-Lint always has a finding on calling `requestUpdate`, because the result of the returned promise is ignored. As async and await is not supported in accessors anyway, my suggestion is to just not return `updateComplete`, as the return value is never used in the source code anyway and can easily be substituted for `updateComplete` get  accessor if needed.


Christian24:
I need to work some more on this, before it is ready. Sorry, for the fast first shot.

Christian24:
I guess the CLA check failed, because I have used to a different email address in my first commit.

Issue #471, opened 2019-01-23
Allowing number values in CSSResults


daKmoR:
hey,

we have some default number values for our styling. 
Reason:
- We want to keep them the same across multiple elements
- These values should not be changeable (e.g. we don't want to use css variables)
  - as we don't want them as part of our "api"
  - as we want to do javascript calculations
  - as we want to keep the door open to optimize these variables "away" inside `css` via a build step

#### Live Demo
https://stackblitz.com/edit/ems7at?file=custom-greeting.js

#### Steps to Reproduce
Use a number

```js
import { defaultPadding } from './css-constants.js';
// const defaultPadding = 5;

static get styles() {
  return [css`
    :host {
      padding: ${defaultPadding*2}px;
    }
  `];
}
```

#### Expected Results
Renders with padding of 10px.

imho Numbers should be allowed as they "are save" and can not be used for any form of css attack right?

#### Actual Results
Error thrown
```
Error: Value passed to 'css' function must be a 'css' function result: 10.
```

### Browsers Affected
- all

### Versions
- lit-element: v2.0.0-rc.x

kenchris:
CSS environment variables is supposed to handle this: https://drafts.csswg.org/css-env-1/#env-function

But it isn't specced yet how to add custom environment variables from either CSS or JS. Maybe @justinfagnani can push internally for moving that forward instead of the above idea

daKmoR:
nice 💪  makes it definitely more useful 👍  especially for import.meta.url 🤗 

I'm still unsure why such a wrapper would be needed for a value of type number? 
I assumed a number is always safe? I feel like I am missing something if that is not the case 🙈 

```
padding: ${unsafeCss( 5*2 )}px;
```
feels kinda strange...

sorvell:
Fixed via https://github.com/Polymer/lit-element/pull/474.

PR #474, opened 2019-01-23
Adds `unsafeCss` for composing values into `css`

### LINKED ISSUE #451: Consider allowing string values in CSSResults?
Currently `css` only accepts `CSSResult` values (other `css` calls) which means we can no longer do something like this:

```ts
static get styles() {
  return [css`
    :host {
      background-image: url('${rootPath}images/foo.png');
    }
  `];
}
```

This is a common pattern I have throughout my project to handle usages of `import.meta` so assets can have absolute paths.

If you can suggest a better way to handle passing in a path, fair enough, but otherwise it would be handy to be able to interpolate string values here.

sorvell:
Fixes #451 and #471. Non-literal values can now be included in styling with `css` by using the `unsafeCss` function. This is named "unsafe" so users know this they must use this carefully to avoid security issues.

### DIFF
--- a/src/lib/css-tag.ts
+++ b/src/lib/css-tag.ts
@@ -12,13 +12,20 @@ found at http://polymer.github.io/PATENTS.txt
 export const supportsAdoptingStyleSheets =
     ('adoptedStyleSheets' in Document.prototype);
 
+const constructionToken = Symbol();
+
 export class CSSResult {
 
   _styleSheet?: CSSStyleSheet|null;
 
   readonly cssText: string;
 
-  constructor(cssText: string) { this.cssText = cssText; }
+  constructor(cssText: string, safeToken: symbol) {
+    if (safeToken !== constructionToken) {
+      throw new Error('CSSResult is not constructable. Use `unsafeCss` or `css` instead.');
+    }
+    this.cssText = cssText;
+  }
 
   // Note, this is a getter so that it's lazy. In practice, this means
   // stylesheets are not created until the first element instance is made.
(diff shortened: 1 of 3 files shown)


kenchris:
Some documentation about what common security issues to be aware of, would be a nice addition

src/lib/css-tag.ts
```diff
@@ -37,6 +44,10 @@ export class CSSResult {
   }
 }
 
+export const unsafeCss = (value: any) => {
+  return new CSSResult(String(value), constructionToken);
+};
+
 const textFromCSSResult = (value: CSSResult) => {
   if (value instanceof CSSResult) {
     return value.cssText;
```
rictic: Consider mentioning unsafeCss in the error message here.

src/lib/css-tag.ts
```diff
@@ -37,6 +44,10 @@ export class CSSResult {
   }
 }
 
+export const unsafeCss = (value: any) => {
```
rictic: Prefer `unknown` to `any` here (and generally).

rictic: Documentation? Maybe:

```
/**
 * Wrap a value for interpolation in a css tagged template literal.
 *
 * This is unsafe because untrusted CSS text can be used to phone home
 * or exfiltrate data to an attacker controlled site. Take care to only use
 * this with trusted input.
 */
```

sorvell: Done, thanks.

src/lib/css-tag.ts
sorvell: Done.

src/lib/css-tag.ts
sorvell: Done.

dfreedm:
> Some documentation about what common security issues to be aware of, would be a nice addition

@kenchris: Good point. Some documentation was added to the function, but I've also filed lit/lit.dev#868 to add some further detail and examples.

PR #476, opened 2019-01-24
Fix issue with Closure Compiler and #470

azakus:


azakus:
@justinfagnani We'd need a whole example app, as the library compiles cleanly.

PR #491, opened 2019-01-29
[docs] TypeScript sample on front page with JS in a code editor

katejeffreys:


docs/index.md
```diff
@@ -58,27 +58,29 @@ Get started
 </h1>
 
 <h2>
-Define a component in JavaScript:
+Define a component
```
justinfagnani: Format as single line

docs/_includes/projects/index-typescript/custom-greeting.ts
```diff
@@ -0,0 +1,10 @@
+import { LitElement, html, property } from 'lit-element';
+
+export class CustomGreeting extends LitElement {
+  @property({ type: String}) name = 'World';
+  render() {
+    return html`<p>Hello, ${this.name}!</p>`;
+  }
+}
+
+customElements.define('custom-greeting', CustomGreeting);
```
justinfagnani: let's use the `customElement` decorator too:

```ts
@customElement('custom-greeting')
export class CustomGreeting extends LitElement { 
```

docs/index.md
```diff
@@ -58,27 +58,29 @@ Get started
...
 </h2>
 
-<h3 style="margin: 12px 0 0 0;">
-custom-greeting.js
-</h3>
+It's easy to define a Web Component with LitElement:
 
 ```js
-{% include projects/index/custom-greeting.js %}
+{% include projects/index-typescript/custom-greeting.ts %}
```
justinfagnani: How about a note like `* This example uses TypeScript and decorators. You can also easily write LitElements in plain JavaScript`

docs/_includes/projects/index-typescript/index.html
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <meta http-equiv="X-UA-Compatible" content="ie=edge">
+  <script src="/node_modules/@webcomponents/webcomponentsjs/custom-elements-es5-adapter.js"></script>
+  <script src="/node_modules/@webcomponents/webcomponentsjs/webcomponents-bundle.js"></script>
```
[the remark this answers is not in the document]
katejeffreys: Easier loading process I think? i forget

docs/_includes/projects/index-typescript/custom-greeting.ts
katejeffreys: Do you also then have to add the `interface HTMLElementTagThingStuff` ?

https://github.com/Polymer/lit-element/blob/0093b439fa53e26707a84690d17bf15f54baae3e/src/lib/decorators.ts#L78

 Seems noisier

Christian24: @katejeffreys No, as in #291 this was changed to only be a string. Before it would actually check if the string was part of `HTMLElementTagNameMap`. Probably the comment in the source code should also be removed.

katejeffreys: can we then remove `customElements.define('custom-greeting', CustomGreeting);`?

docs/_includes/projects/index-typescript/index.ts
```diff
@@ -0,0 +1 @@
+import './custom-greeting.ts';
```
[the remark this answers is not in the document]
katejeffreys: because stackblitz, consistency, & clarity... 

- StackBlitz requires index.ts/js & errors if I import modules from html
- I want to show the user clear chunks of code rather than large files
- with multiple elements in an example, writing them all in index.* becomes clunky so I have separate files for each element - for consistency, I do this even if there is only 1 element

docs/_includes/projects/index-typescript/custom-greeting.ts
Christian24: @katejeffreys Yes, `@customElement('custom-greeting')` and `customElements.define('custom-greeting', CustomGreeting);` do the same thing. Only one is needed.

docs/_includes/projects/index-typescript/index.ts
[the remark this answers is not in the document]
justinfagnani: I agree with separate files for elements in most cases (trying to show only how two elements can compose might be easier with two very small elements in one file).

The stackblitz limitations should be gone with the `lit-html` project template, which will hopefully be usable tomorrow. We can think about migrating to that later...

PR #498, opened 2019-01-30
WIP and a proposal for updating the styles doc.

### LINKED ISSUE #418: [docs] Document style api
see #401 & #391

justinfagnani:
This took a bit, so I had to rush something in before the end of day and it's a bit rough...

I added a brief intro to shadow DOM scoping, sections on static styles, inline styles and external styles, and tried to clarify the organization around what elements styles apply to: host, shadow root, or slotted.

Not sure if the top-level headings are right.
We also need to call into lit-html's styling docs when those are up.

Fixes #418

### DIFF
--- a/docs/_guide/styles.md
+++ b/docs/_guide/styles.md
@@ -6,7 +6,7 @@ slug: styles
 
 ## Shadow DOM
 
-LitElement uses [Shadow DOM](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM) by default for DOM and style encapsulation. Shadow DOM scopes CSS so that styles defind in a shadow root only apply to DOM inside the shadow root and do not leak to outside DOM. Shadow roots are also isolated from styles defined outside the shadow root, whether in the main page or a outer shadow root.
+LitElement uses [Shadow DOM](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM) by default for DOM and style encapsulation. Shadow DOM scopes CSS so that styles defined in a shadow root only apply to DOM inside the shadow root and do not leak to outside DOM. Shadow roots are also isolated from styles defined outside the shadow root, whether in the main page or an outer shadow root.
 
 **This guide applies only if you use the default (shadow DOM) render root.** If you modify your element's render root to render into the main DOM tree instead of a shadow root, these instructions won't apply.
 {.alert-info}
--- a/docs/_guide/styles.md
+++ b/docs/_guide/styles.md
@@ -18,10 +18,10 @@ LitElement uses [Shadow DOM](https://developer.mozilla.org/en-US/docs/Web/Web_Co
 * ToC
 {:toc}
 
-## Styling shadow DOM
+### Styling shadow DOM
 
 Styles can be added to a shadow root in three different ways:
-* With styles defined in the static `styles` property of a LitElement class (reccomended).
+* With styles defined in the static `styles` property of a LitElement class (recommended).
 * With a `<style>` element in the shadow root
 * With a `<link rel="stylesheet">` element in the shadow root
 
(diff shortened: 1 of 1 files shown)


docs/_guide/styles.md
```diff
@@ -4,66 +4,247 @@ title: Styles
...
+    `
+  ];
+}
+```
+
+Here the subclass is explicitly inheriting the superclass styles, then adding its own.
+
+Static styles are the reccomended way to style LitElements because on browsers that support it, they generate and use a new standard called _[Constuctible Stylesheets](https://wicg.github.io/construct-stylesheets/)_, with a fallback for browsers that don't.
+
+Constuctible Stylesheets allow LitElement to parse styles exactly once and reuse the resulting Stylesheet object for maximum efficiency.
```
kenchris: Const*r*uctable

docs/_guide/styles.md
```diff
@@ -4,66 +4,247 @@ title: Styles
...
+
+### Limitations
+
+Static styles apply to all instances of an element. Any expressions in the style text are evaluated and included once, then reused for all instances. If you need to vary styles per-instance we reccomend using CSS custom properties, or inline styles.
```
kenchris: Recommend (one c)

docs/_guide/styles.md
```diff
@@ -4,66 +4,247 @@ title: Styles
...
+    }
   `;
 }
 ```
 
-{% include project.html folder="docs/style/hostselector" openFile="my-element.js" %}
+## Inline styles
 
-See the MDN documentation on [:host](https://developer.mozilla.org/en-US/docs/Web/CSS/:host), [:host()](https://developer.mozilla.org/en-US/docs/Web/CSS/:host()), and [pseudo-classes](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-classes) for more information.
+You can also style a shadow root with inline styles right in your element template. We still reccomend static styles, but in some cases you may want to vary the CSS per-element. One way to do this is with bindings in `<style>` elements. It's important to note that this will not work with shadow DOM polyfills like ShadyCSS.
```
kenchris: Recommend again

docs/_guide/styles.md
```diff
@@ -4,66 +4,247 @@ title: Styles
 slug: styles
 ---
 
+## Shadow DOM
+
+LitElement uses [Shadow DOM](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM) by default for DOM and style encapsulation. Shadow DOM scopes CSS so that styles defind in a shadow root only apply to DOM inside the shadow root and do not leak to outside DOM. Shadow roots are also isolated from styles defined outside the shadow root, whether in the main page or a outer shadow root.
```
jbingham:
```suggestion
LitElement uses [Shadow DOM](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM) by default for DOM and style encapsulation. Shadow DOM scopes CSS so that styles defined in a shadow root only apply to DOM inside the shadow root and do not leak to outside DOM. Shadow roots are also isolated from styles defined outside the shadow root, whether in the main page or an outer shadow root.
```

katejeffreys:
fyi i am editing in [another branch](https://github.com/Polymer/lit-element/tree/styles-doc-update-kj), will PR into this branch

docs/_guide/styles.md
```diff
@@ -4,66 +4,247 @@ title: Styles
 slug: styles
 ---
```
jbingham: I'm wanting an intro section with a more general name. Like:

Overview

LitElement provides flexibility in how you style your components, and ensures interoperability through style encapsulation using Shadow DOM, the browser's built-in way of encapsulating native UI components.

Why use Shadow DOM? Shadow DOM scopes CSS so that...

docs/_guide/styles.md
```diff
@@ -4,66 +4,247 @@ title: Styles
...
 
 **If you're using the Shady CSS polyfill, be aware that it has some limitations.** See the [Shady CSS README](https://github.com/webcomponents/shadycss/blob/master/README.md#limitations) for more information.
+{.alert-info}
 
 {::options toc_levels="1..3" /}
 * ToC
 {:toc}
 
-## Use the :host and :host() CSS pseudo-classes
+### Styling shadow DOM
```
jbingham:
```suggestion
### Styling your custom elements
```

docs/_guide/styles.md
```diff
@@ -4,66 +4,247 @@ title: Styles
...
 
-When styling your custom element, you can use the `:host` and `:host()` CSS pseudo-classes in a `<style>` block to select the host element (the element hosting the root of your shadow DOM). The two pseudo-classes slightly differ in usage:
+Styles can be added to a shadow root in three different ways:
```
jbingham:
```suggestion
Styles can be added to a custom element in three different ways:
```

docs/_guide/styles.md
```diff
@@ -4,66 +4,247 @@ title: Styles
...
+          color: ${this.mainColor};
+        }
+      </style>
+    `;
+  }
+}
+```
 
-Two best practices for working with custom elements are:
+We strongly reccomend static styles, CSS custom properties, or [lit-html's `classMap` or `styleMap` directives](TODO) if you're stying non-host shadow root contents.
```
katejeffreys: what does non-host shadow root contents mean? could you give an example? I'm not sure we need to cover it

docs/_guide/styles.md
```diff
@@ -4,66 +4,247 @@ title: Styles
...
+  }
+}
+```
+
+This can be a good way to load CSS generated from tools like SASS/LESS.
+
+There are some important caveats though:
+
+* External styles can cause a flash-of-unstyled-content (FOUC) while they load.
+* The URL in the `href` attribute is relative to the _main document_, making this technique mostly useful for application elements where asset URLs are well known, and not for reusable elements published publically.
```
katejeffreys: * As we are not documenting how to build apps, (covered in PWA starter kit) should this content be here or somewhere else? @jbingham @graynorton 
* If it belongs here, someone is bound to ask how to get the URL the the element was imported from (can't remember how to do it, import meta something something..) should we cover that somewhere?

PR #508, opened 2019-01-31
Adds `toString` to CSSResult

### LINKED ISSUE #487: Consider implementing `toString()` for CSSResult
### Description
Currently, CSSResult seems to be designed for `LitElement`s having shadow dom. CSSResult modules tend to be quite closely coupled to the `get static styles` array as well.

Imagine however the following situation: I have an implementation of a Design System consisting of css modules (meaning equivalents of BEM/SMACCS etc. type of components) that should be reusable in many 'technology contexts' (meaning with and without shadow DOM and in multiple frameworks (React, Angular, Vue, vanilla, whatever).

This example illustrates how one would be able to reuse a 'technology agnostic' implementation of a Design System across different technologies:
- example of a css module: https://stackblitz.com/edit/css-modules-in-all-contexts?file=style%2Fmodules%2Fbutton-css-module.js
- example of applying local styles (in LitElement) with shadowRoot: https://stackblitz.com/edit/css-modules-in-all-contexts?file=my-lit-app.js
- example of applying global styles (in a non Lit context): https://stackblitz.com/edit/css-modules-in-all-contexts?file=index.ts

For me, it would increase the Developer Experience when I can think about those CSS Modules as being technology agnostic strings. 
So then instead of `{myModule.cssText}`, I could do `{myModule}`.

Seems like a small thing, but for me it would just be a better experience and it would feel like my styles would be less tightly coupled to CSSResult (and thus LitElement and shadow dom).

Also, It will help decrease verbosity in situations like this:
```css
.grid-container {
  margin: auto;
  max-width: ${unsafeCss(Number(small.cssText.slice(0,-2)) -(Number(containerPadding.cssText.slice(0,-2) * 2)))}px;
}
```
https://stackblitz.com/edit/css-variables-for-computations?file=style%2Fmodules%2Fgrid-css-module_numbersNotAllowed.js

Would it be possible to add a `toString()` method that returns the `cssText`?

#### Live Demo
See links above.

tlouisse:
### Reference Issue
Fixes #487

### DIFF
--- a/src/lib/css-tag.ts
+++ b/src/lib/css-tag.ts
@@ -42,6 +42,10 @@ export class CSSResult {
     }
     return this._styleSheet;
   }
+
+  toString(): String {
+    return this.cssText;
+  }
 }
 
 /**
--- a/src/test/lit-element_styling_test.ts
+++ b/src/test/lit-element_styling_test.ts
@@ -768,6 +768,14 @@ suite('Static get styles', () => {
     const div = el.shadowRoot!.querySelector('div');
     assert.equal(getComputedStyle(div!).getPropertyValue('border-top-width').trim(), '4px');
   });
+
+  test('`CSSResult` allows for String type coercion via toString()', async () => {
+    const cssModule = css`.my-module { color: yellow; }`;
+    // Coercion allows for reusage of css-tag outcomes in regular strings.
+    // Example use case: apply cssModule as global page styles at document.body level.
+    const bodyStyles = `${cssModule}`;
+    assert.equal(bodyStyles, '.my-module { color: yellow; }');
+  });
 });
 
 suite('ShadyDOM', () => {


justinfagnani:
See my comments on #487. I'm not sure of the utility here.

src/lib/css-tag.ts
```diff
@@ -42,6 +42,10 @@ export class CSSResult {
     }
     return this._styleSheet;
   }
+
+  toString(): String {
```
balloob: Shouldn't this be type `string` ?

justinfagnani: Absolutely. Good catch!

balloob: https://github.com/Polymer/lit-element/pull/546

PR #516, opened 2019-02-03
Revise per feedback on "try LitElement" tutorial

katejeffreys:


docs/_try/logic.md
```diff
@@ -16,28 +28,57 @@ _my-element.js_
...
+    ```
 
-    ```html
-    ${this.myBool?
-      html`<p>Render some HTML if myBool is true</p>`:
+3.  **Add a loop to your template.**
+    
+    To loop over the new `myArray` property, add the following code to your template:
+
+    ```js
```
justinfagnani: this is HTML at the top level here. I think you need to add a `html\`` to make it highlight correctly.

docs/_try/logic.md
```diff
@@ -16,28 +28,57 @@ _my-element.js_
...
+    <ul>${this.myArray.map(item => html`<li>${item}</li>`)}</ul>
```
justinfagnani: nit: I might put the expression on a new line for readability.

Issue #506, opened 2019-01-31
[docs] Mention directives in docs and link lit-html docs


Christian24:
Hi,

I have a suggestion for the docs: Maybe it makes sense to mention directives in the Templates-Guide in the docs. This is kinda extending on #430. My point is this: People who start learning LitElement might not be aware that directives exist, as they are only exported from and documented in lit-html.  Therefore I think it makes sense to add a couple sentences saying something like "If you want to use advanced templating features of lit-html in conjunction with LitElement you can find more info in the lit-html docs here: https://lit-html.polymer-project.org/guide/writing-templates. To use these features you need to install lit-html alongside LitElement". 

What do you think? I might be able submit an early draft of it as well if desired.

katejeffreys:
Will look at this tonight or tomorrow. Thanks!

justinfagnani:
I think along with this it would be nice to add an intro to the page that mentions lit-html with a link to it's site.

Christian24:
I do agree. I feel sometimes the docs loose the connection to lit-html. Especially the template guide sounds a bit like all these concepts are from LitElement, while in reality they do come from lit-html.

arthurevans:
Fixed in #826.

PR #518, opened 2019-02-03
[docs] Add information about lit-html features in template syntax

Christian24:
This is a draft of a fix for #506. It does add an advanced features section which says how to use lit-html and LitElement together. The link is just to the source code as apparently those docs are not live yet. 

@katejeffreys Is it intentional that lit-html is not really mentioned on this page? It feels like this page duplicates some parts of the lit-html documentation while never mentioning lit-html once. I would probably mention once that `html`is a lit-html function and that the syntax is the same and under the hood LitElement uses lit-html to render the template. 

Also `nothing` does not seem to be mentioned in the lit-html documentation, will open an issue for it.

### DIFF
--- a/docs/_guide/templates.md
+++ b/docs/_guide/templates.md
@@ -503,4 +503,7 @@ of the entire lit-html featureset for writing your templates:
 For example you can use the `nothing` sentinel value in case you want nothing to be rendered or you can use directives.
 You can find more information on writing templates with lit-html in the lit-html documentation:
 [Template Reference](https://github.com/Polymer/lit-html/blob/master/docs/_guide/05-template-reference.md)
-<!-- add link to published to https://github.com/Polymer/lit-html/blob/master/docs/_guide/05-template-reference.md once it is ready --!>
+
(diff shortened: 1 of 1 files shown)


Christian24:
Removed the paragraph about having to install lit-html as a dependency yourself thanks for the info @justinfagnani and @ruphin

docs/_guide/templates.md
```diff
@@ -496,3 +496,14 @@ render() { return html`<slot name="thing"></slot>`; }
   <p slot="thing">stuff</p>
 </my-element>
 ```
+
+## Advanced features
```
justinfagnani: Rather than "Advanced Features", what about calling this section "Further Reading" and lead with a link to the lit-html site?

cc @katejeffreys

Christian24: I do agree with the idea of it being not advanced features, more a further reading. I made some minor changes based on your ideas, especially made the links stand out more. The wording of the list could be improved though.

PR #521, opened 2019-02-04
[docs] Simplify home page

graynorton:
* Simplify home page structure to match lit-html
* Allow custom label on <stack-blitz> element


arthurevans:
Staged at https://simplify-home-page-dot-polymer-lit-element.appspot.com

docs/index.md
```diff
@@ -11,74 +11,47 @@ layout: default
 </div>
 </header>
 
-<section>
+<section id="section-snippet">
 <div class="wrapper">
 <h1 class="title">
-About
+Easily Create Fast, Lightweight Web Components
```
jbingham: +1. This is an important distinction from the previous wording... it's about creating new custom components, not a collection of existing components that are fast and lightweight.

docs/index.md
```diff
@@ -90,51 +51,36 @@ The example above uses TypeScript and decorators. You can also easily write LitE
 <div class="responsive-row">
 <div style="flex: 1">
 
-<h2 class="caption">Fast and light</h2>
+<h2 class="caption">Delightfully Declarative</h2>
```
arthurevans: Capitalization is inconsistent across these three heads.

Delightfully Declarative
Fast and light
Seamlessly Interoperable

Should either capitalize "Light" or lowercase "declarative" and "interoperable".

We use sentence case for headings in the docs, but the context is so different here I don't think that applies.

PR #536, opened 2019-02-05
Add version global

### LINKED ISSUE #526: Add global variable for framework detection
Something like: `window.litElementVersions` which will be an array of the versions used in a page.

justinfagnani:
Fixes #526

### DIFF
--- a/src/lit-element.ts
+++ b/src/lit-element.ts
@@ -22,14 +22,27 @@ export {html, svg, TemplateResult, SVGTemplateResult} from 'lit-html/lit-html';
 import {supportsAdoptingStyleSheets, CSSResult} from './lib/css-tag.js';
 export * from './lib/css-tag.js';
 
-export interface CSSResultArray extends Array<CSSResult | CSSResultArray> {}
+declare global {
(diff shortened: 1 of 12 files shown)


src/lit-element.ts
```diff
@@ -22,14 +22,27 @@ export {html, svg, TemplateResult, SVGTemplateResult} from 'lit-html/lit-html';
...
 
-export interface CSSResultArray extends Array<CSSResult | CSSResultArray> {}
+declare global {
+  interface Window {
+    litElementVersions: string[];
+  }
+}
+
+// IMPORTANT: do not change the property name or the assignment expression.
+// This line will be used in regexes to search for lit-html usage.
```
rictic: lit-html -> LitElement

src/lit-element.ts
```diff
@@ -22,14 +22,27 @@ export {html, svg, TemplateResult, SVGTemplateResult} from 'lit-html/lit-html';
...
+// TODO: inject version number
```
rictic: clarify that you mean at build time?

PR #558, opened 2019-02-10
Remove semicolon from customElement decorator on README example

mnixo:
Fixes the `customElement` decorator usage on the overview example.


justinfagnani:
Thanks!!

PR #560, opened 2019-02-11
Fixed typo.

uwerelax:


keanulee:
Merged #568 instead since CLA is missing.

PR #571, opened 2019-02-15
Fix syntax

arthurevans:
`:host([.important])` doesn't match anything. This should either be an attribute selector:

`:host([important])` matches `<my-element important>`

Or a class selector:

`:host(.important)` matches `<my-element class="important">`

### DIFF
--- a/docs/_guide/styles.md
+++ b/docs/_guide/styles.md
@@ -96,7 +96,7 @@ To define a static `styles` property:
           static styles = [
             super.styles,
             css`
-              :host([important]) {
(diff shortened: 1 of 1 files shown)


arthurevans:
Fixed one of two possible ways; feel free to close & do it the other way if it fits better.

jsilvermist:
I think this is the best way, since it can be easily used for dynamic styling based on reflecting attributes.

justinfagnani:
I had actually intended a class, but it really doesn't matter - it shows the basics either way.

PR #583, opened 2019-02-21
remove in development status from api-readme

armujahid:
Since stable `lit-element` has already been released so `in development` status should be removed.


justinfagnani:
thanks!

Issue #584, opened 2019-02-21
[docs] example code for static style property results in syntax error


anneb:
The example code in the documentation for static styles on https://github.com/Polymer/lit-element/blob/master/docs/_guide/styles.md currently reads:

```
class MyElement extends LitElement {
  static styles = css`
    :host {
      display: block;
    }
  `;
}
```

This code produces a `syntax error: unexpected token '='` when used with poymer-cli (to be used as described in the same documentation under 'getting started')

justinfagnani:
Fixed by #585

Victor-Bernabe:
> The example code in the documentation for static styles on https://github.com/Polymer/lit-element/blob/master/docs/_guide/styles.md currently reads:
> 
> ```
> class MyElement extends LitElement {
>   static styles = css`
>     :host {
>       display: block;
>     }
>   `;
> }
> ```
> This code produces a `syntax error: unexpected token '='` when used with poymer-cli (to be used as described in the same documentation under 'getting started')

Use:


    static get styles(){
      return css`
        :host {
          display: block;
        }`;
    }

PR #585, opened 2019-02-21
change syntax for static styles so that polymer-cli supports it

anneb:
PR for https://github.com/Polymer/lit-element/issues/584


justinfagnani:
thanks!

PR #588, opened 2019-02-22
docs: regenerate api documentation

armujahid:
generated docs were outdated that's why I regenerated docs using `npm run gen-docs`.
I think all docs should be generated automatically during deployment so we don't have to manually do that periodically.


keanulee:
Covered by #593, but thanks for the heads up.

PR #593, opened 2019-02-26
Add section indicator in top nav

keanulee:
Staged at https://20190226t134638-dot-polymer-lit-element.appspot.com


keanulee:
Deployed

PR #596, opened 2019-02-28
Ensure a property is reflect to attribute when it is changed after an attribute change

### LINKED ISSUE #592: Boolean Attribute no longer reflected after adding and removing the attribute
### Description
Boolean type's "reflect attribute" does not work if you manually setAttribute, and removeAttribute, without setting a default value for the boolean property.

#### Live Demo
https://stackblitz.com/edit/lit-element-example-wdqbfc?file=index.js

#### Steps to Reproduce
The attached example shows adding an attribute called "disabled", removing an attribute called "disabled", and then setting the property "disabled" to true.

The result is the disabled property is set to true, but the attribute "disabled" does not appear on the element.

#### Expected Results
Expected that the attribute "disabled" would appear on the element.

#### Actual Results
Attribute "disabled" does not appear on the element.

It appears this can be resolved by setting this.disabled = false, in the constructor.  
I see that the documentation now has the following comment:
"If you implement a static properties getter, initialize your property values in the element constructor." 

Is initializing all property values now required?  If so, the warning his helpful, but it might make sense to do a bit more - like explain what happens if you don't heed the warning.

### Browsers Affected

- [X] Chrome
- [X] Firefox
- [?] Edge
- [?] Safari 11
- [X] Safari 10
- [?] IE 11

### Versions
- lit-element: v2.0.0-rc.2
- webcomponents: v2.2.3

t-soares:
Fixes #592

The update method is responsible to reflect property changes to attributes. Changing a property that has reflect equals true will populate the _reflectingProperties to indicate that the changed property must be reflected to its attribute.

There is a logic that prevents the property being changed from being added to the _reflectingProperties, if the property change was caused by an attribute change.

There is also the _changedProperties map that contains the keys for any properties that have changed since the last update cycle.

If the the same property is changed multiple times on the same update cycle, of course its key is added just once _changedProperties map.

The problem occurs when you change an attribute and on the sae update cycle change the property. In this scenario, when you change the attribute then the property is not added to the _reflectingProperties map. iI after changing the attribute, you change the property then the property is already on the _changedProperties map and the logic that adds the property to the _reflectingProperties is not executed in this case.

This leads to the update method to not reflect the property change to the attribute.

I changed the updating-element.ts code to add to the _reflectingProperties map when the property is changed, even if the property is already on the _changedProperties map. Also if the property change come from an attribute change I remove the property from the  _reflectingProperties map.

### DIFF
--- a/src/lib/updating-element.ts
+++ b/src/lib/updating-element.ts
@@ -581,14 +581,16 @@ export abstract class UpdatingElement extends HTMLElement {
   requestUpdate(name?: PropertyKey, oldValue?: unknown) {
     let shouldRequestUpdate = true;
     // if we have a property key, perform property update steps.
-    if (name !== undefined && !this._changedProperties.has(name)) {
+    if (name !== undefined) {
       const ctor = this.constructor as typeof UpdatingElement;
       const options =
           ctor._classProperties!.get(name) || defaultPropertyDeclaration;
       if (ctor._valueHasChanged(
               this[name as keyof this], oldValue, options.hasChanged)) {
-        // track old value when changing.
-        this._changedProperties.set(name, oldValue);
+        if (!this._changedProperties.has(name)) {
+          // track old value when changing.
+          this._changedProperties.set(name, oldValue);
+        }
         // add to reflecting properties set
         if (options.reflect === true &&
             !(this._updateState & STATE_IS_REFLECTING_TO_PROPERTY)) {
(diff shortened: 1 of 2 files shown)


t-soares:
I signed it!

t-soares:
I signed it!

sorvell:
Thanks for the PR. I didn't notice it before doing https://github.com/Polymer/lit-element/pull/610 which is almost exactly the same. Closing this in favor of https://github.com/Polymer/lit-element/pull/610, but thanks again.

PR #598, opened 2019-03-01
Add homepage URL to package.json

abraham:
This will make it so npm will link to the homepage and GitHub instead of GitHub twice.


abraham:
Ping @justinfagnani

justinfagnani:
Thanks @abraham!

PR #600, opened 2019-03-03
Replaces cssLiteral (removed) with unsafeCSS 

hyperpress:
Updates style guide by removing reference to cssLiteral (previously removed) for using expressions in static styles with unsafeCSS method.

### DIFF
--- a/docs/_guide/styles.md
+++ b/docs/_guide/styles.md
@@ -111,26 +111,32 @@ To define a static `styles` property:
 
 #### Expressions in static styles
 
+The `css` tag only accepts literal strings. You can include expressions in your static styles using the `unsafeCSS` tag instead. 
+
 Static styles apply to all instances of an element. Any expressions in your CSS are evaluated and included **once**, then reused for all instances. 
 
-For security reasons, use **unsafeCSS** to pass non-literal values using only trusted sources:
+Example:
 
 ```js
-import {LitElement, css, unsafeCSS} from 'lit-element';
-
-const displayType = css`block`;
-const mainColor = unsafeCSS('#a50e0e');
+import { LitElement, unsafeCSS } from 'lit-element';
 
 class MyElement extends LitElement {
   static get styles() {
-    return css`
-    :host {
-      display: ${displayType};
-      color: ${mainColor};
-    }`;
+    return unsafeCSS`
+      :host {
+        display: ${this.getDisplayType()};
+        color: ${this.getMainColor()};
+      }
+    `;
   } 
 }
 ```
+
+<div class="alert alert-warning">
+
+**Only use the `unsafeCSS` tag with trusted input.** To prevent LitElement-based components from evaluating potentially malicious code, the `css` tag only accepts literal strings. `unsafeCSS` circumvents this safeguard.
+
+</div>
 
 ### Define styles in a style element {#style-element} 
 
--- a/docs/_guide/styles.md
+++ b/docs/_guide/styles.md
@@ -118,7 +118,7 @@ For security reasons, use unsafeCSS to pass non-literal values using only truste
 ```js
 import { unsafeCSS } from 'lit-element';
 
-const mainColor = unsafeCSS('red');
+const mainColor = css`red`;
 
 class MyElement extends LitElement {
   static get styles() {
--- a/docs/_guide/styles.md
+++ b/docs/_guide/styles.md
@@ -113,10 +113,10 @@ To define a static `styles` property:
 
 Static styles apply to all instances of an element. Any expressions in your CSS are evaluated and included **once**, then reused for all instances. 
 
-For security reasons, use unsafeCSS to pass non-literal values using only trusted sources:
+For security reasons, use **css** to pass non-literal values using only trusted sources:
 
 ```js
-import { unsafeCSS } from 'lit-element';
+import {LitElement, css} from 'lit-element';
 
 const mainColor = css`red`;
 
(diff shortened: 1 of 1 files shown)


web-padawan:
@sorvell @justinfagnani this is important to merge, as the docs are still pointing to `cssLiteral` and the users, who are not aware with LitElement internals, continue to face the issue with incorrect docs.

docs/_guide/styles.md
```diff
@@ -113,19 +113,19 @@ To define a static `styles` property:
 
 Static styles apply to all instances of an element. Any expressions in your CSS are evaluated and included **once**, then reused for all instances. 
 
-For security reasons, expressions must be tagged with the `cssLiteral` template literal tag:
+For security reasons, use unsafeCSS to pass non-literal values using only trusted sources:
```
justinfagnani: The replacement for `cssLiteral` is just `css`.

docs/_guide/styles.md
```diff
@@ -113,19 +113,19 @@ To define a static `styles` property:
...
 
 ```js
-import { LitElement, css, cssLiteral } from 'lit-element';
+import { unsafeCSS } from 'lit-element';
 
-const mainColor = cssLiteral`red`;
+const mainColor = unsafeCSS('red');
```
justinfagnani:
```suggestion
const mainColor = css`red`;
```

docs/_guide/styles.md
hyperpress: Done. Thanks for the clarification.

docs/_guide/styles.md
```diff
@@ -113,19 +113,19 @@ To define a static `styles` property:
 
 Static styles apply to all instances of an element. Any expressions in your CSS are evaluated and included **once**, then reused for all instances. 
 
-For security reasons, expressions must be tagged with the `cssLiteral` template literal tag:
+For security reasons, use unsafeCSS to pass non-literal values using only trusted sources:
 
 ```js
-import { LitElement, css, cssLiteral } from 'lit-element';
+import { unsafeCSS } from 'lit-element';
```
abraham: `LitElement` is used in the example so it should still be imported.

hyperpress: Fixed the import to include LitElement in latest commit.

docs/_guide/styles.md
```diff
@@ -113,19 +113,19 @@ To define a static `styles` property:
 
 Static styles apply to all instances of an element. Any expressions in your CSS are evaluated and included **once**, then reused for all instances. 
 
-For security reasons, expressions must be tagged with the `cssLiteral` template literal tag:
+For security reasons, use **css** to pass non-literal values using only trusted sources:
```
arthurevans: This appears to be the description for `unsafeCss`. You can't use `css` with non-literal values, and "trusted sources" doesn't make sense for a literal. Correct me if I'm wrong here, @justinfagnani

hyperpress: I refactored this a bit to hopefully provide coverage and clarity, but I may be missing something that is planned that will change this, such as #626. Sorry about dragging out what I thought would be a simple doc change. Please advise if I need to address something else here.

arthurevans:
Yeah, I agree it's confusing. 

When you use `css`, you can _only_ use it on a string literal.  This is OK:

```
const mainColor = css`red`;
return css`:host { color: ${mainColor}; }`
```

This is not:

```
const themeColor = 'red';
return = css`:host { color: ${themeColor}; }`;
```

Any expression you interpolate into the `css` string needs to be tagged with either `css` or `unsafeCSS`.  If you want to inject any variable or non-literal into a `css` string, you need to wrap it with `unsafeCSS`.

```
const themeColor = 'red';
return = css`:host { color: ${unsafeCSS(themeColor)}; }`;
```

Or:

```
css`:host { width: ${unsafeCSS(mainWidth + padding)}px; }
```

The outstanding PR adds support for expressions that return a number--in which case you could remove the `unsafeCSS` from the last example.

hyperpress:
@arthurevans Thanks very much for the detailed clarification. If #626 will be merged in time for the next release, I can add a quick example detailing that exemption. If not, I believe the latest commit will suffice as is until that change is made. If you or @justinfagnani can let me know I'll act accordingly. Cheers.

docs/_guide/styles.md
```diff
@@ -111,27 +111,33 @@ To define a static `styles` property:
...
+import { LitElement, unsafeCSS } from 'lit-element';
 
 class MyElement extends LitElement {
   static get styles() {
-    return css`
-    :host {
-      display: ${displayType};
-      color: ${mainColor};
-    }`;
+    return unsafeCSS`
```
arthurevans: I don't think you can use unsafeCSS this way. You can only use it as a filter inside an expression.

For example:

```
css`
  :host {
    color: ${unsafeCSS(mainColor)};
  }`;
```

arthurevans: i.e., it's not a tag function like `css`, it's a method you can use in expressions.

katejeffreys: yep sorry I merged too soon. fixed it

katejeffreys:
@hyperpress @arthurevans @web-padawan 

Sorry for the delay on this & heavy edits to John's PR. If any of y'all get the chance to look over it, that'd be great

katejeffreys:
@justinfagnani @sorvell fyi - thought i should dismiss your reviews because of significant changes to this section. If you get the chance to take another quick look in case I've accidentally reverted a change you asked for that would be super awesome. Sorry for the delay on this.

arthurevans:
LGTM.

hyperpress:
Good to go 👍

hyperpress:
Are w

arthurevans:
Yes, we should. Can you merge this and redeploy, @katejeffreys?

e111077:
ping on status on this pr. #637 is addressed in the contents of this PR

PR #606, opened 2019-03-12
Schedules `requestUpdate` in `initialize` rather than `connectedCallback`

### LINKED ISSUE #594: `updateComplete` should not resolve until the element has connected
### Description
If an element is created but not yet connected, _and_ no property has been set to enqueue an update, `updateComplete` currently resolves immediately (before the first render), rather than waiting until connection.  However, if at least one property has been set, it will wait until connection before resolving.  This is confusing and appears to be a bug.

Since an element will render unconditionally upon connection (regardless of whether it has properties that triggered rendering), the mental model should be that a newly constructed but disconnected element should be in a "pending update" state until it is connected.  Hence `updateComplete` should not resolve until the element has first connected.

#### Live Demo
https://jsbin.com/jiquhez/edit?html,console,output

#### Steps to Reproduce
1. Open link above

#### Expected Results
```
Found 1 element // result of `getItems` before connection
Found 1 element // result of `getItems` after connection
```

#### Actual Results
```
Found 0 element` // result of `getItems` before connection
Found 1 element` // result of `getItems` after connection
```

### Browsers Affected

- All

### Versions
- lit-element: 2.0.1

sorvell:
Fixes #594. Ensures an early access of `updateComplete` will not complete before the element has first updated (at connection time).

### DIFF
--- a/src/lib/updating-element.ts
+++ b/src/lib/updating-element.ts
@@ -441,6 +441,8 @@ export abstract class UpdatingElement extends HTMLElement {
    */
   protected initialize() {
     this._saveInstanceProperties();
+    // ensures first update will be caught by an early access of `updateComplete`
+    this.requestUpdate();
   }
 
   /**
@@ -484,15 +486,13 @@ export abstract class UpdatingElement extends HTMLElement {
 
   connectedCallback() {
     this._updateState = this._updateState | STATE_HAS_CONNECTED;
-    // Ensure connection triggers an update. Updates cannot complete before
+    // Ensure first connection completes an update. Updates cannot complete before
     // connection and if one is pending connection the `_hasConnectionResolver`
     // will exist. If so, resolve it to complete the update, otherwise
     // requestUpdate.
     if (this._hasConnectedResolver) {
       this._hasConnectedResolver();
       this._hasConnectedResolver = undefined;
-    } else {
-      this.requestUpdate();
     }
   }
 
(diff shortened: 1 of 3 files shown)


src/lib/updating-element.ts
```diff
@@ -484,15 +486,13 @@ export abstract class UpdatingElement extends HTMLElement {
...
-    // Ensure connection triggers an update. Updates cannot complete before
+    // Ensure first connection completes an update. Updates cannot complete before
     // connection and if one is pending connection the `_hasConnectionResolver`
     // will exist. If so, resolve it to complete the update, otherwise
     // requestUpdate.
     if (this._hasConnectedResolver) {
       this._hasConnectedResolver();
       this._hasConnectedResolver = undefined;
-    } else {
-      this.requestUpdate();
```
kevinpschaaf: Summary here is we're considering the fact that `requestUpdate` was called _every_ connection a bug, since the intention of this code was only to ensure a "request if not yet requested" one time on boot-up.

If users depend on tree-state in `render()`, they should manually `requestUpdate` in `connectedCallback` to ensure `render` is called when the tree state changes.

sorvell:
Filed https://github.com/Polymer/lit-element/issues/614.

oravecz:
Small issue. 

    // will exist. If so, resolve it to complete the update, otherwise
    // requestUpdate.

The comment that was left in the `connectedCallback` function still refers to the call to `requestUpdate` even after the function was removed.

PR #607, opened 2019-03-12
Handle exceptions during update so that they do not break subsequent updates

### LINKED ISSUE #262: Exceptions in update lifecycle are not handled
### Description
If an exception is throw inside updated method, the next property changes do not updates the element

This occurs because, after the error is throw, the check  `if (!this._hasRequestedUpdate)` in `_invalidate` never evaluates to true (the update Promise is never resolved)

#### Live Demo
https://codepen.io/blikblum/pen/zmEGXL?editors=1001

#### Steps to Reproduce

1. Throw an exception in updated method
2. Change an observed property

#### Expected Results
It should continue to update on property changes. 

#### Actual Results
It stops to update on property changes

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [ ] Edge
- [ ] Safari 11
- [ ] Safari 10
- [ ] IE 11

### Versions
- lit-element: v0.6.2
- webcomponents: latest

sorvell:
Fixes #262.

Catch exceptions during update and ensure the element marks itself as finished updating.

Note, this change also ensures the `updateComplete` promise is rejected if there's an exception during update.

### DIFF
--- a/src/lib/updating-element.ts
+++ b/src/lib/updating-element.ts
@@ -614,22 +614,45 @@ export abstract class UpdatingElement extends HTMLElement {
     // Mark state updating...
     this._updateState = this._updateState | STATE_UPDATE_REQUESTED;
     let resolve: (r: boolean) => void;
+    let reject: (e: Error) => void;
     const previousUpdatePromise = this._updatePromise;
-    this._updatePromise = new Promise((res) => resolve = res);
+    this._updatePromise = new Promise((res, rej) => {
+      resolve = res;
+      reject = rej;
+    });
     // Ensure any previous update has resolved before updating.
     // This `await` also ensures that property changes are batched.
-    await previousUpdatePromise;
+    try {
+      await previousUpdatePromise;
+    } catch (e) {
+      // Ignore any previous errors. We only care that the previous cycle is
+      // done. Any error should have been handled in the previous update.
+    }
     // Make sure the element has connected before updating.
     if (!this._hasConnected) {
       await new Promise((res) => this._hasConnectedResolver = res);
     }
-    // Allow `performUpdate` to be asynchronous to enable scheduling of updates.
-    const result = this.performUpdate();
+    let result = null;
+    // Trap errors in updating so that element is not in a non-functional state.
+    try {
+      // Allow `performUpdate` to be asynchronous to enable scheduling of updates.
+      result = this.performUpdate();
+    } catch (e) {
+      // Ensure subsequent updates are ok but reject this update.
+      this._markUpdated();
+      reject!(e);
+    }
     // Note, this is to avoid delaying an additional microtask unless we need
     // to.
     if (result != null &&
         typeof (result as PromiseLike<unknown>).then === 'function') {
-      await result;
+      try {
+        await result;
+      } catch (e) {
+        // Ensure subsequent updates are ok but reject this update.
+        this._markUpdated();
+        reject!(e);
+      }
     }
     // TypeScript can't tell that we've initialized resolve.
     // tslint:disable-next-line:no-unnecessary-type-assertion
--- a/src/lib/updating-element.ts
+++ b/src/lib/updating-element.ts
@@ -644,10 +644,9 @@ export abstract class UpdatingElement extends HTMLElement {
     try {
       const result = this.performUpdate();
       // If `performUpdate` returns a Promise, we await it. This is done to
-      // enable coordinating updates with a scheduler. Note, the result type is
+      // enable coordinating updates with a scheduler. Note, the result is
       // checked to avoid delaying an additional microtask unless we need to.
-      if (result != null &&
-          typeof (result as PromiseLike<unknown>).then === 'function') {
+      if (result != null) {
         await result;
       }
     } catch (e) {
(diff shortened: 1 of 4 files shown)


src/lib/updating-element.ts
```diff
@@ -613,27 +613,49 @@ export abstract class UpdatingElement extends HTMLElement {
...
     // Make sure the element has connected before updating.
     if (!this._hasConnected) {
       await new Promise((res) => this._hasConnectedResolver = res);
     }
-    // Allow `performUpdate` to be asynchronous to enable scheduling of updates.
-    const result = this.performUpdate();
+    let result = null;
+    // Trap errors in updating so that element is not in a non-functional state.
+    try {
+      // Allow `performUpdate` to be asynchronous to enable scheduling of updates.
```
justinfagnani: I don't immediately see how the comment relates to the code. What part of the statement allows `performUpdate` to be async?

src/lib/updating-element.ts
```diff
@@ -613,27 +613,49 @@ export abstract class UpdatingElement extends HTMLElement {
...
+      // Ensure subsequent updates are ok but reject this update.
+      this._markUpdated();
+      reject(e);
+    }
     // Note, this is to avoid delaying an additional microtask unless we need
     // to.
     if (result != null &&
         typeof (result as PromiseLike<unknown>).then === 'function') {
-      await result;
+      try {
```
justinfagnani: Can you merge with the previous try/catch block?

```js
try {
  const result = this.performUpdate();
  if (result != null &&
      typeof (result as PromiseLike<unknown>).then === 'function') {
    await result;
  }
} catch (e) {
  this._markUpdated();
  reject(e);
}
```

sorvell: Done, thanks.

src/lib/updating-element.ts
sorvell: Fixed.

src/lib/updating-element.ts
```diff
@@ -613,27 +613,42 @@ export abstract class UpdatingElement extends HTMLElement {
...
+      }
+    } catch (e) {
+      // Ensure subsequent updates are ok but reject this update.
+      this._markUpdated();
+      reject(e);
     }
     // TypeScript can't tell that we've initialized resolve.
-    // tslint:disable-next-line:no-unnecessary-type-assertion
-    resolve!(!this._hasRequestedUpdate);
+    resolve(!this._hasRequestedUpdate);
```
justinfagnani: remove comment

src/lib/updating-element.ts
```diff
@@ -613,27 +613,42 @@ export abstract class UpdatingElement extends HTMLElement {
...
     const previousUpdatePromise = this._updatePromise;
-    this._updatePromise = new Promise((res) => resolve = res);
+    this._updatePromise = new Promise((res, rej) => {
+      resolve = res;
+      reject = rej;
+    });
     // Ensure any previous update has resolved before updating.
     // This `await` also ensures that property changes are batched.
-    await previousUpdatePromise;
+    try {
```
justinfagnani: maybe move comment in here

src/lib/updating-element.ts
sorvell: Done

src/lib/updating-element.ts
sorvell: Done

justinfagnani:
@sorvell should we update any documentation for `updateComplete` or `requestUpdate` to mention that they could reject?

sorvell:
Added https://github.com/Polymer/lit-element/issues/608 to deal with the documentation.

sorvell:
A user error could occur in any of the following overridden spots:

- `performUpdate`
    - `shouldUpdate`
    - `update`
    - `firstUpdated`
    - `updated`

After `update`, the element should clear the update state. Given this, it makes sense to try/catch `shouldUpdate` + `update` to ensure that invariant. 

Clearing the update state means the element (a) marks itself as updated, allowing additional updates to be enqueued; and (b) clears any changed properties.

It's important to note that `firstUpdated`/`updated` can enqueue an additional update. 

If `performUpdate` throws, it may be before or after a call to `super` or `super` may not be called. In that case it's hard to decide what state this should leave the element in. It should definitely do (a) but it's unclear if it should do (b). We could set a `didUpdate` state in `update` and only do (b) if that is not set. It's unclear if this is worth the trouble, however.

justinfagnani:
Offline review: catching `shouldUpdate` and `update` directly in `performUpdate` is better so we can ensure `_markUpdated` is called exactly once. Add documentation to `performUpdate` to say that any override must call `super.performUpdate`, even in the case of an exception in the override.

src/lib/updating-element.ts
```diff
@@ -613,27 +622,38 @@ export abstract class UpdatingElement extends HTMLElement {
...
-    if (result != null &&
-        typeof (result as PromiseLike<unknown>).then === 'function') {
-      await result;
+    try {
+      const result = this.performUpdate();
+      // If `performUpdate` returns a Promise, we await it. This is done to
+      // enable coordinating updates with a scheduler. Note, the result type is
+      // checked to avoid delaying an additional microtask unless we need to.
+      if (result != null &&
+          typeof (result as PromiseLike<unknown>).then === 'function') {
```
kevinpschaaf: This code seems superfulous

PR #610, opened 2019-03-12
Ensure attribute reflected from property is correct

### LINKED ISSUE #592: Boolean Attribute no longer reflected after adding and removing the attribute

sorvell:
Fixes #592. Ensures a reflecting property set immediately after a corresponding attribute reflects correctly.

### DIFF
--- a/src/lib/updating-element.ts
+++ b/src/lib/updating-element.ts
@@ -579,16 +579,21 @@ export abstract class UpdatingElement extends HTMLElement {
    */
   requestUpdate(name?: PropertyKey, oldValue?: unknown) {
     let shouldRequestUpdate = true;
-    // if we have a property key, perform property update steps.
-    if (name !== undefined && !this._changedProperties.has(name)) {
+    // If we have a property key, perform property update steps.
+    if (name !== undefined) {
       const ctor = this.constructor as typeof UpdatingElement;
       const options =
           ctor._classProperties!.get(name) || defaultPropertyDeclaration;
       if (ctor._valueHasChanged(
               this[name as keyof this], oldValue, options.hasChanged)) {
-        // track old value when changing.
-        this._changedProperties.set(name, oldValue);
-        // add to reflecting properties set
+        // Track old value when changing.
+        if (!this._changedProperties.has(name)) {
+          this._changedProperties.set(name, oldValue);
+        }
+        // Add to reflecting properties set.
+        // Note, it's important that every change has a chance to add the
+        // property to `_reflectingProperties`. This ensures setting
+        // attribute + property reflects correctly.
         if (options.reflect === true &&
             !(this._updateState & STATE_IS_REFLECTING_TO_PROPERTY)) {
           if (this._reflectingProperties === undefined) {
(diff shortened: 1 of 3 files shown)


src/lib/updating-element.ts
```diff
@@ -579,24 +579,29 @@ export abstract class UpdatingElement extends HTMLElement {
...
   requestUpdate(name?: PropertyKey, oldValue?: unknown) {
     let shouldRequestUpdate = true;
-    // if we have a property key, perform property update steps.
-    if (name !== undefined && !this._changedProperties.has(name)) {
+    // If we have a property key, perform property update steps.
+    if (name !== undefined) {
       const ctor = this.constructor as typeof UpdatingElement;
       const options =
           ctor._classProperties!.get(name) || defaultPropertyDeclaration;
       if (ctor._valueHasChanged(
```
kevinpschaaf: Unrelated to PR, but why do we pass `options.hasChanged` through a static method that just calls it?  It's private so it can't be overridden by user...

src/lib/updating-element.ts
```diff
@@ -579,24 +579,29 @@ export abstract class UpdatingElement extends HTMLElement {
...
               this[name as keyof this], oldValue, options.hasChanged)) {
-        // track old value when changing.
-        this._changedProperties.set(name, oldValue);
-        // add to reflecting properties set
+        // Track old value when changing.
```
justinfagnani: nit: This reads more like a "what" comment than a "why". It'd be nice to either say why we only record the once, or remove the comment since the code seems pretty self-documenting on what it's doing.

src/lib/updating-element.ts
```diff
@@ -579,24 +579,29 @@ export abstract class UpdatingElement extends HTMLElement {
...
+        // attribute + property reflects correctly.
         if (options.reflect === true &&
             !(this._updateState & STATE_IS_REFLECTING_TO_PROPERTY)) {
           if (this._reflectingProperties === undefined) {
             this._reflectingProperties = new Map();
           }
           this._reflectingProperties.set(name, options);
         }
-        // abort the request if the property should not be considered changed.
+        // Abort the request if the property should not be considered changed.
```
justinfagnani: what line is this commenting? The last line of the if case or the else case?

src/lib/updating-element.ts
sorvell: I believe we intend to make this overridable at some point.

Issue #488, opened 2019-01-28
Reconsider usage of `Number`s in CSSResult (based on real life painful use cases)


tlouisse:
### Description
Sorry to bother you again with the CSSResult variables issue, but I was hoping there is some room left for improvement of Developer Experience, after seeing the remark of @mikesamuel: https://github.com/Polymer/lit-element/issues/451#issuecomment-457953258

We appreciate your considerations with regard to security, but we currently are experiencing some drawbacks as far as Developer Experience is concerned.
To illustrate this, see the following example (very verbose and thus harder to read and maintain):
```js
import { css } from 'lit-element';
import { unsafeCss } from 'lit-element/lib/css-tag.js';
import { small, large, containerPadding } from '../variables/breakpoints.js';
// This would become quite unreadable:
export const gridCssModule = css`  
  .grid-container {
    margin: auto;
    max-width: ${unsafeCss(Number(small.cssText.slice(0,-2)) - (Number(containerPadding.cssText.slice(0,-2) * 2)))}px;
  }
  @media (min-width: ${small}) {
    .grid-container {
      max-width: ${unsafeCss(Number(large.cssText.slice(0,-2)) - Number((containerPadding.cssText.slice(0,-2)) * 2))}px;
    }
  }
`;
```
https://stackblitz.com/edit/css-variables-for-computations?file=style%2Fmodules%2Fgrid-css-module_numbersNotAllowed.js

My apologies for annoying you with a question that was already asked before, but we just wanted to show you some of our 'real life implications'.
So the derived question then would be: **can Numbers on their own be a security issue?**

If not, we could rewrite our example above to:
```js
import { css } from 'lit-element';
import { small, large, containerPadding } from '../variables/breakpoints.js';
export const gridCssModule = css`  
  .grid-container {
    margin: auto;
    max-width: ${small - (containerPadding * 2)}px;
  }
  @media (min-width: ${small}px) {
    .grid-container {
      max-width: ${large - (containerPadding * 2)}px;
    }
  }
`;
```
https://stackblitz.com/edit/css-variables-for-computations?file=style%2Fmodules%2Fgrid-css-module_numbersAllowed.js 

#### Live Demo
See above

sorvell:
It may be possible to make this change, but we'll need to do a careful security review first.

However, I want to make sure I understand the examples given above. The intention of `unsafeCss` is to allow you to use any value. To be clear, the function is named this way so that it's clear that you may be incurring some risk.

[...]

PR #626, opened 2019-03-20
feat: allow JavaScript numbers in css template literal

daKmoR:
This allows for real JavaScript numbers within the css tagged template literal.

### Reference Issue
closes: #488

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
    PRs should document their user-visible changes (if any) in the
    Unreleased section, uncommenting the header as necessary.
 -->
-<!-- ### Added -->
+### Added
+* css tagged template literals now allow numbers to be used in expressions ([#488](https://github.com/Polymer/lit-element/issues/488)).
+
 <!-- ### Changed -->
 <!-- ### Removed -->
 <!-- ### Fixed -->
--- a/src/lib/css-tag.ts
+++ b/src/lib/css-tag.ts
@@ -60,9 +60,11 @@ export const unsafeCSS = (value: unknown) => {
   return new CSSResult(String(value), constructionToken);
 };
 
-const textFromCSSResult = (value: CSSResult) => {
+const textFromCSSResult = (value: CSSResult | number) => {
   if (value instanceof CSSResult) {
     return value.cssText;
+  } else if (typeof value === 'number') {
+    return value;
   } else {
     throw new Error(
         `Value passed to 'css' function must be a 'css' function result: ${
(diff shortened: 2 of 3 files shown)


Westbrook:
@daKmoR this looks great! Hopefully wouldn't change anything, but it would be nice to see a test were a number in used in a media query. That _is_ supported by this addition, right?

Also, I'd love to see at least a first pass on adding this to the docs. Though maybe that's best relegated to a follow up PR.

daKmoR:
> but it would be nice to see a test were a number in used in a media query. That is supported by this addition, right?

yes - but why would you like to see a test for a media query there? testing media queries is a little tough and testing the string
```js
const breakpoint = 600;
expect(css`@media screen and (max-width: ${breakpoint}px) {`.cssText).to.equal('@media screen and (max-width: 600px) {');
```
seems kinda redundante with the existing test.

so that's probably something to add to the documentation?

Westbrook:
I’d like to see it because it was one of the most compelling reasonings I saw for this addition and because I could see (naively) support for this very easily falling victim to regression over time. 

You make a good point that the difficulty might outweigh the benefits however it’s not the media query per say that I’d suggest testing as much as that it writes a media query that gets applied. However, it’s likely that I’m just overworried and you’re right that documentation is more in order.

justinfagnani:
I don't think we need additional tests on the CSS side. As long as we're producing the expected text, the CSS parser will take it from there.

CHANGELOG.md
```diff
@@ -9,7 +9,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
    PRs should document their user-visible changes (if any) in the
    Unreleased section, uncommenting the header as necessary.
 -->
-<!-- ### Added -->
+### Added
+* css tagged template literal now allows for JavaScript numbers but not
```
justinfagnani:
```suggestion
* css tagged template literals now allow numbers (but not Infinity or NaN) to be used in expressions
```

daKmoR:
we probably should just adopt [Expressions in static styles](https://lit-element.polymer-project.org/guide/styles).

But as it's also altered in https://github.com/Polymer/lit-element/pull/600 it's probably better to do it step by step to avoid an unneeded merge conflict.

arthurevans:
Recommend we fix the docs in a follow-on PR. #600 is not correct right now.

daKmoR:
friendly reminder :hugs: 

are there any blockers for this PR? anything I should adjust?

src/lib/css-tag.ts
```diff
@@ -60,9 +60,14 @@ export const unsafeCSS = (value: unknown) => {
   return new CSSResult(String(value), constructionToken);
 };
 
-const textFromCSSResult = (value: CSSResult) => {
+const textFromCSSResult = (value: CSSResult | number) => {
   if (value instanceof CSSResult) {
     return value.cssText;
+  } else if (typeof value === 'number' && isFinite(value) && value === value) {
```
justinfagnani: Since there's no security risk from Infinity or NaN, I think we should allow them rather than throw. They might not result in valid CSS, but at least we wouldn't throw and abort rendering. It's also less code and checks.

daKmoR: updated 🤗

justinfagnani:
Thanks @daKmoR!

PR #636, opened 2019-03-28
[docs] Remove z-index on alert style

### LINKED ISSUE #635: Links inside alert boxes can't be clicked
The .alert boxes we use for notes are currently set to z-index: -1. Not sure why, but it prevents you from clicking a link inside the box.

For example, see the first note here:

https://lit-element.polymer-project.org/guide/styles

keanulee:
Fixes #635. Don't see a reason why `z-index: -1` needs to be there.

Staged at https://20190328t150824-dot-polymer-lit-element.appspot.com/guide/styles

### DIFF
--- a/docs/css/main.css
+++ b/docs/css/main.css
@@ -417,7 +417,6 @@ article ol li {
 
 .alert {
   position: relative;
-  z-index: -1;
   margin: 24px 0;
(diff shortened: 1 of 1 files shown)


keanulee:
Deployed. Will update the other sites now.

PR #640, opened 2019-04-05
docs defaultConverter example use lower camel case

guilala:


guilala:
I signed it! (the CLA)

guilala:
Confirmed alternate email

PR #645, opened 2019-04-10
Move LitElement implementation inside lib/

ChALkeR:
Unifies `LitElement` / `UpdatingElement` implementation location, and allows more granular imports than just importing `LitElement` alongside with all the other helpers.

E.g. this allows importing LitElement class without bundling `decorators.js` helper file (which is ~10% of the lib size).


justinfagnani:
Sorry it took so long to respond. I'd be open to this, but the PR needs to be redone with the current code.

justinfagnani:
3.0 separates importing LitElement from decorators and we added a top-level `lit-element.js` module here to follow the same import structure. Thanks for the PR though!

PR #648, opened 2019-04-12
removing unnecessary closing div tag

dlockhart:
This extra `</div>` is visible in the public docs here:
https://lit-element.polymer-project.org/guide/styles

Probably an artifact from an old `<div class="alert alert-info">`?

Screen Shot 2019-04-12 at 4 26 44 PM


justinfagnani:
Thanks @dlockhart!

PR #655, opened 2019-04-15
Fix TypeScript configuration in publish docs

### LINKED ISSUE #574: [docs] Publish: Bug in TypeScript configuration
### Description
"ES2017" is not a valid value for the TypeScript configuration. It will result in:

#### Steps to Reproduce
Follow the guide here https://lit-element.polymer-project.org/guide/publish#transpiling-with-typescript and try to build.

#### Expected Results
No error thrown.

#### Actual Results
```
TS6046: Argument for '--module' option must be: 'none', 'commonjs', 'amd', 'system', 'umd', 'es6', 'es2015', 'esnext'.
```

### Versions
- lit-element": "^2.0.1"

AdamVig:
`tsconfig.json` does not accept a value of `"es2017"` for `module`. Its schema also specifies allowed values in all
lowercase, so `es2015` is valid but `ES2015` is invalid (though the latter is accepted by the runtime).
  
Fixes #574.

### DIFF
--- a/docs/_guide/publish.md
+++ b/docs/_guide/publish.md
@@ -37,10 +37,10 @@ The following JSON sample is a partial tsconfig.json that uses recommended optio
 
 ```json
   "compilerOptions": {
-    "target": "ES2017",
-    "module": "ES2017",
+    "target": "es2017",
+    "module": "es2015",
     "moduleResolution": "node",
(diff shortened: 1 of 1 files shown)


AdamVig:
@justinfagnani Is there anything I can do to help get this merged?

PR #657, opened 2019-04-16
Adds documentation on how to use custom accessors with  @property decorator

### LINKED ISSUE #519: [docs] how to use property accessors with @property decorator
Hi,

I would love to have some documentation on how to use property accessors with `@property`-decorator. I like how I can define a property with a decorator. 
````typescript
@property
info: string = 'Hello world!';
````
However, it is not documented how this works with custom accessors. I have found out myself that putting it on the getter like this works:
````typescript
 _info: string = '';

  @property({ type: String })
  public get info(): string {
    return this._info;
  }
  public set info(value: string) {
    const oldValue = this.info;
    this._info = value;
    this.requestUpdate('info', oldValue);
  }
````
Is this considered best practice in this case? I wouldn't like this, because for other properties I could still write it on the class field. I would prefer to be able to put all property decorators on the class fields. Or am I supposed to use `static get properties()` again like this:
````typescript
static get properties(): PropertyDeclarations {
return {'info': {type: String}}
}
  _info: string = '';
  public get info(): string {
    return this._info;
  }
  public set info(value: string) {
    const oldValue = this.info;
    this._info = value;
    this.requestUpdate('info', oldValue);
  }
````
It would be great if there'd be documentation on how to handle this. It feels a bit like mostly `static get properties()` is addressed, so a statement on whatever decorators are considered an equal alternative would be nice too. 

If no solution to this problem currently exists, I would have a suggestion as well:  
````typescript
//Override property name manually. 
@property({ type: String, name: 'info' })
 _info: string = '';

  public get info(): string {
    return this._info;
  }
  public set info(value: string) {
    const oldValue = this.info;
    this._info = value;
    this.requestUpdate('info', oldValue);
  }

### PR DESCRIPTION
**Pull request by Christian24**
This adds an example to the properties guide that for custom accessors the `@property` decorator can be put on the getter. Closes #519.

### DIFF
--- a/docs/_guide/properties.md
+++ b/docs/_guide/properties.md
@@ -415,6 +415,22 @@ The setters that LitElement generates automatically call `requestUpdate`. If you
 {% include projects/properties/customsetter/my-element.js %}
 ```
 
+If you want to use your own property accessor with the `@property` decorator, you can achieve this by putting the decorator on the getter:
+```ts
+set myProp(value) {
+  const oldValue = this.myProp;
+  // Implement setter logic here...
+  this.requestUpdate('myProp', oldValue);
+}
+@property({ type: String })
+get myProp() { ... }
+
+...
+
+// Later, set the property
+this.myProp = 'hi'; // Invokes your accessor
+```
+
 ### Prevent LitElement from generating a property accessor {#accessors-noaccessor}
 
 In rare cases, a subclass may need to change or add property options for a property that exists on its superclass.
--- a/docs/_guide/properties.md
+++ b/docs/_guide/properties.md
@@ -417,20 +417,19 @@ The setters that LitElement generates automatically call `requestUpdate`. If you
 
 If you want to use your own property accessor with the `@property` decorator, you can achieve this by putting the decorator on the getter:
 ```ts
-set myProp(value) {
-  const oldValue = this.myProp;
-  // Implement setter logic here...
-  this.requestUpdate('myProp', oldValue);
-}
-@property({ type: String })
-get myProp() { ... }
+   _myProp: string = '';
 
-...
-
-// Later, set the property
-this.myProp = 'hi'; // Invokes your accessor
+  @property({ type: String })
+  public get myProp(): string {
+    return this._myProp;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Christian24** [comment]
@gallafent I do agree. However, I tried to stay true to the original listing, which does not show this either. I did adapt the sample code now to basically mirror what I had in #519.

**Comment by Christian24** [comment]
@justinfagnani what do you think?

**Comment by arthurevans** [comment]
This needs some extra compilation, right? Either TypeScript compiler or you need to have Babel/whatever set up to support decorators? Do we talk about that anywhere currently @justinfagnani @katejeffreys ?

**Comment by Christian24** [comment]
Well, I would say that https://lit-element.polymer-project.org/guide/properties#declare-properties-with-typescript-decorators does imply it does need Typescript. However, I am unaware if Babel is mentioned anywhere. Maybe it would be good to add that here: https://lit-element.polymer-project.org/guide/start#use-litelement-typescript-decorators? Maybe it would also be a good idea to replace Typescript decorators with decorators everywhere? Since this also applies to Babel decorators too. I think it makes sense to open a new docs issue for that.

**Comment by Christian24** [comment]
@arthurevans I created a new issue https://github.com/Polymer/lit-element/issues/692 regarding decorators being mostly described as a Typescript feature in the docs.

File: docs/_guide/properties.md [hunk 1]
```diff
@@ -415,6 +415,21 @@ The setters that LitElement generates automatically call `requestUpdate`. If you
...
+  public get myProp(): string {
+    return this._myProp;
+  }
+  public set myProp(value: string) {
+    const oldValue = this.myProp;
+    this._myProp = value;
+    this.requestUpdate('myProp', oldValue);
+  }
+```
+As you can see this works almost exactly like `static get properties()`, except the property declaration goes on the getter.
```
arthurevans: Shouldn't this last sentence say that it works almost exactly like using the decorator on a field declaration?

arthurevans: Actually not sure the sentence adds anything, but I'm going to go ahead and merge and tweak.

PR #659, opened 2019-04-18
Add ugly type assertions for `this`

rictic:
Because of https://github.com/google/closure-compiler/issues/3340 – `this` has an unknown type to closure compiler in static methods. In order to be compatible with the Closure Compiler + JSConformance `BanUnknownThis` check, we must explicitly add type annotations wherever we access `this` in a static method.

Yes, this is pretty awful from a legibility perspective. We could improve legibility at the cost of additional size by doing:

```typescript
const self = this as typeof LitElement;
```

I went with the zero-size-cost option here.

### DIFF
--- a/src/lib/css-tag.ts
+++ b/src/lib/css-tag.ts
@@ -60,7 +60,7 @@ export const unsafeCSS = (value: unknown) => {
   return new CSSResult(String(value), constructionToken);
 };
 
-const textFromCSSResult = (value: CSSResult | number) => {
+const textFromCSSResult = (value: CSSResult|number) => {
   if (value instanceof CSSResult) {
     return value.cssText;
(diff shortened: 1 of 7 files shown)


justinfagnani:
Ugly, but like you said, 0 byte impact. What do we do to ensure we don't regress here?

justinfagnani:
Also, could we file a bug against tsickle asking them to automatically output the `this` types for Closure?

rictic:
Looks like one was just filed a couple days ago: https://github.com/angular/tsickle/issues/1015

rictic:
It should be possible to compile with closure compiler as part of the build process. At the moment the round trip into the internal repo is fast enough though that we should catch and fix regressions quickly.

PR #674, opened 2019-05-09
Remove ugly type assertions for `this`

rictic:
This undoes f552fc47b12bf3e272aaeafe03d9d73bc4fff1f5 now that https://github.com/google/closure-compiler/issues/3340 has been fixed.


rictic:
I've verified that this builds without error internally.

Issue #641, opened 2019-04-05
Need for ShadyCSS


LarsDenBakker:
`LitElement` is no longer supporting `@apply` for styles defined using the static `styles` property, but it's still using `shady-render` and calling ` window.ShadyCSS.styleElement(this);` in the `connectedCallback`. 

Why is that still needed? If it's only for supporting `@apply` in regular `<style>` tags I think this could be made configurable to get a nice performance boost for most users.

The reason I am asking is I am migrating a large code from Polymer to LitElement. Our main entry point does not use any polymer, so it would only be loaded once a user lands on some legacy pages. I'd also like to avoid loading `ShadyCSS` later on as well, but that might throw off some elements built with `LitElement` if `ShadyCSS` suddenly shows up after they have already rendered initially. I'd prefer it if `LitElement` did not deal with `ShadyCSS` at all on modern browsers.

CaptainCodeman:
It's still needed to support polyfilled browers without native support for shadow DOM, it's not _just_ for `@apply`

LarsDenBakker:
Of course, but as far as I can tell that is covered within the static styles already.

jhpratt:
See also: #612.

PR #695, opened 2019-05-30
chore: only use shady-render when using ShadyCSS

LarsDenBakker:
Closes https://github.com/Polymer/lit-element/issues/641

`shady-render` is used also on browsers which don't have `ShadyCSS` loaded. This is wasteful, as `shady-render` does some additional work. This change conditionally picks which render method to use based on whether `ShadyCSS` is loaded. 

This is a non-breaking change, as `shady-render` does not do any functional work on browsers which doesn't have `ShadyCSS` loaded. It will also still work for people who want to use `ShadyCSS` on browsers with native shadow dom (for example, to use `@apply`).


justinfagnani:
Obsolete now that 3.0 drastically changes how Shady CSS support is loaded.

PR #698, opened 2019-06-04
Rewrite styles page

### LINKED ISSUE #353: [docs] Review CSS Styles page for technical accuracy

### LINKED ISSUE #418: [docs] Document style api

### LINKED ISSUE #534: [docs] Styles: lit-html styleMap and classMap
Mention & point to `styleMap` and `classMap`

### LINKED ISSUE #542: [docs] Styles: make it simpler and friendlier
Styling a component feels more complicated than it should. The doc leads with a bunch of jargon and concepts that will be unfamiliar to new users of Web Components. We'd like to communicate that it's super easy -- here's a tiny code snippet that does exactly what you want and sets you on the right path. The stuff about shadow DOM and host styles and all of that should feel like an elegant and better way of doing things. Currently it comes across more as overwhelming and a bit intimidating. Obviously not intended. It's not hard! Let's make it seem as ergonomic as it is.

### LINKED ISSUE #561: [docs] Note that external stylesheets are not supported on polyfill
### Description
Under https://lit-element.polymer-project.org/guide/styles#external-stylesheet, it states:
> ...you can load an external stylesheet into a shadow root with a <link> element

The ShadyDOM scoping polyfill does not support scoping styles referenced via `<link rel="stylesheet">` in a shadow root.  Any styles loaded via the external stylesheet will _not_ be correctly scoped to the element under polyfill.  This is an open issue (https://github.com/webcomponents/shadycss/issues/97) that we haven't considered high-priority to support since the FOUC issue of the native feature makes the ability to load external stylesheets somewhat dubious.

This should at least be added in the list of caveats, or else not recommended at all.

### LINKED ISSUE #570: Nesting in styles section seems off
This section:

_Write CSS styles for a host element and its shadow DOM_

is buried at the bottom of _Styling options for component developers_. I think it should be more prominent. One option would be to drop the "Define styles in ..." sections down to H4s:

- Where to define your styles
    - Define styles in a static styles property
    - Define styles in a style element
    - Define styles in an external stylesheet
- Write CSS styles for a host element and its shadow DOM

Another option would be to break this section into two siblings: one that describes options (where to define styles), and one that covers the nitty-gritty of writing those styles.

katejeffreys:
Ready for review.

@arthurevans  

Fixes #542 
Fixes #561 
Fixes #418 
Fixes #353 
Fixes #570 
Fixes #534

Staged: https://polymer-lit-element-staging.appspot.com/guide/styles


katejeffreys:
Stylemap and classmap stuff in progress https://stackblitz.com/edit/lit-element-zcbgcy?file=src%2Findex.js

PR #709, opened 2019-06-11
Release lit-element 2.2.0

aomarks:
Also:
- Add a script that checks that lit-element.ts contains the latest version, to help ensure our tracker is current before publishing, and run in `prepublishOnly`.
- Add `npm run lint && npm test` to `prepublishOnly`.
- Remove `build` from `prepublishOnly` script, because it's already in `prepare`, which runs before `prepublishOnly`.
- Run clang-format.


aomarks:
> BTW this seems to be a good chance to fix version (still pointing to 2.0.1):

Thanks, done, and added a `prepublishOnly` script to help ensure we always keep this up to date.

PR #710, opened 2019-06-13
Alternative styles page rewrite

katejeffreys:
Staged at https://20190613t124942-dot-polymer-lit-element.appspot.com


docs/_guide/styles.md
```diff
@@ -8,141 +8,233 @@ slug: styles
 * ToC
 {:toc}
 
-This page is a guide to the styling options available for LitElement components. 
+## Define scoped stylesheets {#scoped}
```
jbingham: This page now reads as a detailed reference that assumes you already have a conceptual understanding of how to style components with LitElement. There's not an overview or easy to understand example. Is this intentional? Can we include the introductory overview here, or link to it?

arthurevans: Will add an overview & simple example, yes. IIRC we decided on a somewhat different outline.

PR #712, opened 2019-06-13
Render any value returned by `#render`.

### LINKED ISSUE #168: Allow empty template results?
Currently moving over to lit straight from polymer 2, we commonly have:

```js
if (!this.prop) {
  return html``;
}
return html`some content`;
```

Is it worth us allowing a `null` return type in `render` so we can have an empty template?

and inside lit, keep an `emptyTemplate` constant which is re-used (if needed)? 

so we can have `render(): TemplateResult|null`?

edit:

having thought again about this, in reality it is probably less efficient to have null checks in the core. i guess the only solution to this would be to have an `EmptyTemplate` constant which is an already new'd up template result.

cc @justinfagnani

### LINKED ISSUE #618: Styles don't get added if render() returns undefined at point of attach to DOM
### Description
If `render()` returns `undefined` at the point where it is attached the `static get styles()` attribute is still called but never gets added to the shadow root.

#### Live Demo
https://stackblitz.com/edit/lit-element-example-debgxe?embed=1&file=index.js

#### Steps to Reproduce
The problem happens when you have a control that you don't want to render anything when properties are not set, for instance:

```
static get styles() { ... }

render() {
    if(!this.thingToRender)
        return;

    return html` ...${this.thingToRender.prop1} ...${this.thingToRender.prop2}  `;
}
```

If this control is attached to the DOM and then `thingToRender` is set `render()` calls as expected and populates the HTML, but the styles are never added.

Returning empty content appears to workaround this:

```
render() {
    if(!this.thingToRender)
        return html``;
```

So another option would be to update the documentation to state that `render()` should never return `undefined`.

#### Expected Results
Either the styles should be added on attach, or if skipped because no content it should be added after the first `render()` that contains content.

#### Actual Results
Styles are not added to the component.

### Browsers Affected
I don't think this is browser specific.

### Versions

- lit-element: v2.0.1
- webcomponents: v2.2.7

bicknellr:
As of Polymer/lit-html#910, `render` accepts any type. This PR causes `LitElement` to render anything returned by `render`.

- [x] Wait for Polymer/lit-html#910 to be in a lit-html release, update dependency here.
- [x] Update the change log.

---

Fixes #618, Fixes #168


43081j:
can you also change the base `render` method to have a return type of `unknown` rather than `TemplateResult|void`?

justinfagnani:
@bicknellr is this ready to go in now?

src/lit-element.ts
```diff
@@ -68,10 +67,10 @@ export class LitElement extends UpdatingElement {
    * optimizations. See updating-element.ts for more information.
    */
   protected static['finalized'] = true;
+
   /**
-   * Render method used to render the lit-html TemplateResult to the element's
-   * DOM.
-   * @param {TemplateResult} Template to render.
+   * Render method used to render the value to the element's DOM.
+   * @param {unknown} The value to render.
```
justinfagnani: These need the property names as the first word following the type.

src/lit-element.ts
```diff
@@ -236,10 +232,12 @@ export class LitElement extends UpdatingElement {
...
-   * Invoked on each update to perform rendering tasks. This method must return
-   * a lit-html TemplateResult. Setting properties inside this method will *not*
-   * trigger the element to update.
+   * Invoked on each update to perform rendering tasks. This method may return
+   * any value renderable by lit-html's NodePart - typically a TemplateResult.
+   * Setting properties inside this method will *not* trigger the element to
+   * update.
    */
-  protected render(): TemplateResult|void {
+  protected render(): unknown {
```
justinfagnani: Can you elide the return if you type this as `unknown|void`?

src/test/lit-element_styling_test.ts
```diff
@@ -798,6 +798,62 @@ suite('Static get styles', () => {
         const bodyStyles = `${cssModule}`;
         assert.equal(bodyStyles, '.my-module { color: yellow; }');
       });
+
+  test.only(
```
justinfagnani: remove `.only`

src/lit-element.ts
bicknellr: Doesn't seem like it:
```
src/lit-element.ts:241:23 - error TS2355: A function whose declared type is neither 'void' nor 'any' must return a value.                                                                                         

241   protected render(): unknown|void {
                          ~~~~~~~~~~~~
```

src/test/lit-element_styling_test.ts
bicknellr: Good catch!

bicknellr: Fixed.

src/lit-element.ts
bicknellr: Fixed.

src/lit-element.ts
justinfagnani: It's a shame to add non-observable, non-type annotation code to satisfy the type-checker. Can we cast our way out of this?

bicknellr: I don't know how I would cast the function in place since class syntax isn't expression based (and I assume we don't want to bind it / remove it from the prototype by making it a property), so I tried pulling it out of the class as this:

```javascript
LitElement.prototype.render = function() {} as () => unknown;
```

But then I get another couple of errors that seem like TS is no longer recognizing that `render` is also on the prototype:

```
src/lit-element.ts:219:18 - error TS2576: Property 'render' is a static member of type 'LitElement'

219             this.render(),
                     ~~~~~~

src/lit-element.ts:243:22 - error TS2576: Property 'render' is a static member of type 'LitElement'

243 LitElement.prototype.render = function() {} as () => unknown;
                         ~~~~~~
```

I'm not familiar with pre-classes TS, so maybe that's not the right way to do it?

bicknellr:
The benchmarks check is stuck (it's been in the 'queued' state all weekend) but I don't know how to restart it.

src/lit-element.ts
bicknellr: (from offline) I'm going to leave the `return undefined;` as is and assume that any given minifier should be smart enough to remove it.

darionco:
#618 is extremely serious since it affects FireFox and Safari (and others), that's  over 20% of the total market share!

Please expedite this PR!!!

src/lit-element.ts
```diff
@@ -194,14 +195,11 @@ export class LitElement extends UpdatingElement {
    */
   protected update(changedProperties: PropertyValues) {
     super.update(changedProperties);
-    const templateResult = this.render() as unknown;
-    if (templateResult instanceof TemplateResult) {
```
bicknellr: Does it matter that we're no longer providing a way to tell LitElement not to do any rendering for a particular update? I could see people returning undefined (implicitly) to signal this. It's been a while since I was last working on this PR so I don't remember if / what we decided here.

bicknellr: (from offline) Use lit-html's [`noChange`](https://github.com/Polymer/lit-html/blob/21999f5222a26247e442caa73fb6ce0d52410d17/src/lib/part.ts#L49) or [`nothing`](https://github.com/Polymer/lit-html/blob/21999f5222a26247e442caa73fb6ce0d52410d17/src/lib/part.ts#L54) for these cases.

bschlenk:
Isn't this a breaking change? My team was not returning anything from render in our non-shadow-dom components that accept children, which would previously prevent render from happening. Now, it blows away all children of the component. I figured out a workaround, which is to return `super.render()`, but this isn't documented anywhere.

PR #723, opened 2019-07-02
Update fromAttribute for Arrays and Objects when no valid value is set

### LINKED ISSUE #722: Uncaught SyntaxError: Unexpected token g in JSON at position 0
### Description
Once a new element attribute of type Array or Object is set but the value is a JavaScript object value added to the attribute with no .property notation the following exception is raised: 
`Uncaught SyntaxError: Unexpected token g in JSON at position 0`

I discovered it because of a misconfiguration but as no clear exception is raised it took some time to discover what exactly was the issue.

#### Live Demo
https://lit-element-example-3c7ojq.stackblitz.io

#### Steps to Reproduce
1. Create `my-element` with an Array type attribute and render it as follows
`render(html`<my-element moods=${['a', 'b']}></my-element>`, document.body);`

#### Expected Results
As the set value is wrong I expect the attribute to be set to a initial value `[]` or `undefined` but no exception raised.

#### Actual Results
Console exception:
`Uncaught SyntaxError: Unexpected token g in JSON at position 0`

### Browsers Affected
I did not checked all but I believe the same shall happen in all major browsers.

- [X] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari 11
- [ ] Safari 10
- [ ] IE 11

### Versions
- lit-element: 2.1.0

f-santos:
For better understanding on the issue please check #722 

This happens because `JSON.parse('')` will be fired and empty string is not a valid JSON.

My changes make an empty attribute to be converted to an empty array or an empty object what seems the expected behaviour in these cases not breaking internal methods.

### Alternative solution

Change the default value to return `undefined` instead of the initial value as in my current solution.

### Reference Issue
Fixes #722

### DIFF
--- a/src/lib/updating-element.ts
+++ b/src/lib/updating-element.ts
@@ -151,8 +151,9 @@ export const defaultConverter: ComplexAttributeConverter = {
       case Number:
         return value === null ? null : Number(value);
       case Object:
+        return value === '' ? {} : JSON.parse(value!);
       case Array:
-        return JSON.parse(value!);
+        return value === '' ? [] : JSON.parse(value!);
     }
     return value;
   }


f-santos:
I signed it!

src/lib/updating-element.ts
```diff
@@ -151,8 +151,9 @@ export const defaultConverter: ComplexAttributeConverter = {
       case Number:
         return value === null ? null : Number(value);
       case Object:
+        return value === '' ? {} : JSON.parse(value!);
```
justinfagnani: I think rather than this, which wouldn't fix the demo linked in the issue, we want to put a try/catch around the `JSON.parse()`.

felipeplets:
This PR becomes obsolete since #888

felipeplets:
@justinfagnani @f-santos is an old user of mine which was deleted, so feel free to close this PR and consider #888 to fix the issue

justinfagnani:
In Lit 2 we catch all JSON parse exceptions.

PR #733, opened 2019-07-13
Fixing formatting in use.md

praneshr:
Fixing a formatting issue in the use.md page


justinfagnani:
@praneshr seems like a good fix, why'd you close it?

praneshr:
Forgot to remove the extra spaces 🤦‍♂ 

image

Created another PR with the right change https://github.com/Polymer/lit-element/pull/734

Issue #732, opened 2019-07-12
LitElement sub-optimal/broken with ES5 JSC compilation


aomarks:
**TL;DR** LitElement compiled to ES5 is at best sub-optimal when the Closure JS Compiler property collapsing optimization is enabled, and (interestingly) *actually broken* when property collapsing is *disabled*. We might need to update the way we handle the `finalized`, `properties`, and `styles` static properties.

**Collapsing enabled**

LitElement has code roughly like this (actual code [here](https://github.com/Polymer/lit-element/blob/46c8f404937672ed10587124ec3ddee82fd005fe/src/lib/updating-element.ts#L216)):

```js
class UpdatingElement {
  // UpdatingElement doesn't need to be finalized.
  static finalized = true;

  /** @nocollapse */
  static finalize() {
    if (this.hasOwnProperty('finalized')) {
      return;
    }
    this.finalized = true;
    // do class setup ...
    const superCtor = Object.getPrototypeOf(this);
    if (superCtor.finalize) {
      superCtor.finalize();
    }
  }
}

class MyElement extends UpdatingElement {}
```

Which JS Compiler will transpile roughly to:

```js
var UpdatingElement = function() {};

UpdatingElement.finalize = function() {
  if (this.hasOwnProperty('finalized')) {
    return;
  }
  this.finalized = true;
  // do class setup ...
  const superCtor = Object.getPrototypeOf(this);
  if (superCtor.finalize) {
    superCtor.finalize();
  }
};

var MyElement = function() {
  UpdatingElement.apply(this, arguments);
};
MyElement.prototype = Object.create(UpdatingElement.prototype);
Object.setPrototypeOf(MyElement, UpdatingElement);
MyElement.finalize = UpdatingElement.finalize;

var UpdatingElement$finalized = true;
var MyElement$finalized = null;
```

So, the error here is that `finalize()` doesn't see that `UpdatingElement` has already been finalized, because the `finalized` static property was optimized to a variable -- but not cleverly enough to update the code in the `finalize()` static method. This is not *super* surprising, given
https://google.github.io/styleguide/jsguide.html#features-classes-static-methods implies that the compiler's understanding of static properties is pretty limited (and presumably this same limitation is why we already put `@nocollapse` on the finalize static method itself).

[...]

PR #735, opened 2019-07-15
Fix compatibility with JSC static property optimizations

aomarks:
Fixes https://github.com/Polymer/lit-element/issues/732

### DIFF
--- a/src/lib/updating-element.ts
+++ b/src/lib/updating-element.ts
@@ -211,10 +211,11 @@ export abstract class UpdatingElement extends HTMLElement {
   private static _attributeToPropertyMap: AttributeMap;
 
   /**
-   * Marks class as having finished creating properties.
+   * This property was previously used to record the classes that have already
+   * been finalized. This property is no longer used, but is still declared here
+   * for backwards compatability in TypeScript typings.
    */
-  protected static finalized = true;
-
+  protected static finalized: boolean|undefined;
   /**
    * Memoized list of all class properties, including any superclass properties.
    * Created lazily on user subclasses when finalizing the class.
@@ -315,8 +316,7 @@ export abstract class UpdatingElement extends HTMLElement {
    * @nocollapse
    */
   protected static finalize() {
-    if (this.hasOwnProperty(JSCompiler_renameProperty('finalized', this)) &&
-        this.finalized) {
+    if (finalized.has(this)) {
       return;
     }
     // finalize any superclasses
@@ -324,7 +324,7 @@ export abstract class UpdatingElement extends HTMLElement {
     if (typeof superCtor.finalize === 'function') {
       superCtor.finalize();
     }
-    this.finalized = true;
+    finalized.add(this);
     this._ensureClassProperties();
     // initialize Map populated in observedAttributes
     this._attributeToPropertyMap = new Map();
@@ -793,3 +793,9 @@ export abstract class UpdatingElement extends HTMLElement {
   protected firstUpdated(_changedProperties: PropertyValues) {
   }
 }
+
+/**
+ * Marks class as having finished creating properties.
+ */
+const finalized = new Set<typeof UpdatingElement>();
+finalized.add(UpdatingElement);
--- a/src/lib/updating-element.ts
+++ b/src/lib/updating-element.ts
@@ -193,8 +193,9 @@ type UpdateState = typeof STATE_HAS_UPDATED|typeof STATE_UPDATE_REQUESTED|
 
 /**
  * The Closure JS Compiler doesn't currently have good support for static
- * property semantics where "this" is dynamic, so we use this hack to bypass
- * any rewriting by the compiler.
+ * property semantics where "this" is dynamic (e.g.
+ * https://github.com/google/closure-compiler/issues/3177 and others) so we use
+ * this hack to bypass any rewriting by the compiler.
  */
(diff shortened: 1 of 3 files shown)


src/lit-element.ts
```diff
@@ -85,7 +79,11 @@ export class LitElement extends UpdatingElement {
 
   /** @nocollapse */
   protected static finalize() {
-    super.finalize();
+    if (this !== LitElement) {
```
rictic: Add a comment about this if statement.

Is it to match the previous `finalized = true` behavior?

src/lib/updating-element.ts
```diff
@@ -793,3 +793,9 @@ export abstract class UpdatingElement extends HTMLElement {
   protected firstUpdated(_changedProperties: PropertyValues) {
   }
 }
+
+/**
+ * Marks class as having finished creating properties.
+ */
+const finalized = new Set<typeof UpdatingElement>();
```
jridgewell: Wouldn't a `WeakSet` be better here? As is, this'll strongly hold onto every `LitElement`.

src/lib/updating-element.ts
```diff
@@ -211,10 +211,11 @@ export abstract class UpdatingElement extends HTMLElement {
   private static _attributeToPropertyMap: AttributeMap;
 
   /**
-   * Marks class as having finished creating properties.
+   * This property was previously used to record the classes that have already
+   * been finalized. This property is no longer used, but is still declared here
+   * for backwards compatability in TypeScript typings.
```
rictic: IMO breaking TypeScript typings isn't a breaking change for the library.

It would be better for someone to know that the finalized property no longer means anything. In any event, I doubt this will affect anyone, as the finalized property is pretty inside baseball.

src/lib/updating-element.ts
justinfagnani: This would only hold on to class definitions (`typeof UpdatingElement`). But also, @aomarks is changing the PR to use a property again, just with the index operator and string literal to avoid Closure's bug. That'll allow the GC to collect classes if they every become freeable.

jridgewell: Ah, I missed the `static` keyword in the code. 👍

src/lib/updating-element.ts
aomarks: Fair enough, removed.

src/lit-element.ts
aomarks: Yes, done.

aomarks:
PTAL @justinfagnani @rictic. We now use a string property instead of a Set, and (from discussion with @justinfagnani) I also moved the finalized check before recursing, instead of after, which optimizes out unnecessary super calls.

src/lit-element.ts
aomarks: (Update: ended up wanting the special name in lit-element.ts after all, since it helps optimize the finalization stack, but I just copied the string instead of exporting something somebody might try and depend on)

src/lib/updating-element.ts
```diff
@@ -191,6 +191,13 @@ type UpdateState = typeof STATE_HAS_UPDATED|typeof STATE_UPDATE_REQUESTED|
     typeof STATE_IS_REFLECTING_TO_ATTRIBUTE|
     typeof STATE_IS_REFLECTING_TO_PROPERTY|typeof STATE_HAS_CONNECTED;
 
+/**
+ * The Closure JS Compiler doesn't currently have good support for static
+ * propertiy semantics where "this" is dynamic, so we use this hack to bypass
+ * any rewriting by the compiler.
```
justinfagnani: Link to Closure bug

src/lit-element.ts
```diff
@@ -85,7 +79,9 @@ export class LitElement extends UpdatingElement {
 
   /** @nocollapse */
   protected static finalize() {
-    super.finalize();
+    // The Closure JS Compiler does not always preserve the correct "this"
+    // when calling static super methods, so explicitly bind here.
```
justinfagnani: Link to Closure bug

src/lib/updating-element.ts
```diff
@@ -191,6 +191,13 @@ type UpdateState = typeof STATE_HAS_UPDATED|typeof STATE_UPDATE_REQUESTED|
...
+ */
+const finalized = '__finalized__';
```
rictic: Hm, if we're going this approach, I think we could do something like the following:

```typescript
const finalized = 'finalized';

class UpdateElement {
  protected static [finalized] = false;
  protected static finalize() {
    this[finalized] = true;
    // ...
  }
}
```

src/lib/updating-element.ts
aomarks: There's no one bug for this right now.

aomarks: Added one of them.

src/lit-element.ts
aomarks: Done.

src/lit-element.ts
```diff
@@ -85,7 +79,9 @@ export class LitElement extends UpdatingElement {
...
+    super.finalize.apply(this);
```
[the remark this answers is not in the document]
aomarks: Good call

src/lib/updating-element.ts
aomarks: Done.

PR #745, opened 2019-07-22
Delegate updateComplete getter to getUpdateComplete method for TS ES5

aomarks:
When the target language is ES5, TypeScript does not support calling a superclass getter, which means it is not possible (or at least very awkward) to override the `updateComplete` getter in an element (because a call like `super.updateComplete.then(...)` will not compile). See https://github.com/microsoft/TypeScript/issues/338.

We now delegate the `updateComplete` getter to a new `_getUpdateComplete` method, which can be safely overriden by users. A little awkward, but seemed like the least bad solution we could think of.

See also https://github.com/material-components/material-components-web-components/pull/207 which tried to work around this issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 <!-- ### Added -->
-<!-- ### Changed -->
+### Changed
+* Elements should now override the new `getUpdateComplete` method instead of the `updateComplete` getter, for compatibility with TypeScript ES5 output, which does not support calling a superclass getter (e.g.`super.updateComplete.then(...)`) due to [TypeScript#338](https://github.com/microsoft/TypeScript/issues/338).
 <!-- ### Removed -->
 ### Fixed
 * Fixed compatibility with Closure JS Compiler optimizations relating to static properties ([#732](https://github.com/Polymer/lit-element/issues/732)).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ## Unreleased
 <!-- ### Added -->
 ### Changed
-* Elements should now override the new `getUpdateComplete` method instead of the `updateComplete` getter, for compatibility with TypeScript ES5 output, which does not support calling a superclass getter (e.g.`super.updateComplete.then(...)`) due to [TypeScript#338](https://github.com/microsoft/TypeScript/issues/338).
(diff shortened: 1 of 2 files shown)


src/lib/updating-element.ts
```diff
@@ -736,10 +736,34 @@ export abstract class UpdatingElement extends HTMLElement {
...
+   * This method should be overridden instead. For example:
+   *
+   *   class MyElement extends LitElement {
+   *     ...
+   *     getUpdateComplete() {
+   *       return super.getUpdateComplete().then(this.myWorkIsDone);
+   *     }
+   *   }
+   */
+  protected getUpdateComplete() {
```
justinfagnani: prefix with an underscore

CHANGELOG.md
```diff
@@ -12,7 +12,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 <!-- ### Added -->
-<!-- ### Changed -->
+### Changed
+* Elements should now override the new `getUpdateComplete` method instead of the `updateComplete` getter, for compatibility with TypeScript ES5 output, which does not support calling a superclass getter (e.g.`super.updateComplete.then(...)`) due to [TypeScript#338](https://github.com/microsoft/TypeScript/issues/338).
```
justinfagnani: `getUpdateComplete` -> `_getUpdateComplete`

src/lib/updating-element.ts
```diff
@@ -736,10 +736,34 @@ export abstract class UpdatingElement extends HTMLElement {
    * rendered element before fulfilling this Promise. To do this, first await
    * `super.updateComplete` then any subsequent state.
    *
+   * IMPORTANT: Do not override this getter directly. Override the
```
justinfagnani: This is only import for TypeScript users, yes? Babel does the right thing it seems. Unsure about Closure.

aomarks: Yes, but I was thinking it's preferable to have just one style recommendation that will always be safe.

CHANGELOG.md
aomarks: Done.

src/lib/updating-element.ts
aomarks: Done.

aomarks:
Filed https://github.com/Polymer/lit-element/issues/746 to document this cc @arthurevans

src/lib/updating-element.ts
justinfagnani: I guess. I just find it a little odd that this is really only because of a very specific set of circumstances triggering a bug in one compiler with one particular setting, and only when subclassers themselves are using TypeScript...

Oh, to be done with ES5.

Issue #696, opened 2019-05-31
CSS minification not possible


LarsDenBakker:
lit-html HTML inside template literals are easy to minify because lit-html requires the HTML to be parseable as valid HTML. For example using https://github.com/cfware/babel-plugin-template-html-minifier

With lit-element's CSS templates this is unfortunately not the case because we can concatenate partial css to eventually become valid css. I love this feature a lot, but unfortunately now it's no longer possible to use tools for minifying CSS because there are very few assumptions that can be made about the code.

I'm wondering if this is something that has already been considered, and if there are existing tools/solutions which handle css minification. Otherwise is it perhaps otherwise possible to differentiate partial and non-partial css so that tools can at least minify the non-partial css.

MarkiyanPyts:
Possible with this https://www.npmjs.com/package/lit-scss-loader

LarsDenBakker:
Thanks for your reaponse. In a CSS file you cant write JS expressions and create invalid CSS. So im not sure how this solves my problem.

MarkiyanPyts:
Webpack will minify your css and insert it in mini fried state into your web component, you can write normal sass in external file. Why would you want to write js expressions in css?

LarsDenBakker:
I dont want want to build and bundle during development, so something that relies on webpack is not an option.

Js expressions can be used to share constants without needing special tools.

MarkiyanPyts:
To that I don’t have answer maybe someone from core team can tell us that.

danielbarion:
#747

Guys, maybe this PR help this question.

CaptainCodeman:
The `rollup-plugin-minify-html-literals` plugin works great for me - minifies both html and css.

https://github.com/asyncLiz/rollup-plugin-minify-html-literals

LarsDenBakker:
@danielbarion not sure how documentation fixes the issue?

@CaptainCodeman does that handle partial/invalid css correctly like explained above?

danielbarion:
@LarsDenBakker if you configure your "compilation" proccess, you can transpile SASS, Stylus, etc... To css before import, so, when your minify runs, it will be only a normal css...

 I'm using stylus in my test project (a lib of web componentes written with LitElement), in the main app, when I import some component, I'm just importing js and css and the minify can do they work as expected.

CaptainCodeman:
> does that handle partial/invalid css correctly like explained above?

[...]

PR #747, opened 2019-07-23
Documentation improvement style

danielbarion:
### Reference Issue
issue #696
issue #561

### DIFF
--- a/docs/_guide/styles.md
+++ b/docs/_guide/styles.md
@@ -220,6 +220,27 @@ There are some important caveats though:
 * External styles can cause a flash-of-unstyled-content (FOUC) while they load.
 * The URL in the `href` attribute is relative to the _main document_, making this technique mostly useful for application elements where asset URLs are well known, and not for reusable elements published publicly.
 
+Now, be sure what are you doing before do this: by the way, if you `really need` or ` really want` a `separate css file from main component file`, you can use the `unsafeCSS` function to import the "external css" like the example below.
+
+```js
+import {LitElement} from 'lit-element';
+import style from './style.css';
+
+class MyElement extends LitElement {
+  static get styles() {
+    return [css`${unsafeCSS(style)}`];
+  }
+
+  render() {
+    return html`
+      <link rel="stylesheet" href="./styles.css">
+    `;
+  }
+}
+```
+
+Thus, it is possible to avoid flash-of-unstyled-content (FOUC) and have a good organization in the project structure.
+
 ### Write CSS styles for a host element and its shadow DOM {#shadow-dom-styles}
 
 In this section: 
--- a/docs/_guide/styles.md
+++ b/docs/_guide/styles.md
@@ -8,11 +8,11 @@ slug: styles
 * ToC
 {:toc}
 
-This page is a guide to the styling options available for LitElement components. 
+This page is a guide to the styling options available for LitElement components.
 
 * [Styling options for component developers](#styling-for-developers) describes how to apply styles internally to a component that you're building.
 
-* [Styling options for component consumers](#styling-for-consumers) describes how to style an element that you're importing and using. 
+* [Styling options for component consumers](#styling-for-consumers) describes how to style an element that you're importing and using.
 
 * [Theming](#theming) is relevant to both developers and consumers of a LitElement-based component. It describes how to create easily stylable LitElement components, and how to create a style theme that can easily be applied to imported LitElement components.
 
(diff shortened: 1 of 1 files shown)


danielbarion:
@arthurevans take a look when you can and  if I need modify anything, please let me know.

docs/_guide/styles.md
```diff
@@ -220,9 +220,30 @@ There are some important caveats though:
 * External styles can cause a flash-of-unstyled-content (FOUC) while they load.
 * The URL in the `href` attribute is relative to the _main document_, making this technique mostly useful for application elements where asset URLs are well known, and not for reusable elements published publicly.
 
+Now, be sure what are you doing before do this: by the way, if you `really need` or ` really want` a `separate css file from main component file`, you can use the `unsafeCSS` function to import the "external css" like the example below.
```
arthurevans: I'm going to defer to @justinfagnani and @sorvell on whether this is something that we want to document.

Note that if the CSS you're importing is under your control, you can have the style file export a `css` tagged string, instead of using `unsafeCSS`.

docs/_guide/styles.md
```diff
@@ -220,9 +220,30 @@ There are some important caveats though:
...
+
+```js
+import {LitElement} from 'lit-element';
```
arthurevans:
```suggestion
import {LitElement, css, html} from 'lit-element';
```

docs/_guide/styles.md
danielbarion: Okay, thanks!

danielbarion:
@arthurevans okay, my apologize, I forgot to import `css` and `js`.

Thank you for the review.

Okay, I'll wait @justinfagnani and @sorvell

danielbarion:
Just a little information, in the docs the file are just a `.css` but in my test project, I'm using stylus:

```
import { LitElement, html, css, unsafeCSS } from 'lit-element'
import style from './style.styl'

class MyComponenet extends LitElement {
    static get styles() {
        return [css`${unsafeCSS(style)}`]
    }
.....
```

danielbarion:
@sorvell @justinfagnani Hey guys, any news about this PR?

danielbarion:
@sorvell @justinfagnani any news ?

danielbarion:
@arthurevans a example of how use the information of this PR

https://github.com/danielbarion/instruct-contacts

danielbarion:
@justinfagnani please, can you take a look here when you can?

docs/_guide/styles.md
justinfagnani: Why are "really need", "really want", and "separate css file from main component file" formatted as code?

docs/_guide/styles.md
```diff
@@ -220,9 +220,30 @@ There are some important caveats though:
 * External styles can cause a flash-of-unstyled-content (FOUC) while they load.
 * The URL in the `href` attribute is relative to the _main document_, making this technique mostly useful for application elements where asset URLs are well known, and not for reusable elements published publicly.
 
+Now, be sure what are you doing before do this: by the way, if you `really need` or ` really want` a `separate css file from main component file`, you can use the `unsafeCSS` function to import the "external css" like the example below.
+
+```js
+import {LitElement, css, html} from 'lit-element';
+import style from './style.css';
```
justinfagnani: This will return a CSSStyleSheet, so passing it to `unsafeCSS` will not work.

docs/_guide/styles.md
```diff
@@ -220,9 +220,30 @@ There are some important caveats though:
...
+
+class MyElement extends LitElement {
+  static get styles() {
+    return [css`${unsafeCSS(style)}`];
+  }
+
+  render() {
+    return html`
+      <link rel="stylesheet" href="./styles.css">
```
justinfagnani: Why load the styles twice?

justinfagnani:
@danielbarion I took a look, but I don't see how this applies to #696 or #561. This change doesn't mention minification and it appears to use the external styles that don't work in the polyfill.

docs/_guide/styles.md
danielbarion: I'm using this with .stylus, but work with sass too, I can commit this example with a .sass, what you say ?

docs/_guide/styles.md
danielbarion: I wanted to put a little emphasis on these words

docs/_guide/styles.md
danielbarion: My fault, sorry!
I'll fix it.

danielbarion:
This is not in docs and maybe some peoples want to do this like me.

I prefer a separated file for styles and I like to use stylus, if you think this is not necessary to be in docs, no problem.

I've build a simple app with this information, [app module](https://github.com/danielbarion/instruct-contacts/blob/master/src/modules/app/index.js).

I can make some changes to fix what you think need be modified.

Just let me know, thanks

docs/_guide/styles.md
justinfagnani: This doesn't work at all in the platform yet, and when it does work it'll return a CSSStyleSheet object, see http://w3c.github.io/webcomponents/proposals/css-modules-v1-explainer

I would rather not document importing CSS at all like this until it's available in browsers.

justinfagnani:
Should it be enough to document `unsafeCSS` and let people get the CSS source there however they can, with whatever tools they use? My problem with the way these docs are written is that they appear to assume a specific tool that allowing importing CSS, but the docs aren't in a tools section, nor do they mention the tool needed to make it work.

danielbarion:
Yeah, you're right. When I finish the build, all css, js and html (generated by lit-element) are just one file, so, the css is not a external file in the production or development mode.
I've used the point of "if the `unsafeCSS` import any css code, I just need parse the css in the URL resolve flow before everything.

The `unsafeCSS` give-me the possibility to import a style from another file, but I need parse the Stylus and / or SASS before give him to `unsafeCSS`.

I don't know what I need to do now, I just want to share this possibility with another devs, how can I do this @justinfagnani ? Can you help me please ? Do you have any suggestion ?

arthurevans:
I think from what you and @justinfagnani said, it seems like this is a good topic to address in the upcoming tooling section. I'll add that into my outline.

danielbarion:
That's very great @arthurevans, thank you!

If you guys have any suggestion, I can make some changes; or if I can help with anything, please let me know.

justinfagnani:
I think we can close this in favor of the docs on lit.dev

Issue #746, opened 2019-07-22
Document how to override updateComplete using _getUpdateComplete


aomarks:
I think we should update https://lit-element.polymer-project.org/guide/lifecycle#updatecomplete to explain how users can override `updateComplete` after https://github.com/Polymer/lit-element/pull/745 gets in. Due to a TypeScript ES5 limitation (see PR for more details), you cannot simply override `get updateComplete`, and we've added `_getUpdateComplete` as a `protected` override point for this purpose.

justinfagnani:
Note that this only applies to subclassing LitElement, in TypeScript, and overriding `updateComplete`, while also targeting ES5 with the TypeScript compiler. Subclassing works fine in plain JS, targeting ES2015+, or compiling to ES5 with Babel.

PR #750, opened 2019-07-23
Document _getUpdateComplete in LitElement docs

aomarks:
Fixes https://github.com/Polymer/lit-element/issues/746

### DIFF
--- a/docs/_guide/lifecycle.md
+++ b/docs/_guide/lifecycle.md
@@ -279,13 +279,15 @@ The `updateComplete` Promise resolves when the element has finished updating. Us
   this.updateComplete.then(() => { /* do stuff */ });
   ```
 
-To await additional state before fulfilling the `updateComplete` promise, override the `_getUpdateComplete` method. For example, it may be useful to await the update of a child element here. First await `super._getUpdateComplete()`, then any subsequent state. Note that the `updateComplete` getter should not be *directly* overridden due to a [limitation](https://github.com/microsoft/TypeScript/issues/338) in TypeScript's ES5 output support.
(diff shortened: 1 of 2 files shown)


docs/_guide/lifecycle.md
```diff
@@ -279,6 +279,19 @@ The `updateComplete` Promise resolves when the element has finished updating. Us
   this.updateComplete.then(() => { /* do stuff */ });
   ```
 
+To await additional state before fulfilling the `updateComplete` promise, override the `_getUpdateComplete` method. For example, it may be useful to await the update of a child element here. First await `super._getUpdateComplete()`, then any subsequent state.
```
arthurevans: If overriding `_getUpdateComplete` is less common than using `this.updateComplete`, I think this note should be _after_ the main example.

I'd suggest moving this after the `updateComplete` example, and introducing it with a L4 heading:

`#### Overriding updateComplete`

(to be clear, I'm suggesting inserting it just before `## Examples`, ~L302 in this PR.

docs/_guide/lifecycle.md
```diff
@@ -279,6 +279,19 @@ The `updateComplete` Promise resolves when the element has finished updating. Us
...
+
+Note that it's recommended to override the `_getUpdateComplete` method instead of the `updateComplete` getter to ensure compatibility with users who are using TypeScript's ES5 output (see [TypeScript#338](https://github.com/microsoft/TypeScript/issues/338)).
```
arthurevans: You can cut "Note that" here.

aomarks: Done.

docs/_guide/lifecycle.md
aomarks: SGTM, done!

PR #754, opened 2019-07-25
Add deployment instructions

arthurevans:
@aomarks Adding deployment instructions per our conversation. Would you take a look?

... I included a CLI instruction for migrating traffic, although maybe it would be simpler (if a little slower) to simply re-run the deploy command with `-- --promote`...

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -79,6 +79,8 @@ npm run deploy -- --version=<VERSION>
 
 This command stages the docs at `https://<VERSION>-dot-polymer-lit-element.appspot.com`.
 
+By specifying the version on the command line,  you can update an existing staged version.
+
 To make a staged version live, migrate traffic in the App Engine console or from the 
 command line:
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -54,3 +54,34 @@ API docs unless you're making changes to the doc comments in the lit-element sou
 To regenerate the API docs and rebuild the  site, run the following command:
 
 `npm run build-all`
+
+### Deployment
+
+To build and deploy a new version of the site:
+
+```bash
+npm run deploy
+```
+
+(You must have permission to update the App Engine project to deploy the site.)
+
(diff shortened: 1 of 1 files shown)


docs/README.md
```diff
@@ -54,3 +54,34 @@ API docs unless you're making changes to the doc comments in the lit-element sou
...
+
+```bash
+npm run deploy
+```
+
+(You must have permission to update the App Engine project to deploy the site.)
+
+The site is staged at a new URL (for example, `https://20181207t151022-dot-polymer-lit-element.appspot.com`). When ready, use the App Engine console to migrate traffic to the new version.
+
+To push to a named App Engine version, specify the version on the command line:
```
aomarks: Dumb question, but why would we ever want to push to a specific version? Wouldn't we either stage at the temporary URL shown above, or push to production?

arthurevans: Good question. I do it, for example, when I'm working on a significant new chunk of doc that I expect will take several revisions, and I want people to be able to review it at a stable URL, like:

`https://new-tooling-section-dot-polymer-lit-element.appspot.com`

I considered mentioning that in the PR--I'll go ahead and add that explanation.

arthurevans:
```suggestion
Sometimes it's useful to stage to a specific, named version. For example, when 
iterating on a new doc or design change, you may want to stage to an easily-to-remember
version name, like `new-tooling-section`. 

To push to a named App Engine version, specify the version on the command line:
```

docs/README.md
```diff
@@ -54,3 +54,34 @@ API docs unless you're making changes to the doc comments in the lit-element sou
...
+
+```bash
+npm run deploy -- --version=<VERSION>
+```
+
+This command stages the docs at `https://<VERSION>-dot-polymer-lit-element.appspot.com`.
+
```
arthurevans:
```suggestion

By specifying the version on the command line,  you can update an existing staged version.

```

PR #769, opened 2019-08-08
Minor doc updates regarding decorators.

arthurevans:


arthurevans:
Ping @justinfagnani minor doc update re: decorators, trying to remove the "typescript only" implication.

docs/_guide/properties.md
```diff
@@ -59,7 +59,7 @@ static get properties() {
 ```
 
 ```js
-// TypeScript decorators
+// Decorators (requires transpiling)
```
justinfagnani: What about "requires TypeScript or Babel"?

docs/_guide/properties.md
```diff
@@ -104,15 +104,18 @@ Remember to call `super()` first in your constructor, or your element won't rend
...
-You can also declare properties with TypeScript decorators:
+You can also declare properties with decorators:
 
 ```js
 @property({type : String})  prop1 = 'Hello World';
 ```
 
-**Example: Declare properties with TypeScript decorators** 
+Decorators are a proposed JavaScript feature, so you'll need to use a transpiler like Babel or the TypeScript compiler to use decorators. 
+With Babel, you'll need to use the `@babel/plugin-proposal-decorators` plugin. Decorators are supported by default in TypeScript.
```
justinfagnani: Decorators aren't supported by default in TypeScript. You need `"experimentalDecorators": true` in .tsconfig. You _don't_ need `emitDecoratorMetadata` (and should ideally have that off as it bloats the output).

arthurevans: TIL. Thanks!

arthurevans:
Think I fixed the issues you found, @justinfagnani .

docs/_includes/projects/properties/declaretypescript/index.ts
```diff
@@ -1 +1 @@
-import './my-element.js';
+import './my-element.ts';
```
justinfagnani: The import needs to stay `.js`

arthurevans: That doesn't appear to work in StackBlitz. That sample on the site currently fails in StackBlitz, while ones that import `my-element.ts` work: https://stackblitz.com/edit/z5wgme

justinfagnani: Oh, that's bad. In local TypeScript development the file extensions are not rewritten, so you have to use `.js`. @ericsimons any idea why StackBlitz is different?

EricSimons: Hm perhaps you need to set the preset as `typescript`? It might be set as just `javascript`, which doesn't include tsc/etc

arthurevans: Where does the preset get set? I can't find any setting by that name.

arthurevans: We generate all of our stackblitzes programmatically using `template: 'typescript'` in the project data. IIRC, we need this to get stuff to compile.

EricSimons: Oh sorry I think I misunderstood what's going on here- tbh kinda confused. Are you trying to import a .ts file that exists on disk, but instead using the .js extension? If so, that seems it would break on local too 🤔

justinfagnani: No, that's the way `tsc` works. `tsc` does not transform import specifiers at all, and it does recognize that `./foo.js` maps to `./foo.ts`, so you import `./foo.js` and after compilation everything works. If you import `./foo.ts` things don't work because that's not a valid file after compilation.

EricSimons: Ah gotchya, so this is vanilla tsc without webpack/rollup/etc? Our TS preset is based on webpack's TS loader, so that would explain it. Is it possible to just remove the file extension entirely?

justinfagnani: Yeah, but that doesn't work locally either, at least without a node module specifier resolution step.

EricSimons: Dang. This is a tough one bc all of our presets are based around webpack (minus static/polymer).

Any flexibility on transforms/config? Maybe something like this? https://github.com/Zoltu/typescript-transformer-append-js-extension

I imagine that might help the code make a bit more sense to the user as well; I'd prob be confused why I'm importing a .js file when its origin is actually .ts

arthurevans: I think we already have a bit of an uncanny valley between what works locally and in StackBlitz. In particular, because we use the typescript preset, all of our projects (JS or TS) have an index.ts file that exports the main .js or .ts file.

Users who use tsc directly are used to importing a `.js` file instead of the `.ts` source file. But it sounds like webpack users may be used to importing a bare specifier, so there's going to be some confusion any way we do it.

PR #781, opened 2019-08-20
Update doc references for ?checked

jordanaustin:
This removes the `?checked` reference in the docs in favor of `?disabled`, since this is a cleaner representation of a boolean attribute since there are no user interactions that can set the property.


jordanaustin:
I would love to have additional notes about why a dev should use `.checked` rather than `?checked` but I'm unsure where that could go. 

Would love to have a discussion about this :)

justinfagnani:
Thanks @jordanaustin!

> I would love to have additional notes about why a dev should use .checked rather than ?checked but I'm unsure where that could go.
>
> Would love to have a discussion about this :)

We can have a discussion about this in an issue with @arthurevans

PR #789, opened 2019-08-27
Add decorator examples

abraham:
Fixes https://github.com/Polymer/lit-element/issues/803

`customElement` wasn't rendering `@example` for some reason so it's just part of the description.

### DIFF
--- a/src/lib/decorators.ts
+++ b/src/lib/decorators.ts
@@ -74,7 +74,16 @@ const standardCustomElement =
 /**
  * Class decorator factory that defines the decorated class as a custom element.
  *
- * @param tagName the name of the custom element to define
+ * ```
+ * @customElement('my-element')
+ * class MyElement {
+ *   render() {
+ *     return html``;
+ *   }
+ * }
+ * ```
+ *
+ * @param tagName The name of the custom element to define.
  */
 export const customElement = (tagName: string) =>
     (classOrDescriptor: Constructor<HTMLElement>|ClassDescriptor) =>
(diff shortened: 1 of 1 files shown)


abraham:
@justinfagnani ping.

abdonrd:
I have a question about the `@query()` decorator.
What is the good practice to initialize?
Because TS complains if it is not initialized.

Screenshot 2019-11-15 at 15 22 56

> Property 'form' has no initializer and is not definitely assigned in the constructor.

abraham:
Depending on how often you are referencing the property and how confident the element will always be available you could make it `private form!: HTMLFormElement;`. It's somewhat frowned on but this is one of the usecases. Alternaltivly you could use `?` instead but if you use that property a lot you'll always have to gate it's usage with value checks.

src/lib/decorators.ts
```diff
@@ -173,6 +207,24 @@ export function query(selector: string) {
...
  * that executes a querySelectorAll on the element's renderRoot.
  *
+ * @param selector A DOMString containing one or more selectors to match.
+ *
+ * See: https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll
+ *
+ * @example
+ *
+ *     class MyElement {
+ *       @query('div')
```
justinfagnani: Needs to be `queryAll`

src/lib/decorators.ts
```diff
@@ -150,6 +166,24 @@ export function property(options?: PropertyDeclaration) {
...
  * executes a querySelector on the element's renderRoot.
  *
+ * @param selector A DOMString containing one or more selectors to match.
+ *
+ * See: https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector
+ *
+ * @example
+ *
+ *     class MyElement {
+ *       @query('.first')
```
justinfagnani: Maybe use an id here?

PR #801, opened 2019-09-05
Update Constructor type to fix #217

### LINKED ISSUE #217: Proposal: change the signature of the Constructor<T> type for use in mixins
### Issue
Please change the signature of the `Constructor<T>` type (declared in [decorators.ts](https://github.com/Polymer/lit-element/blob/master/src/lib/decorators.ts#L20)) to use `any` instead of `unknown` so it can be used to create mixins. Without this change, anyone using lit-element will have to declare his/her own `Constructor<T>` type to create mixins.
``` ts
export type Constructor<T> = {
  new (...args: unknown[]): T
};

// should be changed to

export type Constructor<T> = {
  new (...args: any[]): T
};

// or

export type Constructor<T> = new (...args: any[]) => T;
```

### Versions
- lit-element: v0.6.1

justinfagnani:
Fixes #217

### DIFF
--- a/src/lib/decorators.ts
+++ b/src/lib/decorators.ts
@@ -25,7 +25,7 @@ import {LitElement} from '../lit-element.js';
 import {PropertyDeclaration, UpdatingElement} from './updating-element.js';
 
 export type Constructor<T> = {
-  new (...args: unknown[]): T
+  new (...args: any[]): T
 };
 
 // From the TC39 Decorators proposal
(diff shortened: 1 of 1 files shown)


src/lib/decorators.ts
```diff
@@ -25,7 +25,7 @@ import {LitElement} from '../lit-element.js';
 import {PropertyDeclaration, UpdatingElement} from './updating-element.js';
 
 export type Constructor<T> = {
-  new (...args: unknown[]): T
+  new (...args: any[]): T
```
rictic: Add a lint ignore comment, and if this isn't a lint warning, we should make it one

Issue #673, opened 2019-05-09
Webpack Dev Server HMR --> Duplicate Element Issue 


T-Knott-Mesh:
Hello all, so I have recently moved over to a webpack build pipeline for purposes of building, transpiling, and using this guy [https://github.com/GoogleChromeLabs/worker-plugin](WorkerPlugin). All that being said, Hot Module Replacement does not work due to the fact that when it replaces the module, the browser sees this as an element re-registry... 

The folks over at Skate JS baked in a solution [Same issue with Solution](https://github.com/skatejs/skatejs/pull/664)  [Exact Issue](https://github.com/skatejs/skatejs/issues/564)  
[Final Change](https://github.com/skatejs/skatejs/pull/664/commits/9830bc867c011988c9b3d1bed9d250369e55ec51)
they have gotten it to work. 

I am looking to see if there is any way to build in a similar solution into the @customElement decorator.

Would love to know thoughts. This could affect a lot of people.

hyperpress:
Might also be a good question for the folks over at @polymer slack open-wc
channel.

Nachasic:
I am experiencing the same issue with [fuse-box bundler](https://fuse-box.org) HMR. I'm currently looking into writing custom HMR driver to resolve this.

justinfagnani:
Hot module loading is not, in general, a really workable system. All kinds of state will be confused by reloading modules, including module variables, event listeners, and the custom elements registry.

I don't think generating a new element name is useful, as other logic and CSS may depend on that name. More likely is that we can find a way to patch `CustomElementRegistry.define` so that we can redefine an element and re-render all instances.

vegarringdal:
We created custom elements hmr polyfill here:
- https://github.com/vegarringdal/custom-elements-hmr-polyfill
Readme contains a very simple sample on codesandbox

This adds a proxy around CustomElementRegistry.define.
Yesterday we also found a workaround to observedAttributes by using mutationObserver. (not updated npm yet)

If you have suggestion of callbacks that would be useful to have then create issue.
Currently I use it with the fusebox bundler v4-(next).

justinfagnani:
@rictic could you take a look at custom-elements-hmr-polyfill?

rictic:
[...]

PR #802, opened 2019-09-06
Initial hot elements implementation. [Not ready to merge]

rictic:
Two things I'd want to fix before merging:
  1. This code should never show up in production (if for no other reason than it adds unnecessary bytes). What's the best way of ensuring that? Currently this uses `goog.DEBUG` for that purpose, which works well with closure compiler, but not anything else.
  2. Need to use proxies when available, so that elements stay constructable after they've been hot reloaded.

Related to #673

### DIFF
--- a/src/lib/css-tag.ts
+++ b/src/lib/css-tag.ts
@@ -47,6 +47,32 @@ export class CSSResult {
   toString(): string {
     return this.cssText;
   }
+
+  /**
+   * Used when generating files at build time that transform .css files
+   * into something like CSS Modules that export CSSResults.
+   *
+   * Does nothing in production, and doesn't replace existing references
+   * unless the current browser supports adopting stylesheets (at time of
+   * writing that Chrome only).
+   *
+   * We could support replacing styles cross-browser,
+   * but it would be very tricky to do so without leaking memory, becasue we'd
+   * need to keep track of every style element that `this` is written into.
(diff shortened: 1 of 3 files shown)


rictic:
Ideally hot-elements would live in its own package though, and optimizing out the unused hot reloading code could be done at application build time.

If we had such a mechanism we could use it in other places too, e.g. to log errors when unexpected things occur that would be too expensive to check for in production

justinfagnani:
We're really not going to have such a mechanism though, at least any time soon. There's no standard for stripping dead code and we've seen in other packages that dev code can easily end up in prod. My only thought at the moment is that we could vend an API compatible `lit-html-dev` package and allow build tools to swap it in.

In the mean time, patching seems best to me.

abdonrd:
@rictic @justinfagnani there are any plans to move this to `lit-next`?

LarsDenBakker:
Thanks to @43081j we now have a HMR plugin for web dev server: https://modern-web.dev/docs/dev-server/plugins/hmr/

I created a small lit-html demo which applies some of the code of this PR to a simple base class using lit-html, just to check if this can work out: https://github.com/modernweb-dev/web/tree/master/packages/dev-server-hmr/demo/lit-html

It would be great if we could still introduce something like this before `lit-next` if possible.

justinfagnani:
@rictic we should port this to 3.0

PR #810, opened 2019-09-16
A couple of small size optimizations

justinfagnani:
* Use _updateComplete Promise to wait for first connection
* Optimize _enqueueUpdate for size a bit
* Inline _markUpdated
* Inline and combine style flattening and de-duplication

Before:
```
┌─────────────────────────────────────────┐
│                                         │
│   Destination: lit-element.bundled.js   │
│   Bundle Size:  21.04 KB                │
│   Minified and Gzipped Size:  6.75 KB   │
│   Brothli size: 6.13 KB                 │
│                                         │
└─────────────────────────────────────────┘
```

After:
```
┌─────────────────────────────────────────┐
│                                         │
│   Destination: lit-element.bundled.js   │
│   Bundle Size:  20.7 KB                 │
│   Minified and Gzipped Size:  6.65 KB   │
│   Brothli size: 6.04 KB                 │
│                                         │
└─────────────────────────────────────────┘
```

### DIFF
--- a/src/lit-element.ts
+++ b/src/lit-element.ts
@@ -36,30 +36,7 @@ declare global {
 
 export interface CSSResultArray extends Array<CSSResult|CSSResultArray> {}
 
-/**
- * Minimal implementation of Array.prototype.flat
- * @param arr the array to flatten
- * @param result the accumlated result
- */
-function arrayFlat(
-    styles: CSSResultArray, result: CSSResult[] = []): CSSResult[] {
-  for (let i = 0, length = styles.length; i < length; i++) {
-    const value = styles[i];
-    if (Array.isArray(value)) {
-      arrayFlat(value, result);
-    } else {
-      result.push(value);
-    }
-  }
-  return result;
-}
-
-/** Deeply flattens styles array. Uses native flat if available. */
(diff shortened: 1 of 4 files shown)


src/lit-element.ts
```diff
@@ -107,26 +84,29 @@ export class LitElement extends UpdatingElement {
...
+                  set);
+      // Array.from does not work on Set in IE, otherwise return
+      // Array.from(addStyles(userStyles, new Set<CSSResult>())).reverse()
+      const set = addStyles(userStyles, new Set<CSSResult>());
+      const styles: CSSResult[] = [];
+      set.forEach((v) => styles.unshift(v));
+      return styles;
     }
-    return styles;
+    return [userStyles];
```
sorvell: This could produce `[undefined]` if no styles are set. Let's guard against that.

justinfagnani: That's guarded by this call in `finalize()`:

```ts
    this._styles =
        this.hasOwnProperty(JSCompiler_renameProperty('styles', this)) ?
        this._getUniqueStyles() :
        this._styles || [];
```

sorvell: This is guarded against above so it's unneeded here.

PR #816, opened 2019-09-24
Fix broken link to ES dev server

amio:
The Open Web Components __developing server__ is now __ES dev server__
https://open-wc.org/developing/es-dev-server.html


amio:
@googlebot I signed it!

daKmoR:
@amio also thx from us at open-wc 🤗

PR #826, opened 2019-10-04
Styles update

arthurevans:
Take 3 on this section.

Staged: https://styles-update-dot-polymer-lit-element.appspot.com/guide/styles


jbingham:
first reaction: big improvement over the original page!
will look more closely and provide more detailed comments

docs/_guide/templates.md
```diff
@@ -8,21 +8,35 @@ slug: templates
...
+
 class MyElement extends LitElement {
   render() {
     return html`<p>template content</p>`;
   }
 }
 ```
 
-* Write your template in HTML inside a JavaScript template literal by enclosing the raw HTML in back-ticks (<code>``</code>). 
+*   Write your template in HTML inside a JavaScript template literal by enclosing the raw HTML in back-ticks
```
jbingham: [template literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)

docs/_guide/templates.md
```diff
@@ -8,21 +8,35 @@ slug: templates
 * ToC
 {:toc}
 
+Add a template to your component to define internal DOM to implement your component. 
+
+LitElement uses
```
jbingham: How about flipping the order, to focus on the user benefit, rather than the implementation? Ie: 

To encapsulate the templated DOM and provide scoped styling, LitElement uses [shadow DOM](...). That way, you can add styles ...

Maybe we can even point out here, and in other places, why this is better and different than other approaches. Basically, it's a web standard way. No verbose BEM or extra build step required to munge your css class names to scope them.

docs/_guide/templates.md
```diff
@@ -210,7 +232,7 @@ _my-element.js_
 
 Since the introduction of shadow DOM, we use the term "light DOM" to refer to nodes that appear in the main DOM tree.
```
jbingham: It's a little odd that this sentence appears before the reference to light DOM in the text in the next sentence. Merge?

docs/_guide/styles.md
```diff
@@ -162,327 +122,618 @@ class MyElement extends LitElement {
...
+{% include project.html folder="style/superstyles" openFile="my-element.js" %}
 
-</div>
+## Shadow DOM styling overview {#shadow-dom}
+
+This section gives a brief overview of shadow DOM styling.
+
+Styles you add to a component can affect:
+
+* [The shadow tree (your component's rendered template)](#shadowroot).
```
jbingham: I'd flip this, as "The component's rendered template in the shadow tree."

docs/_guide/styles.md
```diff
@@ -162,327 +122,618 @@ class MyElement extends LitElement {
...
+* [The component itself](#host).
+* [The component's children](#slot).
 
-```js
-import {LitElement, property} from 'lit-element';
 
+### Style the shadow tree {#shadowroot}
```
jbingham: "Style the rendered template" ? Or "Style the rendered template in the shadow tree". 

Shadow tree = implementation detail, renderered template = user concept.

docs/_guide/styles.md
```diff
@@ -162,327 +122,618 @@ class MyElement extends LitElement {
...
   }
 }
 ```
 
-### Define styles in an external stylesheet {#external-stylesheet}
+{% include project.html folder="style/styleatemplate" openFile="my-element.js" %}
+
+### Style the component itself
+
+Styles defined in an element's shadow tree can style the element itself (also known as the _shadow host_).
```
jbingham: is it clearer to say "style the custom element", or "style the LitElement component" rather than just "the element"?

docs/_guide/styles.md
```diff
@@ -162,327 +122,618 @@ class MyElement extends LitElement {
...
+Note that the host element can be affected by styles from outside the shadow tree, as well, so you should consider 
+the styles you set in `:host` and `:host()` rules as _default styles_ that can be overridden by the user. For example:
+
+```css
+my-element {
+  display: inline-block;
+}
+```
+
+### Style the component's children
```
jbingham: This section is missing any newbie context. It's pretty close to pure jargon. 

"If your component will have children (like the built-in <select> element has <option>s and <ul> has <li>s), your template must contain placeholders for where the children should be placed inside the rendered template. These are called slots, and are specified with the <slot> element:"

docs/_guide/styles.md
```diff
@@ -162,327 +122,618 @@ class MyElement extends LitElement {
...
-  Inline elements are sized by their content and cannot have their dimensions set via CSS. They can also cause block element children to overflow its bounds.
+{:.alert .alert-info}
+<div>
 
-  ```css
-    :host {
-      display: block;
-    }
-  ```
+**Watch out for Shady CSS limitations around slotted content!** See the [Shady CSS limitations](https://github.com/webcomponents/polyfills/tree/master/packages/shadycss#limitations) for details on how to use the `::slotted()` syntax in a polyfill-friendly way.
```
jbingham: "limitations in the Shady CSS polyfill"

docs/_guide/templates.md
arthurevans: It is a weird sentence. If we're going to use the term "light DOM," we should define it. (Otherwise, maybe we should just say, "children"?)

jbingham: If there's only one use of the term "light DOM" in the doc, might be worth rewording to avoid introducing the terminology. Or maybe define it after the usage?

docs/_guide/styles.md
arthurevans: Rendered template isn't any less any implementation detail, to my mind. lit-html templates aren't part of the platform, and users don't add templates for their own sake. Templates are a way of generating DOM.

This section is specifically about shadow DOM styling, and I'm trying to balance avoiding jargon where possible with not inventing new terminology. "Styling the rendered template" introduces a new term that's specific to LitElement. Shadow DOM is one they can look up many other places.

docs/_guide/styles.md
arthurevans: Hmm... Good thought. It's referring back to "Styles defined in an element's shadow tree..." so would have to rewrite the first half.

Hrm... "Styles defined in a LitElement component's shadow tree" is wordy as heck.

Styles defined in the component's shadow tree can style the component itself.

You can style the component itself with styles defined in the component's shadow tree.

I wanted to introduce the term "host" because otherwise the name of the selector is mysterious.

How about we cut to the chase and say:

You can style the component itself using special _host_ selectors. (The element that owns, or "hosts" a shadow tree is called the _host element_.)

docs/_guide/styles.md
arthurevans: I think this is part and parcel of the discussion below, so not going to comment separately.

docs/_guide/styles.md
arthurevans: Yep, we can do better here. I'll take a stab.

docs/_guide/styles.md
```diff
@@ -162,327 +122,618 @@ class MyElement extends LitElement {
...
 
-### Define styles in an external stylesheet {#external-stylesheet}
+{% include project.html folder="style/styleatemplate" openFile="my-element.js" %}
+
+### Style the component itself
+
+Styles defined in an element's shadow tree can style the element itself (also known as the _shadow host_).
 
-We strongly recommend static styles, CSS custom properties, or [lit-html's `classMap` or `styleMap` directives](TODO) if you're styling non-host shadow root contents.
+To create default styles for the host element, use the `:host` CSS pseudo-class and `::host()` CSS pseudo-class function.
```
arthurevans:
```suggestion
To create default styles for the host element, use the `:host` CSS pseudo-class and `:host()` CSS pseudo-class function.
```

docs/_guide/styles.md
```diff
@@ -162,327 +122,618 @@ class MyElement extends LitElement {
...
+    :host { 
+      display: block; 
+    }
+
+    /* Selects the host element if it is hidden */
+    :host([hidden]) { 
+      display: none; 
+    }
+
+    /* Selects the host element if it has class "blue" */
```
arthurevans: Probably don't need the third example here.

docs/_guide/styles.md
arthurevans:
```suggestion
**Watch out for limitations in the Shady CSS polyfill around slotted content!** See the [Shady CSS limitations](https://github.com/webcomponents/polyfills/tree/master/packages/shadycss#limitations) for details on how to use the `::slotted()` syntax in a polyfill-friendly way. 
```

docs/_guide/templates.md
arthurevans:
```suggestion
*   Write your template in HTML inside a JavaScript [template literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) by enclosing the raw HTML in back-ticks 
```

docs/_guide/styles.md
arthurevans: Actually, there is a separate section on `<slot>` in the templates chapter. It seems like that would be a better place for more detailed background information. I might rephrase this sentence a little bit, but will point to the other section.

docs/_guide/templates.md
arthurevans: Removing the light DOM references.

docs/_guide/templates.md
arthurevans: I've re-written this to flip the order, and also expand a little bit: focusing on just one benefit (style encapsulation) confuses the issue.

Need to decide whether this is getting a bit too long and needs to be moved elsewhere.

arthurevans:
Another batch of updates.

Staging updated at: https://styles-update-dot-polymer-lit-element.appspot.com/guide/styles

docs/_guide/styles.md
```diff
@@ -8,141 +8,101 @@ slug: styles
...
+For more information, see [Resources](templates#resources) in the Templates chapter for links to some shadow DOM primers.
 
-### Define styles in a static styles property {#static-styles}
+## Add styles to your component {#add-styles}
 
-LitElement lets you define static styles that apply to all instances of a component.
+For optimal performance, define scoped styles in a static `styles` property. 
 
-We recommend using static styles for optimal performance. LitElement uses [Constructable Stylesheets](https://wicg.github.io/construct-stylesheets/) in browsers that support it, with a fallback for browsers that don't. Constructable Stylesheets allow the browser to parse styles exactly once and reuse the resulting Stylesheet object for maximum efficiency.
+Define styles in a tagged template literal, using the the `css` tag function:
```
jbingham: s/the the/the/

docs/_guide/styles.md
```diff
@@ -8,141 +8,101 @@ slug: styles
...
-              `
-            ];
-          }
-        }
-        ```
+{:.alert .alert-info}
+<div>
 
-#### Expressions in static styles
+**Consider using CSS variables and custom properties to create [configurable styles](#configurable).** CSS cusom properties work well with app themes, and also allow you to create per-instance styles.
```
jbingham: s/cusom/custom/

docs/_includes/projects/style/perinstanceexpressions/my-element.js
```diff
@@ -0,0 +1,35 @@
+import { LitElement, html } from 'lit-element';
+
+const perClassStyle = html`
```
aomarks: WDYT about adding a comment here like:

`// Note per-class styles should almost always be defined using the static styles array instead of with a <style> tag.`

So that people who jumped straight here don't think this is recommended.

docs/_guide/styles.md
```diff
@@ -162,327 +122,653 @@ class MyElement extends LitElement {
...
+
+*  The [ShadyCSS polyfill](https://github.com/webcomponents/shadycss/blob/master/README.md#limitations) doesn't support external style sheets.
+
+*   External styles can cause a flash-of-unstyled-content (FOUC) while they load.
+
+*   The URL in the `href` attribute is relative to the **main document**. This is okay if you're building an app and your asset URLs are well-known, but avoid using external style sheets when building a reusable element.
+
+## Dynamic classes and styles
+
+One way to make styles dynamic is to add bindings to the `class` or `style` attributes in your template.
```
aomarks: Suggest: ... to the `class` or `style` attributes *of elements* in your template

(kind of dumb, but it confused me for a second, because I initially interpreted this as an attribute on the `<template>`)

docs/_guide/styles.md
```diff
@@ -162,327 +122,653 @@ class MyElement extends LitElement {
...
+
+The lit-html library offers two directives, `classMap` and `styleMap`, to conveniently apply classes and styles in HTML templates. 
+
+For more information on these and other lit-html directives, see the documentation on [lit-html built-in directives](https://lit-html.polymer-project.org/guide/template-reference#built-in-directives).
+
+To use `styleMap` and/or `classMap`:
```
aomarks: This feels a little repetitive, because there are dedicated sections about `styleMap` and `classMap` right below, with similar examples. How about moving the import statements directly into the examples below, and getting rid of this introductory example?

docs/_guide/styles.md
```diff
@@ -162,327 +122,653 @@ class MyElement extends LitElement {
...
+<div style="
+  background-color:blue;
+  font-family:Roboto;
+  --custom-color:#e26dd2;
+  --otherCustomColor:#77e26d;">
+</div>
 ```
 
-{% include project.html folder="style/inherited2" openFile="my-element.js" %}
+Equivalent CSS syntax:
```
aomarks: Seems a little repetitive too. I think you could maybe drop this "Examples" section. There are already a number of examples explaining the same thing above, plus text explaining it, plus a table, so this all feels a little over-explained to me.

Really the only interesting thing to me about `styleMap` is that `dash-case` gets turned into `camelCase` -- quoting property names is just how JavaScript objects works.

I think I basically just want to see *one* example of using `styleMap` that shows all the possible forms:

```
<button style=${styleMap({
  border: '1px solid black',

  'background-color': 'blue',
  backgroundColor: 'blue', // Equivalent to above

  '--custom-color': 'red',
  '--customColor': 'red', // Equivalent to above
})}>A button</button>

File: docs/_guide/styles.md [hunk 17]
```diff
@@ -162,327 +122,653 @@ class MyElement extends LitElement {
...
+```
 
-Custom properties inherit down the DOM tree, and through shadow root boundaries. You can use this to let your users apply styles and themes to your elements.
+{% include project.html folder="style/specificity" openFile="index.html" %}
 
-For example, the following element sets its background color to a CSS variable that uses the value of the custom property `--myBackground` if it is available, and uses `yellow` otherwise:
+</div>
 
-_my-element.js_
+Any properties that a host element inherits will also be inherited by the elements in its template:
```
aomarks: Suggest removing this line and example, it seems like it's already been explained above?

docs/_guide/templates.md
```diff
@@ -8,21 +8,44 @@ slug: templates
 * ToC
 {:toc}
 
+Add a template to your component to define internal DOM to implement your component. 
+
+To encapsulate the templated DOM LitElement uses
+[shadow DOM](https://developers.google.com/web/fundamentals/web-components/shadowdom) . Shadow DOM provides three basic benefits:
```
aomarks: extra space before period

aomarks: maybe remove "basic"?

docs/_guide/templates.md
```diff
@@ -497,12 +527,46 @@ render() { return html`<slot name="thing"></slot>`; }
...
 ```
 
-## Further reading
-Since LitElement uses lit-html's `html` and `render` functions to render templates you can take advantage
-of the entire lit-html feature-set for writing your templates. You can find further information
-* [on the lit-html homepage](https://lit-html.polymer-project.org)
-* [in the Template Reference](https://lit-html.polymer-project.org/guide/template-reference)
+## Using other lit-html features
+
+Since LitElement uses lit-html's `html` and function to render templates you can take advantage
```
aomarks: s/and function/function

docs/_guide/templates.md
```diff
@@ -497,12 +527,46 @@ render() { return html`<slot name="thing"></slot>`; }
...
+installing two different versions of lit-html:
+
+```bash
+npm i lit-element@^2.0.0
+npm i lit-html@^1.0.0
+```
+
+### Import and use a lit-html directive
+
+You can import and use a lit-html directive and use it as shown in the lit-html documentation.
```
aomarks: suggest hyperlink for "lit-html documentation"

docs/_guide/styles.md
arthurevans: 👍

docs/_guide/styles.md
arthurevans: 🌮

docs/_guide/styles.md
```diff
@@ -8,141 +8,101 @@ slug: styles
 * ToC
 {:toc}
 
-This page is a guide to the styling options available for LitElement components. 
+Add styles to your element using a static `styles` property:
```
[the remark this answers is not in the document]
arthurevans: I'm torn on this. It is repetitive, but when we didn't have this up front, people felt there should be a simple example up front instead of just text.

@jbingham WDYT? I think you were a proponent of having the example up front.

docs/_guide/styles.md
```diff
@@ -8,141 +8,101 @@ slug: styles
...
 
-LitElement lets you define static styles that apply to all instances of a component.
+For optimal performance, define scoped styles in a static `styles` property. 
 
-We recommend using static styles for optimal performance. LitElement uses [Constructable Stylesheets](https://wicg.github.io/construct-stylesheets/) in browsers that support it, with a fallback for browsers that don't. Constructable Stylesheets allow the browser to parse styles exactly once and reuse the resulting Stylesheet object for maximum efficiency.
+Define styles in a tagged template literal, using the the `css` tag function:  
 
-The styles in the static `styles` property are evaluated once only and applied to all instances of the element. You can modify styles per element instance by using [CSS custom properties](#cssprops):
+```js
+// Import the `css` helper function
```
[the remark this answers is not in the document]
arthurevans: 👍

docs/_guide/styles.md
```diff
@@ -8,141 +8,101 @@ slug: styles
...
+<div>
 
-#### Expressions in static styles
+**Consider using CSS variables and custom properties to create [configurable styles](#configurable).** CSS cusom properties work well with app themes, and also allow you to create per-instance styles.
 
-Static styles apply to all instances of an element. Any expressions in your CSS are evaluated and included **once**, then reused for all instances. 
+</div>
 
-To prevent LitElement-based components from evaluating potentially malicious code, the `css` tag only accepts literal strings. You can nest them like this:
+To prevent LitElement-based components from evaluating potentially malicious code, the `css` tag only accepts literal strings. With expressions, you can nest `css` literals:
```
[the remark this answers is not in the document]
arthurevans: 👍

docs/_guide/styles.md
jbingham: Hmm, I was thinking of a template as the generic concept, in that sentence. There's like 500 different ways to template and render HTML, server-side and client-side. That's a much more widely understood concept than shadow DOM. A "lit-html template" is a specific term, but the concept of a template is about 20 years old.

docs/_guide/styles.md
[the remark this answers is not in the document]
jbingham: Merging the two sections seems good, as a way to avoid repetition. The caveat is that it would be really good to have as few words as possible before the example, since most people won't read below the fold. Eg, "This page is a public guide to the various ways to style LitElement components. TL;DR: In most cases, you should add styles to your element using the static `styles` property like this:"

docs/_guide/styles.md
arthurevans: 👍

docs/_guide/templates.md
arthurevans: good catch. I think that was supposed to be "tag function."

docs/_guide/styles.md
```diff
@@ -8,141 +8,101 @@ slug: styles
...
-}
+{% include projects/style/nestedcss/my-element.js %}
 ```
 
-However, if you want to inject any variable or non-literal into a css string, you must wrap it with the `unsafeCSS` function. For example:
+{% include project.html folder="style/nestedcss" openFile="my-element.js" %}
 
-```js
-import { LitElement, css, unsafeCSS } from 'lit-element';
+However, if you want to use expressions to add a non-literal to a `css` literal, you must wrap the non-literal with the `unsafeCSS` function. For example:
```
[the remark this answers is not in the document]
arthurevans: I kind of did both: I used your suggested text above, but left in an abbreviated version of the alert box to emphasize the point:

**Only use the `unsafeCSS` tag with trusted input.** Injecting unsanitized CSS is a security risk. For example, malicious CSS can "phone home" by adding an image URL that points to a third-party server.

docs/_guide/styles.md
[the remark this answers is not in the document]
arthurevans: 👍

docs/_guide/styles.md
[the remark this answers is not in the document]
arthurevans: Removed the first example.

arthurevans:
I think I've addressed all the feedback, and restaged at https://styles-update-dot-polymer-lit-element.appspot.com/guide/styles

I'm pretty happy with the current state, and I'd like to merge & deploy at this point if there are no objections.

Issue #603, opened 2019-03-07
Static build for CDN distribution


slightlyoff:
Per a F2F conversation with @justinfagnani, wanted to capture agreement to add a single, modern-browser-exclusive build of LitElement and the most frequently used lit-html directives to the distributed NPM module. The goal is to enable easy inclusion of LitElement from a single source file.

## Rationale

LitElement, when built, is tiny! That isn't apparent when using it from, e.g., unpkg. The need for the Polymer CLI tools also increases the friction to getting started with the edit/compile cycle that historically makes web development so much fun.

Instead of trying to boil the whole ocean by providing every build variant, or having to compromise on size to accomodate legacy, the idea here is to get developers started quickly while providing an easier path to "real" LitElement usage via local builds.

Part of this approach might need to add guards to the build to give up when loaded in the presence of existing LitElement (to prevent over-inclusion) and console logging to flag that this isn't the most efficient approach.

justinfagnani:
The idea we discussed was to build a fairly reasonable set of bundles for LitElement and lit-html, and publish them to npm under `dist/` folders. The bundles would be minified ES2017 modules. LitElement would include UpdateingElement, but not lit-html - it would reference the lit-html bundle. The lit-html bundle would include everything reachable from `lit-html.js`, but not directives. They would be minified separately.

jrobinson01:
thoughts on using https://github.com/pikapkg/web instead of a CDN?

justinfagnani:
To clarify, this isn't for local installations via the npm registry, and we won't be changing the `main` and `module` fields of `package.json`. For most users nothing will change at all.

This is simply a way for us to distribute fairly well optimized bundles to CDNs that host npm files so that people can use them easily. We're a bit unusual in that we only publish plain ES2017 modules. Most packages right now still publish minified bundles in a variety of module formats. Our current approach is more optimal for local development, but gives a bad impression for people directly loading files from a npm CDN.

`dist/` might be a bad folder name - 'prebuilt/` may be better. We'll put a README in the folder saying not to use these builds for production (unless you have a npm CDN that's intended for production use) and probably add a console warning too.

slightlyoff:
Is this in progress now?

e111077:
Regarding pikapkg, the CDN launched today which serves a minified bundle at `https://cdn.pika.dev/lit-element`

justinfagnani:
Ugh... that looks like it bundles lit-html, so if you import lit-html separately to use directives, it won't work.

[...]

PR #840, opened 2019-10-23
amd

### LINKED ISSUE #140: Provide bundled JavaScript files.
### Description

The browser does not support the bare import.
It is impossible to use lit-element without toolchain (webpack or rollup) now
as lit-element use bare import.
https://github.com/Polymer/lit-element/blob/2433ce376521aa8ab5272aa9c12a49ee74ac8c4e/src/lit-element.ts#L14-L23
or the browser will throw the error:

```err
TypeError: Module specifier does not start with "/", "./", or "../".
```
---
It is better to provide different versions of bundled files. Such as UMD, CommonJS and ES Module.
There is an [example](https://vuejs.org/v2/guide/installation.html#Explanation-of-Different-Builds) in Vue project.

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Safari 11
- [x] Safari 10
- [x] IE 11

### Versions

- lit-element: v0.5.2
- webcomponents: v2.0.2

taoqf:
### Reference Issue

Fixes #140

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,8 @@
 #Compiler Output
 lit-element.js
 lit-element.js.map
+lit-element-amd.js
+lit-element-amd.js.map
 lit-element.d.ts
 lit-element.d.ts.map
 
--- a/package.json
+++ b/package.json
@@ -1,71 +1,75 @@
 {
-  "name": "lit-element",
-  "version": "2.2.1",
-  "description": "A simple base class for creating fast, lightweight web components",
-  "license": "BSD-3-Clause",
-  "homepage": "https://lit-element.polymer-project.org/",
-  "repository": "Polymer/lit-element",
-  "main": "lit-element.js",
-  "module": "lit-element.js",
-  "directories": {
-    "test": "test"
-  },
-  "files": [
-    "/lib/",
-    "/src/",
-    "!/src/demo/",
-    "!/src/test/",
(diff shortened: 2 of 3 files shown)


taoqf:
https://github.com/Polymer/lit-html/pulls

taoqf:
https://github.com/Polymer/lit-html/pull/1019

taoqf:
@googlebot I could not signed it as a chinese developer!

taoqf:
@justinfagnani Would you pls take a look at this?

justinfagnani:
Hi @taoqf and thanks for the PR. To be clear, we're not interested in foxing #140 as phrased. #603  is about providing a pre-built file basically for demo consumption only.

We'd still want to distribute modules, since all current browsers support them and do not support AMD without a loader. And I really would want the files to be named appropriately scary so that developers don't use them in production, like `lit-element/prebuilt/lit-element-not-for-production.js`, and include a `console.warn()` to not use them either. The goal of #603 is also to indicate a more realistic file-size, so we'd want to run the bundle through a minifier too.

I think Rollup would be the best tool for this, rather than an alternate tsconfig file.

taoqf:
<https://www.npmjs.com/package/@taoqf/lit-element>
<https://www.npmjs.com/package/@taoqf/lit-html>
will do for now.

PR #849, opened 2019-11-05
Prevent creating ShimAdoptedStyleSheets if not instance of TemplateResult

JaySunSyn:
Pointing to this issue https://stackoverflow.com/questions/58519848/styles-are-not-applied-to-the-litelement-in-safari-firefox

If `templateResult` was not a TemplateResult instance, the ShimAdoptedStyleSheets were still created.

If `templateResult` becomes an instance of TemplateResult afterwards, ShimAdoptedStyleSheets are not created anymore because _needsShimAdoptedStyleSheets prevents creating them more than once.

### DIFF
--- a/src/lit-element.ts
+++ b/src/lit-element.ts
@@ -196,15 +196,15 @@ export class LitElement extends UpdatingElement {
     super.update(changedProperties);
     const templateResult = this.render() as unknown;
 
-    if (templateResult instanceof TemplateResult) {
+    if (templateResult instanceof TemplateResult === false) {
       return;
     }
     
     (this.constructor as typeof LitElement)
-          .render(
-              templateResult,
-              this.renderRoot,
-              {scopeName: this.localName, eventContext: this});
+      .render(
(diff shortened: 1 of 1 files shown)


src/lit-element.ts
```diff
@@ -195,13 +195,17 @@ export class LitElement extends UpdatingElement {
   protected update(changedProperties: PropertyValues) {
     super.update(changedProperties);
     const templateResult = this.render() as unknown;
+
     if (templateResult instanceof TemplateResult) {
-      (this.constructor as typeof LitElement)
+      return;
```
justinfagnani: Did you mean to negate the instanceof check above?

JaySunSyn: Yes!
Wow, that 2nd commit didn't get pushed.

JaySunSyn:
I pushed the fixes.

justinfagnani:
This is fixed in Lit 2, which takes a pretty different approach to appending static styles.

PR #850, opened 2019-11-06
Add a "bundle" build target

slightlyoff:
Adds a supported way to invoke rollup to generate `lit-element.bundled.js` so I don't have to dig it out of `./node_modules/` any more.


justinfagnani:
We're going to publish bundles on lit.dev for Lit 2: https://github.com/lit/lit/issues/1257

PR #853, opened 2019-11-17
Native cssstylesheet

### LINKED ISSUE #774: Support CSSStyleSheet for .styles property
lit-element should support returning a `CSSStyleSheet` object from the static class getter `.styles()`, in order to support the upcoming CSS Modules spec.

This behavior should be supported:

```js
import styles from './my-element.css';

  static get styles() { return [styles]; }
}
```

Currently, I can do this, but it's using browser support only:

```js
import styles from './my-element.css';

  constructor() {
    super();

    // Lit doesn't clobber this, since it has zero styles
    this.shadowRoot.adoptedStyleSheets = [styles];
  }
}
```

samthor:
Fixes #774. Supports native `CSSStyleSheet` if and only if native adopted style sheets on shadow roots are supported.

### DIFF
--- a/src/lib/css-tag.ts
+++ b/src/lib/css-tag.ts
@@ -13,7 +13,7 @@ found at http://polymer.github.io/PATENTS.txt
  * Whether the current browser supports `adoptedStyleSheets`.
  */
 export const supportsAdoptingStyleSheets = (window.ShadowRoot) &&
-    (window.ShadyCSS === undefined) &&
+    (window.ShadyCSS === undefined || window.ShadyCSS.nativeShadow) &&
     ('adoptedStyleSheets' in Document.prototype) &&
     ('replace' in CSSStyleSheet.prototype);
 
--- a/src/lib/css-tag.ts
+++ b/src/lib/css-tag.ts
@@ -13,7 +13,7 @@ found at http://polymer.github.io/PATENTS.txt
  * Whether the current browser supports `adoptedStyleSheets`.
  */
 export const supportsAdoptingStyleSheets = (window.ShadowRoot) &&
-    ('adoptedStyleSheets' in ShadowRoot.prototype) &&
+    ('adoptedStyleSheets' in Document.prototype) &&
     ('replace' in CSSStyleSheet.prototype);
 
 const constructionToken = Symbol();
--- a/src/lib/css-tag.ts
+++ b/src/lib/css-tag.ts
@@ -12,8 +12,7 @@ found at http://polymer.github.io/PATENTS.txt
 /**
  * Whether the current browser supports `adoptedStyleSheets`.
  */
-export const supportsAdoptingStyleSheets =
-    (window.ShadowRoot) &&
+export const supportsAdoptingStyleSheets = (window.ShadowRoot) &&
     ('adoptedStyleSheets' in ShadowRoot.prototype) &&
     ('replace' in CSSStyleSheet.prototype);
 
--- a/src/lib/css-tag.ts
+++ b/src/lib/css-tag.ts
@@ -12,7 +12,7 @@ found at http://polymer.github.io/PATENTS.txt
 /**
  * Whether the current browser supports `adoptedStyleSheets`.
  */
-export const supportsAdoptingShadowStyleSheets =
+export const supportsAdoptingStyleSheets =
     (window.ShadowRoot) &&
     ('adoptedStyleSheets' in ShadowRoot.prototype) &&
     ('replace' in CSSStyleSheet.prototype);
@@ -37,9 +37,9 @@ export class CSSResult {
   // stylesheets are not created until the first element instance is made.
   get styleSheet(): CSSStyleSheet|null {
     if (this._styleSheet === undefined) {
-      // Note, if `supportsAdoptingShadowStyleSheets` is true then we assume
+      // Note, if `supportsAdoptingStyleSheets` is true then we assume
       // CSSStyleSheet is constructable.
-      if (supportsAdoptingShadowStyleSheets) {
+      if (supportsAdoptingStyleSheets) {
         this._styleSheet = new CSSStyleSheet();
         this._styleSheet.replaceSync(this.cssText);
       } else {
--- a/src/lib/css-tag.ts
+++ b/src/lib/css-tag.ts
@@ -13,6 +13,7 @@ found at http://polymer.github.io/PATENTS.txt
  * Whether the current browser supports `adoptedStyleSheets`.
  */
 export const supportsAdoptingStyleSheets = (window.ShadowRoot) &&
+    (window.ShadyCSS === undefined) &&
     ('adoptedStyleSheets' in Document.prototype) &&
     ('replace' in CSSStyleSheet.prototype);
 
--- a/src/lib/css-tag.ts
+++ b/src/lib/css-tag.ts
@@ -12,8 +12,9 @@ found at http://polymer.github.io/PATENTS.txt
 /**
  * Whether the current browser supports `adoptedStyleSheets`.
  */
-export const supportsAdoptingStyleSheets =
-    ('adoptedStyleSheets' in Document.prototype) &&
+export const supportsAdoptingShadowStyleSheets =
+    (window.ShadowRoot) &&
+    ('adoptedStyleSheets' in ShadowRoot.prototype) &&
     ('replace' in CSSStyleSheet.prototype);
 
 const constructionToken = Symbol();
@@ -28,6 +29,7 @@ export class CSSResult {
       throw new Error(
           'CSSResult is not constructable. Use `unsafeCSS` or `css` instead.');
     }
+
     this.cssText = cssText;
   }
 
@@ -35,9 +37,9 @@ export class CSSResult {
   // stylesheets are not created until the first element instance is made.
   get styleSheet(): CSSStyleSheet|null {
     if (this._styleSheet === undefined) {
-      // Note, if `adoptedStyleSheets` is supported then we assume CSSStyleSheet
-      // is constructable.
-      if (supportsAdoptingStyleSheets) {
+      // Note, if `supportsAdoptingShadowStyleSheets` is supported then we
+      // assume CSSStyleSheet is constructable.
+      if (supportsAdoptingShadowStyleSheets) {
         this._styleSheet = new CSSStyleSheet();
         this._styleSheet.replaceSync(this.cssText);
       } else {
--- a/src/lib/css-tag.ts
+++ b/src/lib/css-tag.ts
@@ -37,8 +37,8 @@ export class CSSResult {
   // stylesheets are not created until the first element instance is made.
   get styleSheet(): CSSStyleSheet|null {
     if (this._styleSheet === undefined) {
-      // Note, if `supportsAdoptingShadowStyleSheets` is supported then we
-      // assume CSSStyleSheet is constructable.
+      // Note, if `supportsAdoptingShadowStyleSheets` is true then we assume
+      // CSSStyleSheet is constructable.
       if (supportsAdoptingShadowStyleSheets) {
         this._styleSheet = new CSSStyleSheet();
         this._styleSheet.replaceSync(this.cssText);
--- a/src/lit-element.ts
+++ b/src/lit-element.ts
@@ -18,7 +18,7 @@ import {PropertyValues, UpdatingElement} from './lib/updating-element.js';
 export * from './lib/updating-element.js';
 export * from './lib/decorators.js';
 export {html, svg, TemplateResult, SVGTemplateResult} from 'lit-html/lit-html.js';
-import {supportsAdoptingStyleSheets, CSSResult} from './lib/css-tag.js';
+import {supportsAdoptingShadowStyleSheets, CSSResult} from './lib/css-tag.js';
 export * from './lib/css-tag.js';
 
 declare global {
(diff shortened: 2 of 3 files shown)


src/lib/css-tag.ts
```diff
@@ -49,6 +49,18 @@ export class CSSResult {
   }
 }
 
+export class CSSStyleSheetResult extends CSSResult {
```
justinfagnani: Why is this wrapper necessary?

justinfagnani:
Ping @samthor

src/lib/css-tag.ts
samthor: I don't think it is. I removed it.

samthor:
Sorry, I was on paternity leave and then I forgot this when I got back.

PTAL. I've changed the code a bit so we don't subclass `CSSResult` when we're provided with native `CSSStyleSheet` instances.

samthor:
Oops, I didn't notice that change in my rebase. I've restored that line and modified a test to ensure we _also_ call `.getStyles()`.

src/lib/css-tag.ts
```diff
@@ -9,8 +9,9 @@ part of the polymer project is also subject to an additional IP rights grant
 found at http://polymer.github.io/PATENTS.txt
 */
 
-export const supportsAdoptingStyleSheets =
-    ('adoptedStyleSheets' in Document.prototype) &&
+export const supportsAdoptingShadowStyleSheets =
```
justinfagnani: Why the name change? This is kind of public API, so I wouldn't...

src/lit-element.ts
```diff
@@ -58,59 +60,77 @@ export class LitElement extends UpdatingElement {
...
   /**
    * Array of styles to apply to the element. The styles should be defined
-   * using the `css` tag function.
+   * using the `css` tag function or via constructible stylesheets.
    */
-  static styles?: CSSResult|CSSResultArray;
+  static styles?: CSSResultOrNative|CSSResultArray;
 
   private static _styles: CSSResult[]|undefined;
+  private static _nativeStyles: CSSStyleSheet[]|undefined;
```
justinfagnani: Why does this need to be separate from `_styles`? Doesn't total ordering between the two sets matter too?

src/lit-element.ts
```diff
@@ -152,30 +172,33 @@ export class LitElement extends UpdatingElement {
   }
 
   /**
-   * Applies styling to the element shadowRoot using the `static get styles`
-   * property. Styling will apply using `shadowRoot.adoptedStyleSheets` where
+   * Applies styling to the element shadowRoot using the private styles
```
justinfagnani: I don't think we should mention a private field in a protected API doc

src/lit-element.ts
```diff
@@ -58,59 +60,77 @@ export class LitElement extends UpdatingElement {
...
-      const styles: CSSResult[] = [];
-      set.forEach((v) => styles.unshift(v));
-      this._styles = styles;
+      const set = addStyles(userStyles, new Set<CSSResultOrNative>());
+      set.forEach((v) => work.unshift(v));
+    } else if (userStyles !== undefined) {
+      work.push(userStyles);
+    }
+
+    if (supportsAdoptingShadowStyleSheets) {
```
justinfagnani: Can we do one loop with this check inside? I think it'll be more compact.

src/lib/css-tag.ts
samthor: Sure. I've left the new semantics though, since all we care about is support on a `ShadowRoot`. There's [a polyfill for IE11](https://www.npmjs.com/package/construct-style-sheets-polyfill) which only supports `Document`; in that case, the old behavior was wrong, and would cause a crash.

src/lit-element.ts
samthor: We use one or the other depending on browser support. I also though about using a union TypeScript type but it got particularly awkward.

src/lit-element.ts
samthor: We're creating two different arrays. I could change the `_styles` and `_nativeStyles` to be a type union but I think it'll be more confusing when we apply them in `adoptStyles`.

src/lit-element.ts
samthor: I've reverted this comment. I think it's better the way it was.

My confusion came from... I think the tense is a bit ambiguous. By the time this function is called, we've already created the private "mapping" of styles (via `_getUniqueStyles()`), and this function doesn't call the `static get styles` method itself. Not a big deal, because we've _already_ called it, but that's why I was eager to modify the comment.

samthor:
ping @justinfagnani 

I've rebased the PR again.

samthor:
ugh, sorry for noise—I merged badly. I've just made the changes against master again.

src/lit-element.ts
chase-moskal: well the concern about ordering/precedence still matters, right?

so to merge `_nativeStyles` into `_styles`, you'd want the type signature something like this:

```ts
private static _styles: (CSSResult | CSSStyleSheet)[] | undefined
```

then, when you're looping over each stylesheet, you'd distinguish whether it's a native stylesheet using the `instanceof` operator or something along those lines

samthor: Adding `_nativeStyles` lets us assign at once to each element—we just set `.adoptedStyleSheets = this._nativeStyles`. I think it's much more ergonomic than casting every element of the array when you assign it.

I want to hear what @justinfagnani thinks—mostly I hope I've made it clear that we use _either_ array, so total ordering doesn't matter.

chase-moskal: > I think it's much more ergonomic than casting every element of the array when you assign it.

i'm not sure i follow.. casting every element when assigning to the array? maybe there's just a bit of confusion about the details of that type signature

the type signature `(CSSResult | CSSStyleSheet)[] | undefined` would allow the array to contain stylesheets of *either* type: so you'd be able to push both types of stylesheets to the array without any type casting

```ts
const cssResult: CSSResult = getExampleCssResult()
const nativeCss: CSSStyleSheet= getExampleNativeCss()

this._styles.push(cssResult) // no casting
this._styles.push(nativeCss) // no problem

for (const style of this._styles) {
  // style has type: (CSSResult | CSSStyleSheet)
}
```

however when you access one of the array values, you'd likely want to distinguish it's type to handle them separately

samthor: The type signature should actually be `(CSSResult[] | CSSStyleSheet[])`, because you only have `CSSStyleSheet` where native adopted stylesheets are supported (regardless of input).

And then the only way you check the type is by doing something like `this._styles[0] instanceof CSSStyleSheet`, which is awkward.

justinfagnani:
@samthor I tweaked this change to only use one array for both CSSSyleSheets and CSSResults. It's only a small conditional in `adoptStyles`. I think the rest is simpler. See the diff against master here: https://github.com/Polymer/lit-element/compare/native-stylesheets#diff-2ab97d0f347f8a3887d614f36018898bR187

samthor:
@justinfagnani I pulled your change in and fixed some comments. PTAL

src/lit-element.ts
```diff
@@ -104,35 +107,29 @@ export class LitElement extends UpdatingElement {
     }
     // Take care not to call `this.getStyles()` multiple times since this
     // generates new CSSResults each time.
-    // TODO(sorvell): Since we do not cache CSSResults by input, any
```
justinfagnani, line 107: I think this comment needs to stay. It's referring to there not being a cache from TemplateStringsArray to CSSResult, so we end up creating a new CSSResult and therefore new CSSStyleSheet each time we evaluate the same `css` tag. This can happen when people use `static get styles()` rather than `static styles =`, and the accessor is called multiple times.

src/lit-element.ts
```diff
@@ -174,7 +171,7 @@ export class LitElement extends UpdatingElement {
   }
 
   /**
-   * Applies styling to the element shadowRoot using the [[`styles`]]
+   * Applies styling to the element shadowRoot using the `static get styles`
```
justinfagnani, line 174: I'd remove the "get" part here. We read the `static styles` property. In can incidentally by implemented with a geter.

src/lib/css-tag.ts
```diff
@@ -12,8 +12,8 @@ found at http://polymer.github.io/PATENTS.txt
 /**
  * Whether the current browser supports `adoptedStyleSheets`.
  */
-export const supportsAdoptingStyleSheets =
-    ('adoptedStyleSheets' in Document.prototype) &&
+export const supportsAdoptingStyleSheets = (window.ShadowRoot) &&
+    ('adoptedStyleSheets' in ShadowRoot.prototype) &&
```
justinfagnani, line 16: We need to keep this checks against Document, since we have developers that delete the properties on Document.prototype to disable the use of adopted style sheets so they can edit in DevTools.

src/lit-element.ts
arthurevans: Also, please leave the brackets (`[[]]`) in place around `styles`--this creates a link in our Typedoc API docs.

src/lib/css-tag.ts
samthor: Sure. I've reverted this but kept the check for `ShadowRoot`.

src/lit-element.ts
samthor: Fixed both. I think removing the brackets was a bad merge on my part.

samthor:
I've addressed feedback and pushed again. I presume the PR will pass Travis :)

justinfagnani:
To be clear, you want to test the case where the platform has native constructible stylesheet support, but the polyfill is also forced on. Otherwise there's no way yet to get a constructible stylesheet.

samthor:
I think what y'all mean is, you're worried about the case where there is native constructible stylesheets support but the WC polyfills are forced on. This can only happen in Chromium right now, since that's the only place native constructible stylesheets are supported. Does that sound right?

If the target doesn't have constructible stylesheets _at all_ (or that is polyfilled—although that space is a bit fuzzy, we did it for Santa Tracker), I don't think there's an issue there.

src/lit-element.ts
```diff
@@ -109,31 +112,54 @@ export class LitElement extends UpdatingElement {
...
+      if (s instanceof CSSStyleSheet) {
+        if (s.ownerNode) {
+          throw new Error(`CSSStyleSheet instances used in 'styles' cannot ` +
+              `come from <style> tags.`);
+        }
+        if (!supportsAdoptingStyleSheets) {
+          // Flatten the cssText from the passed constructible stylesheet. The
+          // user might have expected to update their stylesheets over time,
+          // but the alternative was a crash.
+          const cssText = Array.prototype.slice.call(s.cssRules)
```
sorvell, line 155:
```suggestion
          const cssText = Array.prototype.slice.call(s.cssRules).reduce((css, rule) => css + rule.cssText, '');
```

src/lib/css-tag.ts
```diff
@@ -12,7 +12,8 @@ found at http://polymer.github.io/PATENTS.txt
 /**
  * Whether the current browser supports `adoptedStyleSheets`.
  */
-export const supportsAdoptingStyleSheets =
+export const supportsAdoptingStyleSheets = (window.ShadowRoot) &&
```
sorvell, line 15: Think we need to change this to `('adoptedStyleSheets' in window.ShadowRoot.prototype)`, instead of `('adoptedStyleSheets' in Document.prototype)`. This makes sure it'll be false when ShadyDOM is enabled.

sorvell: Seems like this change should be needed to get the test running under the polyfill to run?

samthor:
@sorvell @justinfagnani We're good for another review now.

* we flatten `CSSStyleSheet` rules where native constructible stylesheets are supported _but_ the WC polyfill is enabled
* we throw when a user tries to pass a `CSSStyleSheet` from `<style>` tag (rather than trying to flatten)

And I've updated the tests a bit. We now explicitly check that, in the fully natively supported case, _changing_ the rules of a `CSSStyleSheet` will change the rendered style.

src/lib/css-tag.ts
samthor: @justinfagnani had an issue with this. To quote him:

> We need to keep this checks against Document, since we have developers that delete the properties on Document.prototype to disable the use of adopted style sheets so they can edit in DevTools.

src/lib/css-tag.ts
```diff
@@ -12,7 +12,8 @@ found at http://polymer.github.io/PATENTS.txt
...
+    (window.ShadyCSS === undefined) &&
```
sorvell, line 16: This can be defined but not in use... as in our test environment.

```suggestion
    (window.ShadyCSS === undefined || window.ShadyCSS.nativeShadow) &&
```

src/test/lit-element_styling_test.ts
```diff
@@ -890,6 +897,89 @@ suite('Static get styles', () => {
...
+        assert.equal(
+            getComputedStyle(div).getPropertyValue('border-top-width').trim(),
+            expectedValue);
+      });
+
+  // Test that when ShadyCSS is enabled with native support for
+  // adoptedStyleSheets, we can return a CSSStyleSheet that will be flattened
+  // and play nice with others.
+  const testShadyCSSWithAdoptedStyleSheetSupport = (typeof ShadowRoot === 'function') &&
+      ('adoptedStyleSheets' in window.ShadowRoot.prototype) &&
```
sorvell, line 945: I don't think this test ever runs. When using the Shadow DOM polyfill, `ShadowRoot` does not have an `adoptedStyleSheets` property. Just taking out that check should make this run.

samthor: Well, that makes the test fail, because I can't construct `CSSStyleSheet` without `adoptedStyleSheets` support.

I've changed the logic a bit:
```js
  const testShadyCSSWithAdoptedStyleSheetSupport =
      (window.ShadowRoot) &&
      ('replace' in CSSStyleSheet.prototype) &&
      (window.ShadyCSS !== undefined && !window.ShadyCSS.nativeShadow);
```

This uses `ShadyCSS` the same way we do in css-tag.ts, and basically says whether we're flattening the stylesheets because the polyfill is included.

samthor:
Thanks @sorvell for pointing out the test wasn't running. Fixed now.

samthor:
Thanks for the thorough review. I look forward to the next release so I can pin Santa against it.

PR #860, opened 2019-12-04
Adds minor new API for enabling updating and querying assigned nodes.

sorvell:
Adds a protected method called `enableUpdating` to `UpdatingElement`. Before this, updating is enabled when the element is connected in a way that is not easily customizable. Adding this API enables subclasses to control when updates are allowed to begin.

Adds a `queryAssignedNodes(slotName: string, flatten: boolean)` decorator. This provides a convenient way to retrieve the nodes distributed to a given slot via a property and is a shorthand for `this.shadowRoot.querySelectorAll(`slot[name=${slotName}]`).assignedNodes(flatten)`.


Westbrook:
Will this return the same nodes/be updated in the same way regardless of whether using the polyfill or not?

src/lib/decorators.ts
```diff
@@ -265,3 +265,28 @@ export function eventOptions(options: AddEventListenerOptions) {
              // tslint:disable-next-line:no-any decorator
              any;
 }
+
+/**
+ * A property decorator that converts a class property into a getter that
+ * returns the `assignedNodes` of the given named `slot`.
+ *
+ * @ExportDecoratedItems
```
rictic: This doesn't need to be `@ExportDecoratedItems` any more, as of https://github.com/angular/tsickle/pull/1125

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 <!-- ### Added -->
 <!-- ### Removed -->
 
+### Added
+* Added `enableUpdating()` to `UpdatingElement` to enable customizing when updating is enabled.
```
justinfagnani: SInce it doesn't look like we have issues for these, can you link to this PR?

src/lib/decorators.ts
```diff
@@ -265,3 +265,28 @@ export function eventOptions(options: AddEventListenerOptions) {
...
+ */
+export function queryAssignedNodes(slotName: string = '', flatten: boolean = false) {
+  return (protoOrDescriptor: Object|ClassElement,
+          // tslint:disable-next-line:no-any decorator
+          name?: PropertyKey): any => {
+    const descriptor = {
+      get(this: LitElement) {
+        const selector = `slot${slotName ? `[name=${slotName}]` : ''}`;
```
justinfagnani: Doesn't this need a selector before the `slot` tag name selector to work with ShadyDOM?

src/lib/decorators.ts
```diff
@@ -265,3 +265,28 @@ export function eventOptions(options: AddEventListenerOptions) {
              // tslint:disable-next-line:no-any decorator
              any;
 }
+
+/**
+ * A property decorator that converts a class property into a getter that
+ * returns the `assignedNodes` of the given named `slot`.
```
justinfagnani: I think we should document the type that the property should be annotated with, ie `NodeListOf<HTMLElement>`

src/test/lib/decorators_test.ts
```diff
@@ -389,6 +391,66 @@ suite('decorators', () => {
...
+      render() {
+        return html`
+          <slot></slot>
+          <slot name="footer"></slot>
+        `;
+      }
+    }
+
+    @customElement(generateElementName() as keyof HTMLElementTagNameMap)
+    class C extends LitElement {
```
justinfagnani: Why do you need this element? To test slot forwarding?

src/test/lib/decorators_test.ts
```diff
@@ -389,6 +391,66 @@ suite('decorators', () => {
...
+    }
+
+    test('returns assignedNodes slot', async () => {
+      const c = new C();
+      container.appendChild(c);
+      await c.updateComplete;
+      await c.assignedNodesEl.updateComplete;
+      assert.deepEqual(c.assignedNodesEl.defaultAssigned, [c.div]);
+      const child = document.createElement('div');
+      c.assignedNodesEl.appendChild(child);
```
justinfagnani: Can you add some comments? I'm unsure what's being tested here. If you're assigning elements directly into the `assignedNodesEl`, then what is the slot for in C?

src/lib/updating-element.ts
```diff
@@ -495,14 +494,15 @@ export abstract class UpdatingElement extends HTMLElement {
...
-    // `_hasConnectionResolver` will exist. If so, resolve it to complete the
-    // update, otherwise requestUpdate.
-    if (this._hasConnectedResolver) {
-      this._hasConnectedResolver();
-      this._hasConnectedResolver = undefined;
+    // before connection.
+    this.enableUpdating();
+  }
+
+  protected enableUpdating() {
```
justinfagnani: LGTM on this part of the PR. I wouldn't mind this PR split into two for review purposes.

src/lib/decorators.ts
kevinpschaaf: The polyfilled `querySelector` works fine (it queries the logical tree).  It's only CSS selectors that require a wrapper, since they CSS only works on the actual tree.

src/lib/decorators.ts
sorvell: Done

CHANGELOG.md
sorvell: Done

src/lib/decorators.ts
sorvell: That's right, this works OK under polyfill. Note, it's only for styling that we need a node around slots when using ShadyDOM.

src/lib/decorators.ts
sorvell: Done

src/test/lib/decorators_test.ts
sorvell: Yeah, to test `flatten`. Added comments to explain.

src/test/lib/decorators_test.ts
sorvell: Done. One set of elements is used to test both the flatten and non-flattened cases. Added comments to explain.

src/lib/updating-element.ts
sorvell: Agree would have been better that way. If there's additional feedback on the new decorator, will do.

PR #861, opened 2019-12-06
Replace TSLint with ESLint

abdonrd:
Same PR as https://github.com/Polymer/lit-html/pull/969


abdonrd:
Ping to @rictic 😄

src/lib/updating-element.ts
```diff
@@ -266,8 +271,8 @@ export abstract class UpdatingElement extends HTMLElement {
   /** @nocollapse */
   private static _ensureClassProperties() {
     // ensure private storage for property declarations.
-    if (!this.hasOwnProperty(
-            JSCompiler_renameProperty('_classProperties', this))) {
+    if (!Object.prototype.hasOwnProperty.call(
```
justinfagnani: I'm not a huge fan of this rule. We don't expect someone else to be monkey-patching this class or overriding `hasOwnProperty`, and if they do 🤷‍♂️.

src/lib/updating-element.ts
```diff
@@ -300,12 +305,13 @@ export abstract class UpdatingElement extends HTMLElement {
...
     // still overwrite it
-    if (options.noAccessor || this.prototype.hasOwnProperty(name)) {
+    if (options.noAccessor ||
+        Object.prototype.hasOwnProperty.call(this.prototype, name)) {
       return;
     }
     const key = typeof name === 'symbol' ? Symbol() : `__${name}`;
     Object.defineProperty(this.prototype, name, {
-      // tslint:disable-next-line:no-any no symbol in index
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
justinfagnani: I think we can just replace `any` with `unknown` here.

src/lit-element.ts
```diff
@@ -34,9 +34,9 @@ declare global {
 (window['litElementVersions'] || (window['litElementVersions'] = []))
     .push('2.2.1');
 
-export interface CSSResultArray extends Array<CSSResult|CSSResultArray> {}
+export type CSSResultArray = Array<CSSResult|CSSResultArray>
```
justinfagnani: Add semicolon at the end of line.

src/test/lib/decorators_test.ts
```diff
@@ -16,14 +16,15 @@ import {eventOptions, property} from '../../lib/decorators.js';
...
+/* eslint-disable @typescript-eslint/no-explicit-any */
 
 let hasOptions;
 const supportsOptions = (function() {
   if (hasOptions !== undefined) {
     return hasOptions;
   }
-  const fn = () => {};
+  const fn =
+      () => {};  // eslint-disable-line @typescript-eslint/no-empty-function
```
justinfagnani: `no-empty-function` doesn't strike me as a very useful rule either.

abdonrd:
@justinfagnani all comments addressed!

package.json
```diff
@@ -27,25 +28,34 @@
     "test": "npm run build && npm run build:babel-test && wct",
     "checksize": "rollup -c ; rm lit-element.bundled.js",
     "format": "find src test | grep '\\.js$\\|\\.ts$' | xargs clang-format --style=file -i",
-    "lint": "tslint --project ./",
+    "lint": "npm run lint:eslint",
+    "lint:eslint": "eslint 'src/**/*.{js,ts}'",
```
justinfagnani: Why two separate scripts here?

abdonrd: Nothing really special. In other projects I can do more lint (CSS, LitAnalayzer, etc)

abdonrd: Updated!

abdonrd:
How to fix the TS error?

```
Property 'adoptedStyleSheets' does not exist on type 'ShadowRoot'
```

abdonrd:
Rebased on master and updated again!

bennypowers:
> 'adoptedStyleSheets' does not exist on type 'ShadowRoot'

see https://github.com/microsoft/TypeScript/issues/10401#issuecomment-255964760
for an example on how to add properties to a global interface

abdonrd:
Thanks @bennypowers! Fixed!

PR #881, opened 2020-01-20
Add @internalProperty decorator, for private or protected fields.

rictic:


web-padawan:
IMO the name `@internal` would be enough.

rictic:
`@internal` reads to me like it can and perhaps should be combined with `@property`

> Are there any public tools that can take advantage of the `property` vs `internalProperty` distinction? Something in tsickle? If this is mainly useful inside Google, do we want to leave this out of the main docs as is, or add a note to the property docs? cc @arthurevans and @graynorton

tsickle will care, and I intend to add an optional, disabled-by-default pass to lit-analyzer that will require that `@property` is always public and `@internalProperty` is always private or protected.

Unless a project is compiling with closure compiler, they probably won't care about this distinction at all, and so it's probably not worth advertising this.

PR #888, opened 2020-01-30
fix Update fromAttribute complex type when value is not valid

### LINKED ISSUE #722: Uncaught SyntaxError: Unexpected token g in JSON at position 0

benjaminParisel:
For better understanding on the issue please check #722

Following pull request #723. I added test and take in account return made.

This happens because JSON.parse('') will be fired when value is an invalid JSON. 

So now it's return an empty array or empty object.

**Reference Issue**

Fixes #722

### DIFF
--- a/src/lib/updating-element.ts
+++ b/src/lib/updating-element.ts
@@ -112,6 +112,14 @@ export interface PropertyDeclaration<Type = unknown, TypeHint = unknown> {
    * the property changes.
    */
   readonly noAccessor?: boolean;
+
+  /**
+   * Indicates fallback value will be return when parsing in
+   * `fromAttribute` catch an exception.
+   * By default, fallback value is `undefined`
+   */
+  // tslint:disable-next-line:no-any
+  readonly onError?: any;
 }
 
 /**
@@ -157,18 +165,12 @@ export const defaultConverter: ComplexAttributeConverter = {
       case Number:
         return value === null ? null : Number(value);
       case Object:
-        try {
-          return JSON.parse(value!);
-        } catch (e) {
-          // Ignore any previous errors. Only return a empty object
-          return {};
-        }
       case Array:
         try {
           return JSON.parse(value!);
         } catch (e) {
-          // Ignore any previous errors. Only return a empty array
-          return [];
+          // Ignore any previous errors. Only return a empty object
+          return defaultPropertyDeclaration.onError;
         }
     }
     return value;
@@ -194,7 +196,8 @@ const defaultPropertyDeclaration: PropertyDeclaration = {
   type: String,
   converter: defaultConverter,
   reflect: false,
-  hasChanged: notEqual
+  hasChanged: notEqual,
+  onError: undefined
 };
 
 const STATE_HAS_UPDATED = 1;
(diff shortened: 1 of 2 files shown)


benjaminParisel:
I signed it!

benjaminParisel:
@googlebot I signed it!

src/lib/updating-element.ts
```diff
@@ -157,8 +157,19 @@ export const defaultConverter: ComplexAttributeConverter = {
       case Number:
         return value === null ? null : Number(value);
       case Object:
+        try {
+          return JSON.parse(value!);
+        } catch (e) {
+          // Ignore any previous errors. Only return a empty object
+          return {};
```
justinfagnani: I don't think we should be returning empty objects/arrays in this case. Downstream code would not be able to differentiate between an error value and an empty object. I think `undefined` is a better fallback, even though it's still valid JSON.

Another option is to add an `onError` property to `PropertyDeclaration` and call it here.

benjaminParisel: onError property sounds good

benjaminParisel:
Can i have an update about this contribution please @justinfagnani ?

src/lib/updating-element.ts
```diff
@@ -112,6 +112,14 @@ export interface PropertyDeclaration<Type = unknown, TypeHint = unknown> {
...
    */
   readonly noAccessor?: boolean;
+
+  /**
+   * Indicates fallback value will be return when parsing in
+   * `fromAttribute` catch an exception.
+   * By default, fallback value is `undefined`
+   */
+  // tslint:disable-next-line:no-any
+  readonly onError?: any;
```
felipeplets, line 122:
```suggestion
  readonly onError?: (error: Error) => void;
```

benjaminParisel:
@justinfagnani Can you check my PR, please?  thank you in advance

benjaminParisel:
Job is done in another [commit](https://github.com/Polymer/lit-html/commit/15ec2b7d02c0292df2a33d5c4ff4e850753f5991). I close this PR.

PR #902, opened 2020-02-21
Fix compilation with TypeScript 3.8, which updated custom element types.

rictic:


justinfagnani:
You need to commit the package lock too.

rictic:
Ok, PTAL

PR #904, opened 2020-02-24
Adds `@queryAsync` decorator

### LINKED ISSUE #903: Add `@asyncQuery` decorator
## Initiative / goal
The `@query` decorator returns an element from the currently rendered element DOM; however, it is sometimes useful to query after the element has updated. This ensures the element update state is resolved before the query is performed. In addition, if the result of the query is itself a LitElement, it is often useful for it to be updated before the query is made.

### Scope
Create an `@asyncQuery` decorator which creates a getter which returns a promise that resolves to the result of querying the passed selector after the element has updated.

For example, given a render method like:

```
@asyncQuery('#foo)
foo: HTMLDivElement;

render() {
  return html`${this.special ? 
    html`<div id=foo>Special Foo</div>` : 
    html`<div id=foo>Foo</div>`
  };
}

async toggleSpecialAndDoSomethingWithFoo() {
  this.special = !this.special;
  const text = (await this.foo).textContent;
}
```

Feature is implemented and tests pass.

sorvell:
Fixes #903 by adding an `@asyncQuery` decorator.


src/lib/decorators.ts
```diff
@@ -225,6 +225,54 @@ export function query(selector: string) {
...
+  return (protoOrDescriptor: Object|ClassElement,
+          // tslint:disable-next-line:no-any decorator
+          name?: PropertyKey): any => {
+    const descriptor = {
+      async get(this: LitElement) {
+        await this.updateComplete;
+        const el = this.renderRoot.querySelector(selector);
+        // if this is a LitElement, then await updateComplete
+        if (el) {
+          await (el as LitElement).updateComplete;
```
kevinpschaaf: Is this good/required?

It sort of confuses the question "why should a user use `asyncQuery`?"
1. Because you need a reference to a child element but it won't be there until after _this_ element flushes? 
1. Because you've changed state on the _child_ element and for some reason don't want to interact with it until after it's flushed?

The latter should be rare; flushing a child should normally be unobservable except for measurement (and we know that's fraught due to nested/chained updates). It's sort of lumping two use cases together, and given the second one is problematic, just want to discuss whether the child await is defensible vs. just pushing the entire measurement case to user responsibility.

src/lib/decorators.ts
```diff
@@ -225,6 +225,54 @@ export function query(selector: string) {
   };
 }
 
+/**
+ * A property decorator that converts a class property into a getter that
```
justinfagnani: Why is this preferable over:

```ts
class {
  @query('#foo') foo;

  x() {
    await this.updateComplete;
    this.foo;
  }
}
```

ie, when would we tell someone to use one pattern over the other?

src/lib/decorators.ts
justinfagnani: This also causes another microtask wait for every use, regardless of whether the element is a LitElement. And it doesn't work for other asynchronously rendering components like Stencil.

src/lib/decorators.ts
```diff
@@ -225,6 +225,54 @@ export function query(selector: string) {
...
+ *
+ * @param selector A DOMString containing one or more selectors to match.
+ *
+ * See: https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector
+ *
+ * @example
+ *
+ *     class MyElement {
+ *       @asyncQuery('#first')
+ *       first;
```
justinfagnani: I think the example should show the usage of this.first somewhere too.

src/lib/decorators.ts
sorvell: Yup, removing this part.

src/lib/decorators.ts
sorvell: Inside the element, this is arguably just slightly sugared. For a property that needs to be accessed external to the element, using `@asyncQuery` is much cleaner since the user doesn't have to know about `updateComplete`.

justinfagnani: Do we have use cases where an element wants to expose part of it's internals like this?

sorvell: This is useful when a property that is a query can change with element state. It's preferable when the property is public so that the user does not have to know about the `updateComplete` promise.

src/lib/decorators.ts

src/lib/decorators.ts
justinfagnani: I think this is potentially very useful if made a bit more flexible:

```ts
export function asyncQuery(selector: string, waitFor?: (el) => Promise<unknown>) {
  if (waitFor) { await waitFor(el); }
...
```

Then one could do:

```ts
@asyncQuery('#foo', (e) => e.updateComplete) foo;
```

Or the equivalent for Stencil...

sorvell: Added a comment about supporting this in the future.

PR #914, opened 2020-03-05
Makes `createProperty` easier to use to customize properties

### LINKED ISSUE #911: Make LitElement's property accessors creation and metadata easily customizable
## Initiative / goal
LitElement configures reactive properties, but this behavior is currently difficult to extend and customize. Making property creation and metadata easily customizable would unlock many use specific cases that users have asked for but that we do not want to burden everyone with by adding the behavior to the core library.

### Scope
* Make property metadata easily customizable. 
  * One approach is to create static methods for setting/getting property metadata for a property.
  * This work must also ensure that metadata can continue to be typed correctly in TypeScript. 
* Make property accessors easily customizable. While users can currently override `static createProperty`, the code for making a property reactive is private.
  * One approach is to rename and expose the private `_requestUpdate` method.
  * This leaves private storage as the only not easily callable work done in `createProperty`. This could be sugared or left to overriders to deal with.

Add a set of tests that prove that the requested use cases can be implemented by adding this features. For example:
* ability to override default property metadata (#891)
* providing a property validator (#905)
* making properties have synchronous effects (may only be part of this) (#643)

sorvell:
Fixes #911. It is now easier to override property accessor creation. A new static `getPropertyDescriptor(name: PropertyKey, key: string|symbol, options: PropertyDeclaraction)` is added here. This method should return a [property descriptor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty) to use when defining the property. If no descriptor is returned, the property will not become an accessor.

Example:

```
  static getPropertyDescriptor(
      name: PropertyKey,
      key: string|symbol,
      options: PropertyDeclaration) {

    const defaultDescriptor = super.getPropertyDescriptor(name, key, options); 
    const setter = defaultDescriptor.set;
    return Object.assign(descriptor, {
      set(this: MyElement, value: unknown) {
        setter.call(this, value);
        console.log('Property', name, 'was set to', value);
     }
  }
  //...
}
```

### DIFF
--- a/src/lib/updating-element.ts
+++ b/src/lib/updating-element.ts
@@ -174,7 +174,7 @@ export interface HasChanged {
 }
 
 export type PropertyDescriptorFactory = (options: PropertyDeclaration,
-  descriptor: PropertyDescriptor, key: string|symbol) => PropertyDescriptor|null|void;
+  descriptor: PropertyDescriptor, key: string|symbol) => PropertyDescriptor|undefined;
 
 /**
  * Change function that returns true if `value` is different from `oldValue`.
@@ -312,7 +312,7 @@ export abstract class UpdatingElement extends HTMLElement {
       return;
     }
     const key = typeof name === 'symbol' ? Symbol() : `__${name}`;
-    let descriptor: PropertyDescriptor|null|void = {
+    let descriptor: PropertyDescriptor|undefined = {
       // tslint:disable-next-line:no-any no symbol in index
       get(): any {
         return (this as {[key: string]: unknown})[key as string];
@@ -329,7 +329,7 @@ export abstract class UpdatingElement extends HTMLElement {
     if (typeof descriptorFactory === 'function') {
       descriptor = descriptorFactory(options, descriptor, key);
     }
-    if (descriptor != null) {
+    if (descriptor !== undefined) {
       Object.defineProperty(this.prototype, name, descriptor);
     }
   }
--- a/src/lib/updating-element.ts
+++ b/src/lib/updating-element.ts
@@ -51,7 +51,7 @@ export interface ComplexAttributeConverter<Type = unknown, TypeHint = unknown> {
 }
 
 type AttributeConverter<Type = unknown, TypeHint = unknown> =
-    ComplexAttributeConverter<Type>|((value: string, type?: TypeHint) => Type);
+    ComplexAttributeConverter<Type>|((value: string|null, type?: TypeHint) => Type);
 
 /**
  * Defines options for a property accessor.
@@ -115,7 +115,7 @@ export interface PropertyDeclaration<Type = unknown, TypeHint = unknown> {
 
   // Allows extension while preserving the ability to use the
   // @property decorator.
-  [index: string]: any;
+  [index: string]: unknown;
 }
 
 /**
@@ -172,6 +172,9 @@ export const defaultConverter: ComplexAttributeConverter = {
 export interface HasChanged {
   (value: unknown, old: unknown): boolean;
 }
+
+export type PropertyDescriptorFactory = (options: PropertyDeclaration,
+  key: string|symbol,descriptor: PropertyDescriptor) => PropertyDescriptor|null|void;
 
 /**
  * Change function that returns true if `value` is different from `oldValue`.
@@ -293,11 +296,13 @@ export abstract class UpdatingElement extends HTMLElement {
    */
   static createProperty(
       name: PropertyKey,
-      options: PropertyDeclaration = defaultPropertyDeclaration) {
+      options: PropertyDeclaration = defaultPropertyDeclaration,
+      descriptorFactory?: PropertyDescriptorFactory) {
     // Note, since this can be called by the `@property` decorator which
     // is called before `finalize`, we ensure storage exists for property
     // metadata.
-    this.setOptionsForProperty(name, options);
+    this._ensureClassProperties();
+    this._classProperties!.set(name, options);
     // Do not generate an accessor if the prototype already has one, since
     // it would be lost otherwise and that would never be the user's intention;
     // Instead, we expect users to call `requestUpdate` themselves from
@@ -307,7 +312,7 @@ export abstract class UpdatingElement extends HTMLElement {
       return;
     }
     const key = typeof name === 'symbol' ? Symbol() : `__${name}`;
-    Object.defineProperty(this.prototype, name, {
+    let descriptor: PropertyDescriptor|null|void = {
       // tslint:disable-next-line:no-any no symbol in index
       get(): any {
         return (this as {[key: string]: unknown})[key as string];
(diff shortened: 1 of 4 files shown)


src/lib/updating-element.ts
```diff
@@ -304,7 +312,7 @@ export abstract class UpdatingElement extends HTMLElement {
       return;
     }
     const key = typeof name === 'symbol' ? Symbol() : `__${name}`;
-    Object.defineProperty(this.prototype, name, {
+    let descriptor: PropertyDescriptor|null|void = {
```
justinfagnani: Is `descriptorFactory` expected to mutate the passed in descriptor, or return a new one? Rather than passing a description, what's the information needed to create a new one?

src/lib/updating-element.ts
```diff
@@ -304,7 +312,7 @@ export abstract class UpdatingElement extends HTMLElement {
       return;
     }
     const key = typeof name === 'symbol' ? Symbol() : `__${name}`;
```
justinfagnani: I could see a custom property definer wanting to customize this key. If you move this into the default property descriptor factory, then overriders can do whatever they want here.

src/lib/updating-element.ts
```diff
@@ -289,7 +296,8 @@ export abstract class UpdatingElement extends HTMLElement {
    */
   static createProperty(
       name: PropertyKey,
-      options: PropertyDeclaration = defaultPropertyDeclaration) {
+      options: PropertyDeclaration = defaultPropertyDeclaration,
+      descriptorFactory?: PropertyDescriptorFactory) {
```
justinfagnani: Rather than a function-valued argument, why not another overridable method? I'd rather be consistent in allowing customization via overrides than a mix of overrides and values. Unless there's a reason the factory needs to be effectively final.

src/lib/updating-element.ts
```diff
@@ -169,6 +173,9 @@ export interface HasChanged {
   (value: unknown, old: unknown): boolean;
 }
 
+export type PropertyDescriptorFactory = (options: PropertyDeclaration,
```
justinfagnani: Could this type be just: `(options: PropertyDeclaration) => PropertyDescriptor | undefined`?

The return type should be tightened up or documented here. `void` in the type seems to allow for implementations that don't return anything, but that looks like it'll nuke the descriptor at the call site. Likewise, what exactly does `null` signify, and why `null` over `undefined`?

sorvell: Changed.

src/lib/updating-element.ts
sorvell: That's fair. The goal was to not change the shape of the existing API, which could be breaking, while providing the ability to customize. Currently `createProperty` does 4 things:

 1. stores property metadata
 1. decides whether or not to create an accessor
 1. creates a property descriptor 
 1. creates the accessor using the descriptor 

I considered making new entry points for (1) and (4) that overriders could simply call, without the need to call `super`; however, this meant users would have to do (2) on their own and always do (3) from scratch (cumbersome).

I also wanted to avoid reshaping the API, for example having `createProperty` return a descriptor, to avoid breaking users that are currently overriding `createProperty`. Admittedly it's currently difficult to override in a useful way, but it is possible.

src/lib/updating-element.ts
sorvell: The key is for private property storage and is provided only so that users can follow the same convention as LitElement. Overriders can choose to use or ignore the key in the descriptor they return via the factory.

src/lib/updating-element.ts
sorvell: Either. To create an entirely new descriptor, one mainly needs the `PropertyDeclaration` provided in `options`. The `key` is available so that the user can easily follow LitElement's convention for private property storage.

The previous descriptor is provided because in some of the use cases evaluated (e.g. making a property validator), only the setter needs to be customized.

sorvell: Removed factory in favor of `createPropertyDescriptor`.

src/lib/updating-element.ts
sorvell: Removed factory in favor of `createPropertyDescriptor` which is called in the default implementation of `createProperty`.

src/test/lib/updating-element_test.ts
```diff
@@ -1802,6 +1802,134 @@ suite('UpdatingElement', () => {
         assert.equal(sub.getAttribute('foo'), '5');
       });
 
+    test('can provide a default property declaration', async () => {
```
kevinpschaaf: Just pointing out, now that `getDeclarationForProperty` is exposed, this use case _could_ be covered by overriding that method, albeit less efficiently (O.assign on every get).  But might want to double-check the code is safe to someone doing that, i.e. the map is never accessed directly but always through that method?

src/lib/updating-element.ts
```diff
@@ -412,11 +453,14 @@ export abstract class UpdatingElement extends HTMLElement {
...
-        defaultConverter.toAttribute;
-    return toAttribute!(value, type);
+    let converter = options.converter;
+    let toAttribute =
+        converter && (converter as ComplexAttributeConverter).toAttribute;
+    if (!toAttribute) {
+      converter = defaultPropertyDeclaration.converter;
+      toAttribute = converter && (converter as ComplexAttributeConverter).toAttribute;
+    }
+    return toAttribute ? toAttribute(value, type) : value;
```
kevinpschaaf: Didn't follow the significance of the change from `defaultConverter` to `defaultPropertyDeclaration.converter`, nor how `toAttribute` could end up being falsey by the end here....

src/lib/updating-element.ts
```diff
@@ -679,6 +721,9 @@ export abstract class UpdatingElement extends HTMLElement {
    * ```
    */
   protected performUpdate(): void|Promise<unknown> {
+    if (!this._hasRequestedUpdate) {
+      return;
```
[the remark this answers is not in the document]
sorvell: Oops, will remove.

src/lib/updating-element.ts
sorvell: From another refactoring. Will revert...

src/test/lib/updating-element_test.ts
sorvell: Added doc note that it should not be implemented.

src/lib/updating-element.ts
sorvell: This is now included in `createPropertyDescriptor(name, key, options)` so implementers can do whatever they want with it.

src/lib/updating-element.ts
```diff
@@ -317,7 +367,22 @@ export abstract class UpdatingElement extends HTMLElement {
...
+
+  /**
+   * Returns the property options associated with the given property.
+   * These options are defined with a PropertyDeclaration via the `properties`
+   * object or the `@property` decorator and are registered in
+   * `createProperty(...)`.
+   *
+   * Note, this method should be considered "final" and not overridden. To
+   * customize the options for a given property, override `createProperty`.
+   *
```
justinfagnani: Maybe add a `@final` annotation. Not sure if these flow through from TypeScript to Closure via tsickle or not, but it doesn't hurt.

src/lib/updating-element.ts
```diff
@@ -317,7 +367,22 @@ export abstract class UpdatingElement extends HTMLElement {
...
+   */
+  protected static getPropertyOptions(name: PropertyKey) {
```
justinfagnani: Can this be private? It seems to only be accessed from other private methods. If so, rename to `_getPropertyOptions`

src/lib/updating-element.ts
```diff
@@ -304,7 +323,38 @@ export abstract class UpdatingElement extends HTMLElement {
...
+   *         },
+   *         configurable: true,
+   *         enumerable: true
+   *       }
+   *     }
+   *   }
+   *
+   * @nocollapse
+   */
+  protected static createPropertyDescriptor(name: PropertyKey,
```
justinfagnani: Should we name this `get...`? On `getX` vs. `createX` it seems like the distinction is that `createX` is side effectful and `getX` isn't.

src/lib/updating-element.ts
sorvell: It's protected so options can be used in the element lifecycle. For example, in `updated` you might want to act on a given changed property if it has a specific option.

src/lib/updating-element.ts

src/lib/updating-element.ts
sorvell: Sure.

PR #917, opened 2020-03-06
Only call lit-html render if LitElement subclass implements render

dfreedm:
- #712 Introduced a breaking behavior change for situations where
  `render` is unimplemented, and DOM is added before being connected to
  the document.

### DIFF
--- a/src/test/lit-element_test.ts
+++ b/src/test/lit-element_test.ts
@@ -277,10 +277,6 @@ suite('LitElement', () => {
           createRenderRoot() {
             return this;
           }
-          connectedCallback() {
-            super.connectedCallback();
-            this.addedDom = this.renderRoot.querySelector('div');
-          }
         }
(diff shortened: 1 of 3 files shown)


src/test/lit-element_test.ts
```diff
@@ -266,4 +269,32 @@ suite('LitElement', () => {
...
       });
+
+  test(
+      'if `render` is unimplemented, do not overwrite renderRoot', async () => {
+        class A extends LitElement {
+          addedDom: HTMLElement|null = null;
+          createRenderRoot() {
+            return this;
+          }
+          connectedCallback() {
```
justinfagnani: This test works, but I think the connected callback is unnecessary. The test code can just assert that the child is still present after the update.

dfreedm: Done

Issue #924, opened 2020-03-14
Docs: background-color is not inherited CSS property


ianwith:
## Initiative / goal
Fix document inaccurate description about CSS inheritance.
https://lit-element.polymer-project.org/guide/styles#inheritance

### Scope
`background` and `background-color` are not inherited CSS properties.

According to https://developer.mozilla.org/en-US/docs/Web/CSS/background-color

`background-color` has an Initial value `transparent`, and that's why they look like they inherited their parents' `background-color`.

PR #925, opened 2020-03-14
Correct docs: styles - CSS inheritance

ianwith:
### Reference Issue
Fixes https://github.com/Polymer/lit-element/issues/924


ianwith:
@googlebot I signed it!

PR #931, opened 2020-03-17
Changed `_getUpdateComplete` to `getUpdateComplete`

sorvell:
Protected methods should not have a leading underscore.


sorvell:
This AIP has already been released so we're not going to merge this change. =(

PR #933, opened 2020-03-18
Prepare 2.3.0 release

justinfagnani:


src/lib/updating-element.ts
```diff
@@ -333,12 +333,9 @@ export abstract class UpdatingElement extends HTMLElement {
    *
    *   class MyElement extends LitElement {
    *     static getPropertyDescriptor(name, key, options) {
-   *       const defaultDescriptor = super.getPropertyDescriptor(name, key, options);
```
kevinpschaaf: Some unfortunate auto-formatting here... :(

justinfagnani: fixed. clang-format isn't that smart

PR #936, opened 2020-03-19
Use downlevel-dts, test compile in TS 3.4

### LINKED ISSUE #935: Broken compile with TypeScript 3.4
## Initiative / goal
Clarity regarding versioning.

## Background
Today our builds broke because we are using lit-element@^2.2.0 combined with Angular 8 (using typescript 3.4.x). I debugged for a while and found that this happens because In v2.3.0 of lit-element the generated typescript definitions were generated with typescript 3.8.x and that version is incompatible with 3.4.x. I was assuming lit-element was following semver but after a closer look i couldn't find any documentation on it. The point of this PR is just to make it clearer that we may have breaking changes when the minor flag changes.

### Scope
Clarify, in README.md possibly, on what versioning strategy lit-element is following.

rictic:
Fixes #935


ts3.4/tsconfig.json
```diff
@@ -0,0 +1,28 @@
+{
```
justinfagnani: Why do we need this file? It's not mentioned in the downlevel-dts instructions.

rictic: See step #4 of https://github.com/sandersn/downlevel-dts#usage

justinfagnani: Oops. I missed that.

Do we want to check this copy in though?

rictic: Good point, done

PR #937, opened 2020-03-19
Prepare for release of v2.3.1

rictic:
I don't have the full release process paged in, but I know this is an important step!


justinfagnani:
You need to update the version numbers in package.json, src/lit-element.ts, and regenerate the package-lock after that.

PR #941, opened 2020-03-21
Improve API docs

### LINKED ISSUE #940: Improve API docs
The current version of Typedoc is not generating very useful docs. Investigate upgrading to the latest beta to take advantage of the new "library" mode, and test other improvements like the @packageDocumentation annotation.

arthurevans:
Fixes #940 

This branch currently uses a beta version of Typedoc; we may want to hold off on merging until Typedoc 0.18 is released.

Improvements:

- Re-exported symbols are documented (for example: css, html, unsafeCSS show up in the lit-element modules).
- Decorator API docs now show up. On the lit-element module page, they show up in their own section ("Decorator functions").
- Example code for decorator functions is formatted correctly.
- The API doc for the updating-element module is no longer a license header.
- Functional links added in a number of places in place of static text ("css tag function" replaced with "[[`css`]] tag function" which creates a link to the right API member).

Staged: https://20200320t185110-dot-polymer-lit-element.appspot.com/api/modules/_lit_element_.html


justinfagnani:
This is a _massive_ improvement. I say ship it ASAP :)

src/lit-element.ts
```diff
@@ -41,6 +41,14 @@ export interface CSSResultArray extends Array<CSSResult|CSSResultArray> {}
  */
 const renderNotImplemented = {};
 
+/**
+ * Base element class that manages element properties and attributes, and
+ * renders a lit-html template.
+ * 
+ * To define a component, subclass `LitElement` and implement a
+ * `render` method to provide the component's template. Define properties 
+ * using the [[`properties`]] property or the [[`property`]] decorator.
```
justinfagnani: Do we need both the `[[` and the ` here?

src/lib/decorators.ts
```diff
@@ -82,7 +82,7 @@ const standardCustomElement =
  *   }
  * }
  * ```
- *
+ * @category Decorator
```
justinfagnani: 😮🎉

package.json
```diff
@@ -24,7 +24,8 @@
   "scripts": {
     "build": "tsc && downlevel-dts . ts3.4 && cp tsconfig.json ./ts3.4/",
     "build:babel-test": "babel src/test/lib/decorators_test.ts --out-file test/lib/decorators-babel_test.js",
-    "gen-docs": "typedoc --readme docs/_api/api-readme.md --tsconfig tsconfig_apidoc.json --mode modules --theme docs/_api/theme --excludeNotExported --excludePrivate --ignoreCompilerErrors --exclude '{**/*test*,**/node_modules/**,**/test/**}' --out ./docs/api --gaID UA-39334307-23 src/**/*.ts",
+    "gen-docs-test": "typedoc --readme docs/_api/api-readme.md --mode library --theme docs/_api/theme --excludeNotExported --excludePrivate --ignoreCompilerErrors --exclude '{**/*test*,**/node_modules/!(lit-html)/**,**/test/**}' --out ./docs/api --gaID UA-39334307-23 src/lit-element.ts src/lib/updating-element.ts",
```
justinfagnani: what's the difference between these two?

arthurevans: Ahh... One belongs here, and the other I should've deleted before sending the PR 😁

src/lit-element.ts
arthurevans: If we use just `[[`, we get the link in regular text font. To my eye it looks a little funny if we have the API members in code font everywhere _except_ in links.

PR #944, opened 2020-03-23
Change order of benefits of LitElement

jbingham:
To put the biggest differentiator first.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -51,11 +51,11 @@ Easily Create Fast, Lightweight Web Components
 <div class="responsive-row">
 <div style="flex: 1">
 
-<h2 class="caption">Delightfully declarative</h2>
+<h2 class="caption">Seamlessly interoperable</h2>
 
-LitElement's simple, familiar development model makes it easier than ever to build Web Components.
+LitElement follows the [Web Components standards](https://developer.mozilla.org/en-US/docs/Web/Web_Components), so your components will work with any framework.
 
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -51,11 +51,11 @@ Easily Create Fast, Lightweight Web Components
...
 <div style="flex: 1">
 
-<h2 class="caption">Delightfully declarative</h2>
+<h2 class="caption">Seamlessly interoperable</h2>
 
-LitElement's simple, familiar development model makes it easier than ever to build Web Components.
+LitElement follows the [Web Components standards](https://developer.mozilla.org/en-US/docs/Web/Web_Components), so your components will work with any framework.
 
-Express your UI declaratively, as a function of state. No need to learn a custom templating language – you can use the full power of JavaScript in your templates. Elements update automatically when their properties change.
+LitElement uses Custom Elements for easy inclusion in web pages, and Shadow DOM for encapsulation. There’s no new abstraction on top of the web platform.
```
kenchris: You can opt out of using shadow DOM though

jbingham: True!

graynorton: Sure, but that's a detail, not a headline.

jbingham:
Yes, my thinking is that the main comparison we're targeting with this page is LitElement vs a non-WC framework.

Developers who already know they want WCs are are motivated to scan the headings and read more.

justinfagnani:
I think we can close this and focus on the new site :)

PR #946, opened 2020-03-24
[docs] call template functions in composition examples

elf-pavlik:
Passing just the reference to the function doesn't seem to work, also lit-html docs don't state support for it: https://lit-html.polymer-project.org/guide/template-reference#supported-data-types-for-text-bindings

> Text content bindings accept a large range of value types:
>  * Primitive values.
>  * TemplateResult objects.
>  * DOM nodes.
>  * Arrays or iterables.

Calling the function results in passing TemplateResult which seems to work fine.

### DIFF
--- a/docs/_guide/templates.md
+++ b/docs/_guide/templates.md
@@ -366,9 +366,9 @@ You can compose LitElement templates from other LitElement templates. In the fol
 class MyPage extends LitElement {
   render() {
     return html`
-      ${this.headerTemplate}
-      ${this.articleTemplate}
-      ${this.footerTemplate}
+      ${this.headerTemplate()}
+      ${this.articleTemplate()}
+      ${this.footerTemplate()}
     `;
   }
   get headerTemplate() {
@@ -493,7 +493,7 @@ html`<button @click="${this.doStuff}"></button>`;
 
 render() {
   return html`
-    ${this.headerTemplate}
+    ${this.headerTemplate()}
     <article>article</article>
   `;
 }


justinfagnani:
The functions in questions are getters, so they don't need to be called.

I agree this is a bit confusing though. I would be open to rewriting as plain instance methods. WDYT @arthurevans ?

elf-pavlik:
🤦‍♂ my bad not noticing use of getters! Actually I find it very elegant way for templates that don't need any parameters passed.

It might come instructive to add some variables to composition example. For example making articles template a list:

```ts
class MyPage extends LitElement {

  @property({ type: Array })
  articles = []

  render() {
    return html`
      ${this.headerTemplate}
      ${this.articlesList(this.articles)}
      ${this.footerTemplate}
    `;
  }

  articleTemplate(article) {
    return html`<article>article</article>`;
  }

  articlesList(articles) {
    return html`${articles.map(this.articleTemplate)}`
  }

  get headerTemplate() {
    return html`<header>header</header>`;
  }

  get footerTemplate() {
    return html`<footer>footer</footer>`;
  }
}
```

In case above personally I find it easier to follow if we pass articles `this.articlesList(this.articles)` rather than in `articlesList` access `this.articles` directly. Not sure if it makes any difference other than personal style. Still even if one would also make `articlesList` a getter, `articleTemplate` would still need to stay a function to use it with `map`.

Please let me know if you would prefer to close this PR and discuss possible improvements of composition examples in a new issue.

arthurevans:
I've updated this example in an ongoing PR so it includes both regular functions (for header and article) and a getter (for footer) just to demonstrate these two approaches. 

https://github.com/Polymer/lit-element/pull/1004/commits/84698253d88068c11b00e84d7bfa079accd4771a

If that addresses the concern, we can close this PR.

elf-pavlik:
Great, thank you!

PR #949, opened 2020-04-06
Document common use cases for lifecycle

### LINKED ISSUE #912: Improve documentation on lifecycle methods
## Initiative / goal

On the Polymer slack (new) users of lit-element often ask "How can I achieve ...", and in many cases the answer is "use lifecycle method ...". One question I noticed frequently is "How can I access attribute values in the constructor?"
(The answer is, of course: "You can't. Use a lifecycle method.")

This indicates to me, that this information is not easy to extract from the documentation. I believe the current documentation can be improved by adding guidance, which method to use for which common case. (In its current state, the documentation seems more like a technical reference to me.)

### Scope

My suggestions would be to do the following:

- [ ] Split current lifecycle documentation in two parts: "reference" and "common use cases" (titles open for discussion)
- [ ] Write set of common use cases (should refer to the appropriate paragraphs in "reference")

- Lifecycle documentation contains reference section
- Lifecycle documentation contains common use cases (or scenarios)

djlauk:
As discussed on polymer slack, I gave it a shot and documented some common use cases.

I decided to use strong emphasis (bold) in the list to make it easily scannable.

### Reference Issue
Fixes #912

### DIFF
--- a/docs/_guide/lifecycle.md
+++ b/docs/_guide/lifecycle.md
@@ -33,7 +33,7 @@ You can change this behavior so that Step 3 awaits a Promise before performing t
 
 For a more detailed explanation of the browser event loop, see [Jake Archibald's article](https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/).
 
-#### Lifecycle callbacks
+#### Lifecycle callbacks {#lifecyclecallbacks}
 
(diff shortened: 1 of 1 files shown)


arthurevans:
Thank you! I'm kind of swamped right now, but wanted to let you know it was in my queue.

Issue #643, opened 2019-04-07
As a developer I would like to do sync actions on property changes


daKmoR:
### Description
There is no public API that allows for sync actions like logging or "computed" properties.

Example use case:
Logging `We like cats too :)` whenever the property `mood` gets set to `cat`.

#### Live Demo
https://stackblitz.com/edit/lit-element-example-nkwdco?file=index.js

#### Expected Results
```
=> set to cat
We like cats too :)
=> set to foo
=> set to cat
We like cats too :)
```

#### Actual Results (when using public API `update`)
```
=> set to cat
=> set to foo
=> set to cat
We like cats too :)
```

Code for reference:
```js
  update(changedProperties) {
    super.update(changedProperties);
    if (changedProperties.has('mood')) {
      if (this.mood === 'cat') {
        console.log('We like cats too :)');
      }
    }
  }
```

## Note
Workaround with _requestUpdate works but it's a private API.

Code for reference:
```
  _requestUpdate(name, oldValue) {
    super._requestUpdate(name, oldValue);
    if (name === 'mood') {
      if (this.mood === 'cat') {
        console.log('We like cats too :)');
      }
    }
  }
```

### Versions
- lit-element: v2.1.0

daKmoR:
another usage example
```js
this.firstName = 'foo';
// fullName should be updated whenever firstName or lastName changes

if (this.fullName.contains('foo')) {
  // do something 
  // never reached as fullName will only be updated in the next microtask if using `update`
}
```

CaptainCodeman:
You could use the `hasChanged` method:
```
  static get properties() {
    return { 
      mood: {type: String, hasChanged: (newVal, oldVal) => {
        if (newVal === 'cat') {
          console.log('We like cats too :)');
        }
        return true
      }}
    }
  }
```

daKmoR:
hmmm using `hasChanged` probably could work but that certainly looks like abuse 🙈 
also when extending web components it's tough as you can't easily add another function in your child...

blikblum:
> using hasChanged probably could work but that certainly looks like abuse

and you don't have access to component instance (`this`). So anything that needs to access the instance won't work

Westbrook:
@daKmoR On the side of logging, to better inform the outcomes you hope to see in this request, can you explain how you see this as being different than a custom getter/setter?
```
get mood() { return this._mood; }
set mood(mood) {
  if (mood === this.mood) return;
[...]

PR #959, opened 2020-04-13
Makes `performUpdate` possible to call synchronously

sorvell:
This should address #643. While we're not adding direct support, it's now possible to create properties that perform synchronous updates by calling `performUpdate` in the setter.


blikblum:
I fear this will not help #643 at all.

Basically it needs to intercept when a arbitrary, previously defined property is set / changed independently if will flush changes synchronous or not.

src/lib/updating-element.ts
```diff
@@ -751,6 +751,12 @@ export abstract class UpdatingElement extends HTMLElement {
    * ```
    */
   protected performUpdate(): void|Promise<unknown> {
+    // Abort any update if one is not pending when this is called.
+    // This can happen if `performUpdate` as called early to "flush"
```
kevinpschaaf, line 755:
```suggestion
    // This can happen if `performUpdate` is called early to "flush"
```

src/lib/updating-element.ts
```diff
@@ -723,10 +723,10 @@ export abstract class UpdatingElement extends HTMLElement {
     if (result != null) {
       await result;
     }
-    return !this._hasRequestedUpdate;
+    return !this.isUpdatePending;
   }
 
-  private get _hasRequestedUpdate() {
+  protected get isUpdatePending() {
```
kevinpschaaf, line 729: Why is it necessary to expose this (what use case is unlocked)?

kevinpschaaf:
@blikblum As @sorvell mentioned it does not solve #643 directly, but the idea is now a user can easily make a `LitElement` subclass that overrides `getPropertyDescriptor` as added/shown in #914 so that setters for `@property`s can call `performUpdate` synchronously after the base setter enqueues an update.

We're of the opinion that for most users, the default async/batched-render behavior is the correct default approach.  However, this change should allow users with special needs for synchronous updates to be unblocked.

src/lib/updating-element.ts
sorvell: No clear use case. Removing for now.

blikblum:
> As @sorvell mentioned it does not solve #643 directly

It does not solve #643 directly nor indirectly

>  However, this change should allow users with special needs for synchronous updates

#643 is not about changing the update to be synchronous. Is meant to detect synchronously when the property value changed. How LitElement reacts to this change (asynchronously or not) is irrelevant to that use case.

PR #971, opened 2020-04-22
Use Rollup for build

### LINKED ISSUE #970: Use Rollup for bundle build
Docs site currently uses Polymer CLI to build the JS bundle--basically only used for the stackblitz example buttons.

We should replace this with Rollup. 

At the moment the StackBlitz doesn't work at all on IE. We should make sure that something gets displayed in place of the button in this case.

arthurevans:
Fixes #970. Replaces deprecated Polymer CLI with rollup build.

Also ensures that something is displayed in place of the stack-blitz buttons when they fail to load.

Staged here:  https://rollup-test-dot-polymer-lit-element.uk.r.appspot.com/try

On that page, you'll see a 'LAUNCH CODE EDITOR' button. If that button's working, the JavaScript bundle has loaded correctly. (If you load the site in IE or disable JavaScript, you should see a grayed-out button that says, "Code editor not supported in this browser" or words to that effect.)

### DIFF
--- a/docs/_includes/project.html
+++ b/docs/_includes/project.html
@@ -2,4 +2,5 @@
     folder="{{ site.baseurl }}/{{ site.projects }}/{{ include.folder | default: template }}"
     openFile="{{ include.openFile | default: my-element.js }}"
     label="{{ include.label | default: 'Launch Code Editor' }}">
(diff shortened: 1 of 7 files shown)


arthurevans:
Pinging @justinfagnani for review by default.

PR #980, opened 2020-04-29
Update template sample. Fixes #729.

### LINKED ISSUE #729: Docs: code sample should include properties getter
In this section:

https://lit-element.polymer-project.org/guide/templates#design-a-performant-template

The inline code sample doesn't show the `properties` getter. That's important to the sample (since behavior depends on whether or not `message` is defined in `properties`).

arthurevans:
Added a properties getter to this example as suggested in #729. 

Also fixed several examples with broken syntax.

Oh, yeah. Also deleted one example project that we weren't using anywhere.


docs/_includes/projects/docs/templates/databinding/my-element.js
```diff
@@ -3,10 +3,10 @@ import { LitElement, html } from 'lit-element';
 class MyElement extends LitElement {
   static get properties() {
     return {
-      prop1: String,
-      prop2: String,
-      prop3: Boolean,
-      prop4: String
+      prop1: {type: String},
```
justinfagnani, line 6: How about only including one `{type: string}` and make the others `{}` with a `// type: string is the default` comment?

arthurevans: I think if we're going to show an empty declaration, it should probably be in the Properties chapter, not here.

I wasn't sure whether we should consider `{}` an anti-pattern or a convenient shorthand.

justinfagnani: `{}` in JS is the same as `@property()` in TS. It's perfectly fine.

arthurevans: OK. I'm going to try to rationalize these samples in a subsequent PR. Just wanted to unbreak them here.

PR #977, opened 2020-04-27
Minor API doc improvements.

arthurevans:
- Move API doc config to a file.
- FIX API doc metadata (favicon, meta description)
- Improve landing page slightly.

PR #982, opened 2020-05-01
API docs landing page => lit-element module

arthurevans:
This removes one click from almost every visit to the API docs by pointing the main "API" link in the topnav at the lit-element module, instead of a hand-rolled landing page (the landing page is still there, in case people have it bookmarked). 

Adds some module-level TypeDoc to the lit-element module, matching text from the landing page.

Also includes the changes previously in #977:

- Move API doc config to a file.
- FIX API doc metadata (favicon, meta description)
- Improve landing page slightly.

Staged as https://api-landing-page-dot-polymer-lit-element.uk.r.appspot.com


docs/_api/api-readme.md
```diff
@@ -1,35 +1,48 @@
...
+    return html`<p>your ${adjective} template here</p>`;
   }
 }
 
 customElements.define('my-element', MyElement);
 ```
 
+`LitElement` extends [`UpdatingElement`](/api/classes/_lib_updating_element_.updatingelement.html) and adds lit-html templating.
+
 ### [lib/updating-element](/api/modules/_lib_updating_element_.html)
```
justinfagnani, line 40: Should we update the module name to `lit-element/lib/updating-element.js` since that's the import that needs to be used?

src/lit-element.ts
```diff
@@ -11,14 +11,63 @@
...
+ */
 import {render, ShadyRenderOptions} from 'lit-html/lib/shady-render.js';
 
 import {PropertyValues, UpdatingElement} from './lib/updating-element.js';
 
 export * from './lib/updating-element.js';
 export * from './lib/decorators.js';
-export {html, svg, TemplateResult, SVGTemplateResult} from 'lit-html/lit-html.js';
-import {supportsAdoptingStyleSheets, CSSResult, unsafeCSS} from './lib/css-tag.js';
+export {
```
justinfagnani, line 60: All this reformatting looks like it was done with something other than `npm run format`? Did your editor do this?

arthurevans: Frak. Missed this, let me see what I can do about it.

docs/_api/api-readme.md
arthurevans: SGTM

PR #983, opened 2020-05-01
Micro-opt property sets and refactor initialization

sorvell:
* pass property options in default setter to avoid the map lookup
* move property initialization from constructor to initialize to provide additional control over construction

### DIFF
--- a/src/lib/updating-element.ts
+++ b/src/lib/updating-element.ts
@@ -357,7 +357,7 @@ export abstract class UpdatingElement extends HTMLElement {
    * @nocollapse
    */
   protected static getPropertyDescriptor(
-      name: PropertyKey, key: string|symbol, _options: PropertyDeclaration) {
+      name: PropertyKey, key: string|symbol, options: PropertyDeclaration) {
     return {
       // tslint:disable-next-line:no-any no symbol in index
       get(): any {
@@ -367,7 +367,7 @@ export abstract class UpdatingElement extends HTMLElement {
         const oldValue =
             (this as {} as {[key: string]: unknown})[name as string];
         (this as {} as {[key: string]: unknown})[key as string] = value;
-        (this as unknown as UpdatingElement)._requestUpdate(name, oldValue);
+        (this as unknown as UpdatingElement)._requestUpdate(name, oldValue, options);
       },
       configurable: true,
       enumerable: true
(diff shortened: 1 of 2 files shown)


src/lib/updating-element.ts
```diff
@@ -698,8 +699,8 @@ export abstract class UpdatingElement extends HTMLElement {
    * @param oldValue {any} (optional) old value of requesting property
    * @returns {Promise} A Promise that is resolved when the update completes.
    */
-  requestUpdate(name?: PropertyKey, oldValue?: unknown) {
-    this._requestUpdate(name, oldValue);
+  requestUpdate(name?: PropertyKey, oldValue?: unknown, options?: PropertyDeclaration) {
```
kevinpschaaf, line 702: This change isn't _strictly_ necessary, right?

src/lib/updating-element.ts
```diff
@@ -653,12 +654,12 @@ export abstract class UpdatingElement extends HTMLElement {
...
    * not free to access.
    */
-  private _requestUpdate(name?: PropertyKey, oldValue?: unknown) {
+  private _requestUpdate(name?: PropertyKey, oldValue?: unknown, options?: PropertyDeclaration) {
     let shouldRequestUpdate = true;
     // If we have a property key, perform property update steps.
     if (name !== undefined) {
       const ctor = this.constructor as typeof UpdatingElement;
-      const options = ctor.getPropertyOptions(name);
+      options = options || ctor.getPropertyOptions(name);
```
justinfagnani, line 662: In a corner case, this is an observable change. It prevents the property options from being replaced without also overriding `_requestUpdate`.

src/lib/updating-element.ts
justinfagnani: Do we have cases where element authors would even have a reference to the options?

sorvell: Removed the change to `requestUpdate`.

src/lib/updating-element.ts
sorvell: There is no API for mutating property options so you'd have to hold onto and mutate the object yourself. I don't think it's reasonable to assume that was ever supported.

kevinpschaaf: There's already a note on `getPropertyOptions` that says
> Note, this method should be considered "final" and not overridden. To customize the options for a given property, override `createProperty`.
So this seems safe to me.

src/lib/updating-element.ts
kevinpschaaf: @justinfagnani The use case is overriding `getPropertyDescriptor` (which gets `options`) to return your own setter.  In that case you have `options`, and passing it to `requestUpdateInternal` eliminates the need to do the map lookup.

Issue #985, opened 2020-05-02
Change _attributeNameForProperty to protected


bschlenk:
### Description

[_attributeNameForProperty](https://github.com/Polymer/lit-element/blob/bcd500a6ebcd3c32d44b4b974dd4f2ed82aae62e/src/lib/updating-element.ts#L436) can be overridden, e.g. to support kebab-cased attributes. However, it requires a `// @ts-ignore` comment because the method is currently private. It should be `protected` to make it easier to override.

### Versions

- lit-element: v2.3.1

sorvell:
Custom attribute naming is already supported via specifying the `attribute` property option. Here's an example: https://stackblitz.com/edit/lit-element-demo-dtkn9t?file=index.html

We can consider this change, but would like to avoid it if the current support is sufficient.

bschlenk:
I’d like to be able to set that behavior once in a subclass of LitElement, instead of on every single property of every single component.

sorvell:
Sorry for the delay on this. We're opting not to make this change as it's possible to customize centrally via an override of `createProperty`. Here's an example: https://stackblitz.com/edit/lit-element-demo-3h2hdz?file=src%2Fmy-element.js

bschlenk:
👍 Thanks for the example!

PR #986, opened 2020-05-02
Make _attributeNameForProperty protected

bschlenk:
### Reference Issue

Fixes https://github.com/Polymer/lit-element/issues/985

### DIFF
--- a/docs/_guide/templates.md
+++ b/docs/_guide/templates.md
@@ -611,7 +611,7 @@ The `@query`, `@queryAll`, and `@queryAsync` decorators all provide a convenient
 
 </div>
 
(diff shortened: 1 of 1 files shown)


src/lib/updating-element.ts
```diff
@@ -433,7 +433,7 @@ export abstract class UpdatingElement extends HTMLElement {
    * Returns the property name for the given attribute `name`.
    * @nocollapse
    */
-  private static _attributeNameForProperty(
+  protected static _attributeNameForProperty(
```
sorvell, line 436: Please remove the leading underscore. Our convention is that this should only be used for private properties.

Issue #909, opened 2020-03-01
Document how to use render.js instead of shady-render.js


web-padawan:
## Initiative / goal

- there are developers who don't have to support IE11 nowadays
- these developers want to import `render.js` instead of `shady-render.js`

### Scope

- add a section in the docs describing how to achieve that:
  - via Babel plugin, e.g. [babel-plugin-replace-imports](https://github.com/041616/babel-plugin-replace-imports)
  - via webpack, e.g. [NormalModuleReplacementPlugin](https://webpack.js.org/plugins/normal-module-replacement-plugin/) seems an option

## Acceptance criteria 

- it should be easy to get rid of `shady-render.js` by copy-pasting few lines of code

bennypowers:
I'd like to see steps for rollup builds, as well.

georges-gomes:
via a "MODERN_TARGET" variable replaced by the bundler? so more things can be opt-out at the same time?

justinfagnani:
We should add this both to the Tools section as well as a link to those instructions from the Styling section.

CaptainCodeman:
For rollup, using `@plugin/rollup-alias', this shaves off a few bytes:

```js
    alias({
      entries: [{
        find: 'lit-html/lib/shady-render.js',
        replacement: 'node_modules/lit-html/lit-html.js'
      }]
    }),
```

georges-gomes:
I just tried it on the little counter [here](https://webcomponents.dev/blog/all-the-ways-to-make-a-web-component/) and I have no improvements on `lit-html`-only or `lit-element`.

my setup looks like this
```js
import commonjs from '@rollup/plugin-commonjs';
import resolve from '@rollup/plugin-node-resolve';
import replace from '@rollup/plugin-replace';
import { terser } from 'rollup-plugin-terser';
import visualizer from 'rollup-plugin-visualizer';
import alias from '@rollup/plugin-alias';
import * as fs from 'fs';
import * as path from 'path';

function b(src, withdeps) {
  return {
    input: `download/${src}/dist/index.js`,
    output: {
      dir: `dist/${src}/bundle` + (withdeps ? '' : '-nodeps'),
      format: 'esm',
      sourcemap: true,
    },
    plugins: [
      alias({
        entries: [
          {
            find: 'lit-html/lib/shady-render.js',
            replacement: 'node_modules/lit-html/lit-html.js',
          },
        ],
      }),
      replace({
        'process.env.NODE_ENV': JSON.stringify('production'),
[...]

PR #998, opened 2020-05-15
Add hydrate to lit-element

sorvell:
* does not make shadowRoot if one already exists
* detects need to hydrate based on shadowRoot existence at construction time
* first update calls hydrate instead of render if need for hydration was detected


web-padawan:
@sorvell what would be the expected impact of `hydrate` addition on the bundle size?

Another thing that I'm interested is whether it would be possible to factor this out to a mixin.
As a developer, I would like to have an option to easily override this by replacing an import, as I previously  requested for `shady-render` in #909, see possible solution at https://github.com/Polymer/lit-element/issues/909#issuecomment-619294064.

kevinpschaaf:
@web-padawan Please consider everything in the `hydration` branches very early implementations, nothing here is actual proposed API yet.

That said, it is definitely a goal to make SSR/hydration support opt-in and pay-for-play to the greatest extent possible.  The exact mechanics of how we achieve that (mixin, global configuration, build-time, etc. are still a bit TBD).

PR #1001, opened 2020-05-19
Consider npm ci rather than npm i in starter guide

rictic:
Compared to npm i, npm ci:

* is faster
* installs a version of dependencies that have been tested to work together (by whoever last updated package-lock.json)

Downsides:

* Does not get the latest version of deps, so may miss new features and bug fixes

IMO it's better to give people who're just getting started a known good state.

### DIFF
--- a/docs/_guide/start.md
+++ b/docs/_guide/start.md
@@ -55,7 +55,7 @@ The quickest way to try out a project locally is to download one of the starter 
 
     ```bash
     cd <project folder>
-    npm i
+    npm ci
     ```
 
 <div class="alert alert-info">


arthurevans:
I don't have much of a feeling on this one. @justinfagnani @sorvell ?

Based on what @rictic said, it sounds like `npm ci` is an improvement. My only objection would be that it's less familiar to users.

justinfagnani:
It's my understanding that `npm i` and `npm ci` are exactly equivalent for fresh installs. So this shouldn't have any changes in practice - in both cases users will get the versions from the package lock.

rictic:
Hm, you're right. What remains is that `npm ci` is faster, 13s vs 17s on my machine

justinfagnani:
I'm just not sure that's enough to throw in another command here. I don't see other packages recommending this. There's also the difference that if package.json and package-lock.json disagree on a dependency that `npm ci` fails. I hope we don't ever publish a package like that, but if we do I think the `npm i` behavior is more reasonable for our customers.

rictic:
If `npm i` is consistent, then I'm fine closing this

PR #1003, opened 2020-05-20
Fix queryAssignedNodes when unnamed slot is not first

### LINKED ISSUE #1002: queryAssignedNodes doesn't correctly locate default slot
## Description

Currently, if you use the `queryAssignedNodes` decorator with no argument, it retrieves the first `<slot>` in the template, whether or not it has a `name` attribute.

For example, given the code:

```
@queryAssignedNodes()
defaultNodes;

render() {
  return html`
     <slot name="header"></slot>
    <slot></slot>
    <slot name="footer"></slot>
}
```

The `defaultNodes` parameter returns the contents of the _first_ slot, not the second, as you'd expect.

justinfagnani:
Fixes #1002

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,9 +23,11 @@
   ],
   "scripts": {
     "build": "tsc && downlevel-dts . ts3.4 && cp tsconfig.json ./ts3.4/",
+    "build:watch": "tsc --watch",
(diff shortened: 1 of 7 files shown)


arthurevans:
Ya beat me to it :D

PR #1004, opened 2020-05-20
Add decorators to events, templates chapters.

### LINKED ISSUE #886: Documentation on accessing elements in the template
I know that shadowRoot is DOM standard but it may increase acceptance if you add some notes about accessing elements in the template to the documentation. I was converting a Polymer app to lit-element and got stuck at the following line:

`this.$.myDialog.open()`

It took me a while to find out that there is shadowRoot in the DOM which would allow me the following:

`this.shadowRoot.getElementById("myDialog").open();`

I found a line like this in an anti-pattern example in the templates chapter. But in this case I am not setting a property but calling a method of a Polymer paper-dialog. I hope this makes sense?

arthurevans:
Decorator docs part 2/2.

Follow-on to #999.
Fixes #886.

Staged as: https://decorator-docs-2-dot-polymer-lit-element.uk.r.appspot.com

(Note that this branch doesn't include the new decorators chapter--that's in #999--so the links to that chapter are broken in the staging link above.)

### DIFF
--- a/docs/_guide/events.md
+++ b/docs/_guide/events.md
@@ -29,7 +29,7 @@ You can use lit-html `@event` bindings in your template to add event listeners t
 
 ```js
 render() {
-  return html`<button @click="${this.handleClick}">`;
+  return html`<button @click="${this._handleClick}">`;
 }
 ```
 
@@ -46,7 +46,7 @@ The component constructor is a good place to add event listeners on the host ele
 ```js
 constructor() {
   super();
-  this.addEventListener('focus', this.handleFocus);
+  this.addEventListener('focus', this._handleFocus);
 }
 ```
 
@@ -65,10 +65,10 @@ If your component adds an event listener to anything except itself or its childr
 ```js
 connectedCallback() {
   super.connectedCallback();
-  window.addEventListener('resize', this.handleResize);
+  window.addEventListener('resize', this._handleResize);
 }
 disconnectedCallback() {
-  window.removeEventListener('resize', this.handleResize);
+  window.removeEventListener('resize', this._handleResize);
   super.disconnectedCallback();
 }
 ```
@@ -116,14 +116,26 @@ this.boundResizeHandler = this.handleResize.bind(this);
 window.addEventListener('resize', this.boundResizeHandler);
 ```
 
+Or use an arrow function as a class field:
+
+```ts
+export class MyElement extends LitElement {
+  private _handleResize = () => { /* handle the event */ }
+
+  constructor() {
+    window.addEventListener('resize', this._handleResize);
+  }
+}
+```
+
 See the [documentation for `this` on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this) for more information.
 
 ## Setting event listener options
 
-When you add an event listener imperatively, using `addEventListener`, you can specify various event listener options. For example, to use a capture phase event listener in plain JavaScript you'd do something like this:
+When you add an event listener imperatively, using `addEventListener`, you can specify various event listener options. For example, to use a passive event listener in plain JavaScript you'd do something like this:
 
 ```js
-someElement.addEventListener('click', this._handleClick, {capture: true});
+someElement.addEventListener('touchstart', this._handleTouchStart, {passive: true});
 ```
 
 The `eventOptions` decorator allows you to add event listener options to a listener that's added declaratively in your template.
@@ -132,19 +144,19 @@ The `eventOptions` decorator allows you to add event listener options to a liste
 import {LitElement, html, eventOptions} from 'lit-element';
 ...
 
-@eventOptions({capture: true})
-_handleClick_() { ... }
+@eventOptions({passive: true})
+private _handleTouchStart() { ... }
 
 render() { 
   return html`
-    <button @click=${this._handleClick}>Click me!</button>
+    <div @touchstart=${this._handleTouchStart}><div>
   `;
 }
 ```
 
 <div class="alert alert-info">
 
-**Using decorators.** Decorators are a proposed JavaScript feature, so you’ll need to use a compiler like Babel or the TypeScript compiler to use decorators. See Using decorators for details.
+**Using decorators.** Decorators are a proposed JavaScript feature, so you’ll need to use a compiler like Babel or TypeScript to use decorators. See [Using decorators](decorators) for details.
 
 </div>
 
--- a/docs/_guide/events.md
+++ b/docs/_guide/events.md
@@ -8,96 +8,156 @@ slug: events
 * ToC
 {:toc}
 
-## Overview
 
-### Where to add your event listeners
+## Where to add your event listeners
 
 You need to add event listeners in a method that is guaranteed to fire before the event occurs. However, for optimal loading performance, you should add your event listener as late as possible.  
 
-You can add event listeners:
+The most common ways to add an event listener:
 
-*   **Via your component's template.**
+* Declaratively, in  your component's template
+* In the component constructor, for listeners added on the component itself.
+* In the `connectedCallback`, for listeners that need to reference DOM outside the component (for example, `Window` or `Document`).
+* After first paint, if you're adding a lot of listeners and first paint performance is critical.
 
-    You can use lit-html `@event` bindings in your template inside the `render` function to add event listeners to your component. 
 
-    **Example**
+### Add declarative event listeners
 
-    ```js
-    render() {
-      return html`<button @click="${this.handleClick}">`;
-    }
-    ```
+You can use lit-html `@event` bindings in your template to add event listeners to your component. 
 
-*   **In the component constructor.**
+**Example**
 
-    If you need to listen for an event that might occur before your component has been added to DOM, you might need to add the event listener in your component's constructor. 
+```js
+render() {
+  return html`<button @click="${this.handleClick}">`;
+}
+```
 
-    **Example**
+Declarative event listeners are added when the template is rendered. This is usually the best way to add listeners to elements in your templated DOM.
 
-    ```js
-    constructor() {
-      super();
-      this.addEventListener('DOMContentLoaded', this.handleLoaded);
-    }
-    ```
+### Add event listeners in the constructor
 
-*   **In `firstUpdated`**.
+If you need to listen for an event that might occur before your component has been added to DOM, you might need to add the event listener in your component's constructor. 
 
-    `firstUpdated` is a LitElement lifecycle callback. `firstUpdated` fires after the first time your component has been updated and rendered. See [firstUpdated](/guide/lifecycle#firstupdated) in the Lifecycle documentation for more information.
+The component constructor is a good place to add event listeners on the host element itself.
 
-    If the event you're handling can't occur before your component has been updated and rendered for the first time, it's safe and efficient to add a listener for it in `firstUpdated`. 
+**Example**
 
-    **Example**
+```js
+constructor() {
+  super();
+  this.addEventListener('focus', this.handleFocus);
+}
+```
 
-    ```js
-    firstUpdated(changedProperties) {
-      this.addEventListener('click', this.handleClick);
-    }
-    ```
+### Add event listners in `connectedCallback`
 
-*   **In `connectedCallback`**. 
+`connectedCallback` is a lifecycle callback in the custom elements API. `connectedCallback` fires each time a custom element is appended into a document-connected element. See [the MDN documentation on using custom elements lifecycle callbacks](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements#Using_the_lifecycle_callbacks) for more information.
 
-    `connectedCallback` is a lifecycle callback in the custom elements API. `connectedCallback` fires each time a custom element is appended into a document-connected element. See [the MDN documentation on using custom elements lifecycle callbacks](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_custom_elements#Using_the_lifecycle_callbacks) for more information.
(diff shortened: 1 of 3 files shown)


arthurevans:
Updated staging with the latest commits. PTAL.

arthurevans:
@justinfagnani @sorvell could one of you take a look at this one?

docs/_guide/events.md
```diff
@@ -8,96 +8,156 @@ slug: events
...
+render() { 
+  return html`
+    <button @click=${this._handleClick}>Click me!</button>
+  `;
+}
+```
+
+<div class="alert alert-info">
+
+**Using decorators.** Decorators are a proposed JavaScript feature, so you’ll need to use a compiler like Babel or the TypeScript compiler to use decorators. See Using decorators for details.
```
justinfagnani, line 147: make "Using decorators" a link

justinfagnani, line 147: maybe?

```suggestion
**Using decorators.** Decorators are a proposed JavaScript feature, so you’ll need to use a compiler like Babel or TypeScript to use decorators. See Using decorators for details.
```

docs/_guide/events.md
```diff
@@ -8,96 +8,156 @@ slug: events
...
-    }
-    ```
+See [firstUpdated](/guide/lifecycle#firstupdated) in the Lifecycle documentation for more information.
+
 
-### Using `this` in event handlers
+## Using `this` in event listeners
 
-The default JavaScript context object (`this`) inside an event handler belonging to a LitElement-based component is the component itself. 
+Event listeners added using the declarative (`@event`) syntax in the template are automatically _bound_ to the component.
```
justinfagnani, line 97: You should probably keep the "bound" verbiage here, but technically we don't bind the function, we just call it with the right `this` value. The difference is that we don't create a new bound function for each event binding * element instance, but just reuse the prototype type method. It's only a thing because people often criticize React for creating so many closures.

docs/_guide/events.md
```diff
@@ -8,96 +8,156 @@ slug: events
...
+  _handleClick(e) {
     console.log(this.prop);
   }
 }
 ```
 
+When adding listeners imperatively with `addEventListener`, you'll need to bind the event listener yourself if you need a reference to the component instance. For example:
+
+```js
+this.boundResizeHandler = this.handleResize.bind(this);
```
justinfagnani, line 115: I think the most ergonomic way to do this is like:

```ts
  private _onResize = (e) => { ... };

  constructor() {
    super();
    window.addEventListener('resize', this._onResize);
  }
}
```

The only downside here is you can't call `super._onResize()` in the handler, but that's a rare thing to need to do.

docs/_guide/events.md
```diff
@@ -8,96 +8,156 @@ slug: events
...
+
 See the [documentation for `this` on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this) for more information.
 
-### Use cases
+## Setting event listener options
+
+When you add an event listener imperatively, using `addEventListener`, you can specify various event listener options. For example, to use a capture phase event listener in plain JavaScript you'd do something like this:
+
+```js
+someElement.addEventListener('click', this._handleClick, {capture: true});
```
justinfagnani, line 126: I would suggest using a `'scroll'` event and the `passive: true` option just 'cause more people should know about it (it eliminates jank from scroll listeners).

docs/_guide/templates.md
```diff
@@ -375,21 +386,22 @@ You can compose LitElement templates from other LitElement templates. In the fol
 
 ```js
 class MyPage extends LitElement {
+  ...
   render() {
     return html`
-      ${this.headerTemplate}
-      ${this.articleTemplate}
+      ${this.headerTemplate(this.article.title)}
```
justinfagnani, line 392: Instead of passing an argument I would just access `this.article.title` in the headerTemplate. Or keep the argument and make header template standalone function to show that they don't need to be class methods.

docs/_guide/templates.md
```diff
@@ -569,6 +581,162 @@ html`${until(content, html`<span>Loading...</span>`)}`
 
 For a list of directives supplied with lit-html, see [Built-in directives](https://lit-html.polymer-project.org/guide/template-reference#built-in-directives) in the Template syntax reference.
 
+## Accessing nodes in the shadow DOM
+
+Templated DOM is usually rendered into shadow DOM, so the nodes are not direct children of the component. Use `this.shadowRoot.querySelector()` or `this.shadowRoot.querySelectorAll()` to find nodes in the
```
justinfagnani, line 586: maybe?

```suggestion
The `render()` method result is usually rendered into shadow DOM, so the nodes are not direct children of the component. Use `this.shadowRoot.querySelector()` or `this.shadowRoot.querySelectorAll()` to find nodes in the 
```

only because "templated DOM" seems to be a phrase of our own creation. I might change "templated DOM" below to "rendered DOM" too.

docs/_guide/templates.md
```diff
@@ -569,6 +581,162 @@ html`${until(content, html`<span>Loading...</span>`)}`
...
+
+**Using decorators.** Decorators are a proposed JavaScript feature, so you’ll need to use a compiler like Babel or the TypeScript compiler to use decorators. See [Using decorators](decorators) for details.
+
+</div>
+
+The `@query` decorator modifies a class property, turning it into a getter that returns a node from the render root.
+
+```js
+class MyElement {
+  @query('#first')
```
justinfagnani, line 618: I would suggest making this a TypeScript example, making the field private, and naming it with a `_` prefix, because the queried elements are internal implementation details and `@query` is really meant to internal usage - we don't want authors accidentally making private things public.

docs/_guide/templates.md
```diff
@@ -569,6 +581,162 @@ html`${until(content, html`<span>Loading...</span>`)}`
...
 
+## Accessing nodes in the shadow DOM
+
+Templated DOM is usually rendered into shadow DOM, so the nodes are not direct children of the component. Use `this.shadowRoot.querySelector()` or `this.shadowRoot.querySelectorAll()` to find nodes in the 
+shadow DOM.
+
+You can query the templated DOM after its initial render (for example, in `firstUpdated`), or use a getter pattern, like this:
+
+```js
+get closeButton() {
```
justinfagnani, line 592: Same as the comment on the `@query` example below, I would emphasize that this should probably be "private" unless you really want to expose the element publicly. So name it with a `_` prefix.

docs/_guide/templates.md
```diff
@@ -569,6 +581,162 @@ html`${until(content, html`<span>Loading...</span>`)}`
...
+```
+@queryAll('button')
+buttons!: NodeListOf<HTMLButtonElement>
+```
+
+The exclamation point (`!`) after `buttons` is TypeScript's [non-null assertion operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator). It tells the compiler to treat `buttons` as always being defined, never `null` or `undefined`.
+
+Finally, `@queryAsync` works like `@query`, except that instead of returning a node directly, it returns a `Promise` that resolves to that node. External code can use this instead of waiting for the `updateComplete` promise. 
+
+This is useful, for example, if the node returned by `@queryAsync` can change as a result of another property change.
```
justinfagnani, line 672: Maybe @sorvell can add some insight on documenting `@queryAsync` externally. It's most for internal tool usage, and I'm a little worried that it seems like we're encouraging exposing shadow DOM contents publicly here. I wonder how we can describe this well, or if we should just ignore `@queryAsync` in the main docs.

docs/_guide/templates.md
```diff
@@ -569,6 +581,162 @@ html`${until(content, html`<span>Loading...</span>`)}`
...
+*   [HTMLSlotElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSlotElement) on MDN.
+
+
+### @queryAssignedNodes decorator
+
+The `@queryAssignedNodes` decorator converts a class property into a getter that returns all of the assigned nodes for a given slot in the component's shadow tree. 
+
+<div class="alert alert-info">
+
+**Using decorators.** Decorators are a proposed JavaScript feature, so you’ll need to use a compiler like Babel or the TypeScript compiler to use decorators. See [Using decorators](decorators) for details.
```
justinfagnani, line 715: Is there a way to define this snippet in a common location and include it?

docs/_guide/templates.md
```diff
@@ -569,6 +581,162 @@ html`${until(content, html`<span>Loading...</span>`)}`
...
+
+</div>
+
+```js
+// First argument is the slot name
+// Second argument is `true` to flatten the assigned nodes.
+@queryAssignedNodes('header', true)
+headerNodes;
```
justinfagnani, line 723: Maybe make these private too, though it's a little less of a concern because child nodes are already public external to the element. I just want to reinforce that class members defined only to help with the component implementation should be private by default.

docs/_includes/projects/docs/templates/compose/my-page.js
```diff
@@ -1,21 +1,36 @@
...
+      text: 'Some witty text.',
+    };
+  }
+
   render() {
     return html`
-      ${this.headerTemplate}
-      ${this.articleTemplate}
-      ${this.footerTemplate}
+      ${this.headerTemplate(this.article.title)}
```
justinfagnani, line 22: Same comment with the argument as above.

docs/_guide/events.md
arthurevans: My understanding was that `scroll` itself can't be canceled, so it doesn't need `passive`? I think you'd use it on `touchstart`, `touchend`, etc... https://developers.google.com/web/updates/2016/06/passive-event-listeners

docs/_guide/templates.md
arthurevans: Urgh. I was trying to show functions here because someone was confused by the getters. 

Are standalone methods too confusing? I like showing that they don't need to be class methods, so I'll go with that.

docs/_guide/events.md
arthurevans: Done.

arthurevans: Updated this in all the places it appears.

docs/_guide/templates.md
arthurevans: Possibly? I think we'll only end up with a total of 4 or 5 instances of it in the doc set, though.

docs/_guide/templates.md
arthurevans: Sure, SG.

docs/_guide/events.md
arthurevans: There is also a certain weird tax. I've seen this before and I still had to read a half-dozen pages of doc to remind myself why it works.

Also: assumes class fields support. Should I include both examples, perhaps?

docs/_guide/events.md
arthurevans: OK, so just for the record--I'm _not_ changing any doc here, right?

docs/_guide/templates.md
arthurevans: It *is* a phrase of our own creation. We already use it a couple of times at the beginning of the chapter.

In the lit-html docs, we actually just say "templates"--like, "Styling templates"--when we really mean, "styling the DOM you create by rendering templates." If you think "rendered DOM" is clearer and less invented-y, we could rephrase this to, say, "Rendered DOM is usually placed in an open shadow root," or "The component's template is usually rendered into shadow DOM" and use "rendered DOM" below as you suggested.

Added this to today's docs meeting agenda if we need to discuss it further.

docs/_guide/templates.md
arthurevans: So named.

docs/_guide/templates.md
arthurevans: At the moment, we don't have anything to show it's a Typescript example, so I've been trying to limit Typescript-specific syntax to examples that are explicitly labeled Typescript (one more reason I'm dying to get a working code-switcher up and running).

How about I just use the `_` and add a note about how these properties should typically be protected or private (`_` convention or Typescript's `private` or `protected` keywords).

docs/_guide/templates.md
arthurevans: By "internal tool usage" do you mean for testing? I'm not actually too clear on the use cases for this decorator. 

Help me @sorvell you're my only hope!

docs/_includes/projects/docs/templates/compose/my-page.js
arthurevans: Fixed, thanks!

docs/_guide/events.md
arthurevans: Used "touchstart" with passive.

docs/_guide/events.md
arthurevans: Included both.

docs/_guide/templates.md
sorvell: Sorry, that comment was unclear.

I was imagining the use case where a dev decides to expose a public property via `@query('.foo') fooElement`. In order for that to work correctly at all times, the user would have to `await element.updateComplete`:

```
await element.updateComplete;
element.fooElement.textContent = 'hi';
```
Instead the dev may choose to expose the property with `@queryAsync('.foo') fooElement;` so the user code can be more compact:

```
(await element.fooElement).textContent = 'hi';
```

arthurevans: OK, understood. I think the concern @justinfagnani had was that these properties that expose shadow DOM children should probably be internal so we don't leak implementation details.

That said, a property getter provides more controlled access than having the user code call `element.shadowRoot.querySelector()` themselves. So...

arthurevans: For now, just leaving this in with the word "external" removed.

PR #1010, opened 2020-06-01
Improve accuracy of property decorator

### LINKED ISSUE #1009: Accurate typings for property and internalProperty decorators
## Description

Recently, a new `internalProperty` decorator was added with the following intent.
```js
 * This (property) decorator should only be used for public fields. Private or protected
 * fields should use the [[`internalProperty`]] decorator.
```

The problem is that the typings of both `property` and `internalProperty` decorators are very loose and match any property, be it public, protected or private.

I propose to make the typings more accurate so that `property` can only be applied to public properties and `internalProperty` can only be applied to protected or private properties.

## Acceptance criteria

The typings for the `property` decorators are easy to implement, since `keyof` will provide all public properties of a class. Currently, there is no way to obtain the protected and private properties.  However if [typescript merged support for negated types](https://github.com/microsoft/TypeScript/pull/29317), we would be able to workaround this limitation and provide accurate typings for `internalProperty`.

I propose to merge first the typings for `property` decorator, and postpone `internalProperty` typings until it becomes feasible.

javier-garcia-meteologica:
### Reference Issue
Partly fixes #1009

Makes sure that `property` decorator is applied to public properties, and maintains the old signature for `internalProperty` until more accurate typings become possible.

### DIFF
--- a/src/lib/decorators.ts
+++ b/src/lib/decorators.ts
@@ -160,7 +160,7 @@ const legacyProperty =
  */
 export function property(options?: PropertyDeclaration) {
   // tslint:disable-next-line:no-any decorator
-  return (protoOrDescriptor: Object|ClassElement, name?: PropertyKey): any =>
+  return <T extends Object|ClassElement>(protoOrDescriptor: T, name?: keyof T): any =>
              (name !== undefined) ?
       legacyProperty(options!, protoOrDescriptor as Object, name) :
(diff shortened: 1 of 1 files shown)


sorvell:
Thanks for the PR, as noted [here](https://github.com/Polymer/lit-element/issues/1009#issuecomment-639046404) we've elected not to change this now since this would be breaking, but we'll consider it for the next major version.

PR #1031, opened 2020-07-06
Implements _childrenPendingHydration queue to coordinate parent/child hydration

kevinpschaaf:
Fixes an issue where children will typically be registered and thus boot up/hydrate bottom-up, before having received its bound hydration state from its parent:
* All child custom elements rendered into a shadow root will have a `defer-hydration` attribute applied
* When an element upgrades, if the `defer-hydration` attribute is applied it will refrain from running its UpdatingElement:connectedCallback() implementation, keeping the element in a non-enabled state and deferring hydration
* After a parent hydrates, it queries for `[defer-hydration]` children in its shadow root and removes those attributes
* When the `defer-hydration` attribute is removed, a child will resume running its `UpdatingElement:connectedCallback()`, thus enabling itself, running its hydration step, and causing the process to recurse.

This change also includes some minimal code necessary for the `element-as-renderer` change in `lit-ssr`:
* Adds a `noUpdatedCallbacks` argument to `performUpdate`, used when running update synchronously during SSR
* Adds the (optional) `ssrRenderChildren` interface to its definition; this will be assigned by `lit-element-renderer`

### DIFF
--- a/src/lib/updating-element.ts
+++ b/src/lib/updating-element.ts
@@ -755,7 +755,7 @@ export abstract class UpdatingElement extends HTMLElement {
    * }
    * ```
    */
-  protected performUpdate(): void|Promise<unknown> {
+  performUpdate(noUpdatedCallbacks?: boolean): void|Promise<unknown> {
     // Abort any update if one is not pending when this is called.
     // This can happen if `performUpdate` is called early to "flush"
     // the update.
@@ -783,7 +783,7 @@ export abstract class UpdatingElement extends HTMLElement {
       this._markUpdated();
       throw e;
     }
-    if (shouldUpdate) {
+    if (shouldUpdate && !noUpdatedCallbacks) {
       if (!(this._updateState & STATE_HAS_UPDATED)) {
         this._updateState = this._updateState | STATE_HAS_UPDATED;
         this.firstUpdated(changedProperties);
--- a/src/lib/updating-element.ts
+++ b/src/lib/updating-element.ts
@@ -754,8 +754,11 @@ export abstract class UpdatingElement extends HTMLElement {
    *   super.performUpdate();
    * }
    * ```
+   * 
+   * @param runUpdatedCallbacks When `false`, `firstUpdated` and `updated`
+   *   callbacks are not run (defaults to `true`).
    */
-  performUpdate(noUpdatedCallbacks?: boolean): void|Promise<unknown> {
+  protected performUpdate(runUpdatedCallbacks = true): void|Promise<unknown> {
     // Abort any update if one is not pending when this is called.
     // This can happen if `performUpdate` is called early to "flush"
     // the update.
@@ -783,7 +786,7 @@ export abstract class UpdatingElement extends HTMLElement {
       this._markUpdated();
       throw e;
     }
-    if (shouldUpdate && !noUpdatedCallbacks) {
+    if (shouldUpdate && runUpdatedCallbacks) {
       if (!(this._updateState & STATE_HAS_UPDATED)) {
         this._updateState = this._updateState | STATE_HAS_UPDATED;
         this.firstUpdated(changedProperties);
--- a/src/lit-element.ts
+++ b/src/lit-element.ts
@@ -336,7 +336,7 @@ export class LitElement extends UpdatingElement {
         (this.constructor as typeof LitElement)
             .hydrate(templateResult, this.renderRoot);
         // Hydrate any children waiting on parent hydration
-        this.renderRoot.querySelectorAll('[defer-hydration]').forEach(el => {
+        this.renderRoot.querySelectorAll('[defer-hydration]').forEach((el) => {
           el.removeAttribute('defer-hydration');
         });
       } else {
--- a/src/lit-element.ts
+++ b/src/lit-element.ts
@@ -126,7 +126,7 @@ export class LitElement extends UpdatingElement {
       (result: unknown, container: Element|DocumentFragment,
        options: ShadyRenderOptions) => void = render;
 
-  ssrRenderChildren: IterableIterator<string> | undefined;
+  ssrRenderChildren: IterableIterator<string>|undefined;
 
   /**
    * LitElement detects if it should try to `hydrate` based on whether or not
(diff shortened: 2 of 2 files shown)


src/lib/updating-element.ts
```diff
@@ -755,7 +755,7 @@ export abstract class UpdatingElement extends HTMLElement {
    * }
    * ```
    */
-  protected performUpdate(): void|Promise<unknown> {
+  performUpdate(noUpdatedCallbacks?: boolean): void|Promise<unknown> {
```
justinfagnani, line 758: Can you add docs for the param?

justinfagnani, line 758: Maybe consider a default value so you can invert value and the if-check below:

```ts
performUpdate(updatedCallbacks: boolean = true): void|Promise<unknown> {
```

src/lit-element.ts
```diff
@@ -126,6 +126,8 @@ export class LitElement extends UpdatingElement {
       (result: unknown, container: Element|DocumentFragment,
        options: ShadyRenderOptions) => void = render;
 
+  ssrRenderChildren: IterableIterator<string> | undefined;
```
justinfagnani: I have to review the other PRs, but docs would be good here and this is unused locally, so what's it do?

src/lit-element.ts
```diff
@@ -137,6 +139,18 @@ export class LitElement extends UpdatingElement {
   static hydrate:
       (result: unknown, container: Element|DocumentFragment) => void;
 
+  static get observedAttributes() {
+    return [...super.observedAttributes, 'defer-hydration'];
+  }
+
+  attributeChangedCallback(name: string, old: string | null, value: string | null) {
```
justinfagnani: I wonder if this should be in UpdatingElement so that it's a good base class for SSR-compatible components, even for other template syntaxes?

src/lit-element.ts
```diff
@@ -288,11 +302,15 @@ export class LitElement extends UpdatingElement {
...
   connectedCallback() {
-    super.connectedCallback();
-    // Note, first update/render handles styleElement so we only call this if
-    // connected after first update.
-    if (this.hasUpdated && window.ShadyCSS !== undefined) {
-      window.ShadyCSS.styleElement(this);
+    if (this.hasAttribute('defer-hydration') && this._needsHydration) {
+      // Wait until parent hydrates
+    } else {
+      super.connectedCallback();
```
justinfagnani, line 308: I feel a little strange about deferring a super call. I wonder if this initialization logic can move into a method called conditionally by connectedCallback. ie rename enableUpdates to something like bootInstance and call that from connectedCallback and attributeChangedCallback. Super small nit though...

src/lit-element.ts
```diff
@@ -317,6 +335,10 @@ export class LitElement extends UpdatingElement {
         this._needsHydration = false;
         (this.constructor as typeof LitElement)
             .hydrate(templateResult, this.renderRoot);
+        // Hydrate any children waiting on parent hydration
```
justinfagnani, line 338: I think we may want a hydrate method to hold this soon

src/lit-element.ts
kevinpschaaf: All `super.connectedCallback()` does in UpdatingElement is call `this.enableUpdating()`.  I made the change to just call that method directly rather than `super.connectedCallback()`. That maybe looks less weird, but not sure if it's an improvement since the implementation is now disconnected from `UpdatingElement:connectedCallback`.

src/lib/updating-element.ts
kevinpschaaf: Done.

kevinpschaaf: Done

sorvell:
I have perf concerns over this RP due to:

1. the need to serialize/download lots of `defer-hydration` attributes
2. the need to querySelectorAll for `defer-hydration` attributes in every `shadowRoot`
3. the need to remove all `defer-hydration` attributes in every `shadowRoot`

Here's an alternative idea.

1. In `connectedCallback`, if the element `needsHydration` and its `getRootNode()` is a `shadowRoot`, assume it *might* be hydrated by its parent and 
   1.  Add it to a global set of `pendingHydration` elements. This is a backstop so that it can be enabled without being provoked by its host.
   2. Set a flag `canEnableUpdate = true`. This puts the element into a state that allows any property set to provoke an update.
2. At the top of `_requestUpdate`,  if `canEnableUpdate == true`
   1. Call `enableUpdating()`. This property set, presumably by a hydrating host will actually cause an update
   2. Set `canEnableUpdate = false` to reset state.
   3. Remove the element from the `pendingHydration` set because we don't need to do anything else.
3. At the end of hydration, for all elements in the `pendingHydration` set, call `enableUpdating()` and `performUpdate()`. This ensures any elements that were not provoked by their parents are fully enabled. Note, we know the order of updating of these elements does not matter.

src/lit-element.ts
```diff
@@ -317,6 +344,10 @@ export class LitElement extends UpdatingElement {
         this._needsHydration = false;
         (this.constructor as typeof LitElement)
             .hydrate(templateResult, this.renderRoot);
+        // Enable any children waiting on parent hydration
+        if (this._childrenPendingHydration) {
+          this._childrenPendingHydration.forEach(el => el.enableUpdating());
```
sorvell, line 349: Clear this list.

src/lit-element.ts
```diff
@@ -288,7 +294,28 @@ export class LitElement extends UpdatingElement {
...
+    // If element is pending hydration, in a shadow root, and it's host has not
+    // yet updated, add self to host's `_childrenPendingHydration` list and wait
+    // for host to enable
+    if (this._needsHydration) {
+      const root = this.getRootNode();
+      if (root instanceof ShadowRoot) {
+        // This assumes hydration host is always another LitElement (or
+        // something) implementing the `_childrenPendingHydration` protocol
+        const host = root.host as LitElement;
+        if (!host.hasUpdated) {
```
sorvell, line 307: `hasUpdated` is protected, right? Let's just change that to public.

kevinpschaaf: Why? This isn't an access violation since it's from a subclass (even if it's on a different instance).

src/lit-element.ts
kevinpschaaf: Good catch. Done!

justinfagnani:
This is superseded by LitElement 3.0

PR #1032, opened 2020-07-08
Adds decorator options

sorvell:
* Adds a `cache: boolean` argument to the `@query` decorator as a performance optimization for properties whose queried element is not expected to change.

* Adds a `selector: string` argument to the `@queryAssignedNodes` decorator as a convenience to filter the assigned nodes by the given selector.


src/lib/decorators.ts
```diff
@@ -411,16 +419,21 @@ export function eventOptions(options: AddEventListenerOptions) {
  * @category Decorator
  */
 export function queryAssignedNodes(
-    slotName: string = '', flatten: boolean = false) {
+    slotName = '', flatten = false, selector = '') {
```
kevinpschaaf, line 422: Just note to future selves, we should make decorators take option bags in the future for optional arguments. Too late to change now.

src/lib/decorators.ts
```diff
@@ -224,6 +224,14 @@ export function query(selector: string) {
       enumerable: true,
       configurable: true,
     };
+    if (cache) {
+      const key = typeof name === 'symbol' ? Symbol() : `__${name}`;
+      descriptor.get = function(this: LitElement) {
+        return (this as unknown as {[key: string]: Element|null})[key as string] ||
```
kevinpschaaf, line 230: This will re-query every time if the query returns `null`... I suppose that's probably okay, but should it be documented behavior?

src/lib/decorators.ts
```diff
@@ -224,6 +224,14 @@ export function query(selector: string) {
       enumerable: true,
       configurable: true,
     };
+    if (cache) {
+      const key = typeof name === 'symbol' ? Symbol() : `__${name}`;
+      descriptor.get = function(this: LitElement) {
```
justinfagnani, line 229: I think this PR should update the docs, and mention that `null` isn't cached.

And are you sure you want `null` not cached? What's the use case vs not caching?

src/lib/decorators.ts
```diff
@@ -411,16 +419,21 @@ export function eventOptions(options: AddEventListenerOptions) {
...
       get(this: LitElement) {
-        const selector =
+        const slotSelector =
             `slot${slotName ? `[name=${slotName}]` : ':not([name])'}`;
-        const slot = this.renderRoot.querySelector(selector);
-        return slot && (slot as HTMLSlotElement).assignedNodes({flatten});
+        const slot = this.renderRoot.querySelector(slotSelector);
+        let nodes = slot && (slot as HTMLSlotElement).assignedNodes({flatten});
+        if (nodes && selector) {
+          nodes = nodes.filter((node) => node instanceof Element &&
```
justinfagnani, line 433: I'd replace `instanceof` with `node.nodeType === 1`

docs/api/classes/_lit_element_.cssresult.html
```diff
@@ -153,7 +153,7 @@ <h3>constructor</h3>
 						<li class="tsd-description">
 							<aside class="tsd-sources">
 								<ul>
-									<li>Defined in <a href="https://github.com/Polymer/lit-element/blob/557de8c/src/lib/css-tag.ts#L25">src/lib/css-tag.ts:25</a></li>
+									<li>Defined in <a href="https://github.com/PolymerLabs/polymer-lit/blob/3e18aa4/src/lib/css-tag.ts#L25">src/lib/css-tag.ts:25</a></li>
```
kevinpschaaf, line 156: Weird that these changed to an old URL...

justinfagnani: Oh, weird. I noticed the SHA change but not the repo change. @sorvell are you working from an old git client with an old origin?

kevinpschaaf: Ah, the docs generator uses the git origin to make the links? Makes sense.

@sorvell You can do `git remote -v` to check.

Probably just need to do this and re-generate:
```
git remote set-url origin https://github.com/Polymer/lit-element.git
```

src/lib/decorators.ts
sorvell: Updated docs. The value `null` is now cached. The first value is always cached.

src/lib/decorators.ts

docs/api/classes/_lit_element_.cssresult.html
sorvell: Good catch. Fixed.

src/lib/decorators.ts
sorvell: Changed behavior to always cache first result.

PR #1060, opened 2020-08-14
Prep 2.4.0-pre.3

justinfagnani:
This includes type: module and removes the browserstack badge


justinfagnani:
@aomarks indeed. Good catch!

PR #1079, opened 2020-09-21
Update properties.md, fix typo.

heibor:
Fixed unequal sign typo.


arthurevans:
We don't have a CLA on this from @heibor so I think we need to close this. I'll check for this typo in the new site.

PR #1086, opened 2020-10-08
Add type assertion to JSON.parse().

tylerhou:
This makes this repo compliant with Bazel's TypeScript rules: https://github.com/bazelbuild/rules_typescript/blob/master/internal/tsetse/rules/must_type_assert_json_parse_rule.ts

### DIFF
--- a/src/lib/updating-element.ts
+++ b/src/lib/updating-element.ts
@@ -165,7 +165,8 @@ export const defaultConverter: ComplexAttributeConverter = {
         return value === null ? null : Number(value);
       case Object:
       case Array:
-        return JSON.parse(value!);
+        // Type assert to adhere to Bazel's "must type assert JSON parse" rule.
+        return JSON.parse(value!) as unknown;
     }
     return value;
   }


src/lib/updating-element.ts
```diff
@@ -165,7 +165,7 @@ export const defaultConverter: ComplexAttributeConverter = {
         return value === null ? null : Number(value);
       case Object:
       case Array:
-        return JSON.parse(value!);
+        return JSON.parse(value!) as unknown;
```
justinfagnani, line 168: Can you add a comment here so we don't loose this cast accidentally?

justinfagnani (changes requested):
Interesting. Thanks for the PR!

Would you mind also doing this to the LitElement 3.x code that lives in the lit-html monorepo? https://github.com/Polymer/lit-html/tree/lit-next/packages/lit-element

src/lib/updating-element.ts
tylerhou: Done.

tylerhou (commented):
> Would you mind also doing this to the LitElement 3.x code that lives in the lit-html monorepo? https://github.com/Polymer/lit-html/tree/lit-next/packages/lit-element

Done. https://github.com/Polymer/lit-html/pull/1358

justinfagnani (approved):
Thanks!

PR #1089, opened 2020-10-11
Fix "not not" typo

stephenwade:
This fixes a typo I noticed in the documentation.


arthurevans (approved):
LGTM. Thanks for the fix!

PR #1092, opened 2020-10-16
Fixes queryAssignedNodes when querying for nodes on browsers that do …

### LINKED ISSUE #1088: TypeError: Illegal invocation error is thrown when using  queryAssignedNodes with text nodes
### Description

queryAssignedNodes throws A "TypeError: Illegal invocation" when slotted content contains text nodes.

#### Live Demo
https://webcomponents.dev/edit/FXrFMojSYzYa8eJNaJzh

#### Steps to Reproduce

1. Create `my-element` with child nodes containing text.

#### Expected Results
No error is throw

#### Actual Results
TypeError: Illegal invocation is thrown

### Browsers Affected (Haven't tested in other browsers)
- [x ] Chrome
- [ ] Firefox
- [ ] Edge
- [ ] Safari 11
- [ ] Safari 10
- [ ] IE 11

### Versions
Uses the default/current version in https://webcomponents.dev

sorvell:
…have Element.matches

Fixes #1088.

### DIFF
--- a/src/lib/decorators.ts
+++ b/src/lib/decorators.ts
@@ -465,6 +465,7 @@ export function queryAssignedNodes(
         if (nodes && selector) {
           nodes = nodes.filter(
               (node) => node.nodeType === Node.ELEMENT_NODE &&
+                      // tslint:disable-next-line:no-any testing existence on older browsers
                       ((node as any).matches ?
                   (node as Element).matches(selector) :
                   legacyMatches.call(node as Element, selector)));
--- a/src/lib/decorators.ts
+++ b/src/lib/decorators.ts
@@ -465,9 +465,9 @@ export function queryAssignedNodes(
         if (nodes && selector) {
           nodes = nodes.filter(
               (node) => node.nodeType === Node.ELEMENT_NODE &&
-                      (node as Element).matches ?
+                      ((node as any).matches ?
                   (node as Element).matches(selector) :
-                  legacyMatches.call(node as Element, selector));
+                  legacyMatches.call(node as Element, selector)));
         }
         return nodes;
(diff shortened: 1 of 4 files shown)


src/lib/decorators.ts
```diff
@@ -465,9 +465,10 @@ export function queryAssignedNodes(
         if (nodes && selector) {
           nodes = nodes.filter(
               (node) => node.nodeType === Node.ELEMENT_NODE &&
-                      (node as Element).matches ?
+                      // tslint:disable-next-line:no-any testing existence on older browsers
+                      ((node as any).matches ?
```
justinfagnani, line 469: Since the DOM types include `.matches` you shouldn't need the `as any` cast. Did the old code not type-check?

justinfagnani (approved):

kevinpschaaf (approved):

src/lib/decorators.ts
```diff
@@ -465,9 +465,10 @@ export function queryAssignedNodes(
         if (nodes && selector) {
           nodes = nodes.filter(
               (node) => node.nodeType === Node.ELEMENT_NODE &&
-                      (node as Element).matches ?
-                  (node as Element).matches(selector) :
-                  legacyMatches.call(node as Element, selector));
+                  // tslint:disable-next-line:no-any testing existence on older browsers
+                  ((node as any).matches ?
```
justinfagnani, line 469: Still curious about this cast here. Why is it necessary?

sorvell: TypeScript complains that the check is unnecessary because`matches` always exists. The previous code probably didn't analyze path.

PR #1093, opened 2020-10-21
Fix @query decorator cache for modern mode

bramkragten:
Right now the `name` is used for the key for the cache of the element.

But `name` is `undefined` for modern decorators, meaning all elements will be saved as `__undefined` and multiple `@query` decorators with `cache` all return the same element of the first one used.

This uses the `protoOrDescriptor.key` when `name` is `undefined`.

### DIFF
--- a/src/lib/decorators.ts
+++ b/src/lib/decorators.ts
@@ -227,15 +227,16 @@ export function query(selector: string, cache?: boolean) {
       configurable: true,
     };
     if (cache) {
-      const key = typeof name === 'symbol' ? Symbol() : `__${name}`;
+      const key = name || (protoOrDescriptor as ClassElement).key;
+      const cacheKey = typeof key === 'symbol' ? Symbol() : `__${key}`;
       descriptor.get = function(this: LitElement) {
         if ((this as unknown as
-             {[key: string]: Element | null})[key as string] === undefined) {
-          ((this as unknown as {[key: string]: Element | null})[key as string] =
+             {[key: string]: Element | null})[cacheKey as string] === undefined) {
+          ((this as unknown as {[key: string]: Element | null})[cacheKey as string] =
                this.renderRoot.querySelector(selector));
         }
         return (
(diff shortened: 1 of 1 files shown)


justinfagnani (changes requested):
Good catch and fix. We should be catching this in tests though, and definitely should fix that with the merge. I wonder if our Babel tests aren't running?

@sorvell @kevinpschaaf

src/lib/decorators.ts
```diff
@@ -227,15 +227,16 @@ export function query(selector: string, cache?: boolean) {
       configurable: true,
     };
     if (cache) {
-      const key = typeof name === 'symbol' ? Symbol() : `__${name}`;
+      const key = name || (protoOrDescriptor as ClassElement).key;
+      const cacheKey = typeof key === 'symbol' ? Symbol() : `__${key}`;
```
sorvell, line 231: How about we just use a symbol if the name is undefined?

sorvell (changes requested):
Please add a test that verifies this is working as expected. Thanks!

justinfagnani:
The new test would be around here: https://github.com/Polymer/lit-element/blob/master/src/test/lib/decorators_test.ts#L360

src/lib/decorators.ts
bramkragten: I think we should either always use a symbol, or keep it as I have it now for consistency.
Otherwise, the `cacheKey` is different based on the compiler you use.

I'm fine with always using a symbol as cacheKey, I guess `__${name}` was chosen for debugging purposes?

justinfagnani: We should only use a Symbol if the name is a Symbol, since we don't require the Symbol polyfill currently.

justinfagnani:
@bramkragten ping. Are you able to add a test for this?

bramkragten:
Sorry! Been very busy with other stuff, I'll come back to this soon.

PR #1106, opened 2020-11-13
Docs(guide): Bad link for ShadyCSS Polyfill

pranshuchittora:


arthurevans (approved):
Thanks for the fix! LGTM.

Issue #1121, opened 2020-12-07
Use webpack and babel-loader, @query is undefined


WUSO01:
## Description
I wirted lit-element with typescript, i use webpack and babel compile my project. When i use babel-loader, `console.log(@query('xxx'))` is undefined, but i find ts-loader  won't occur this problem.

This is the minimalist demo:

**index.ts**:

```typescript
import { customElement, LitElement, html, query } from 'lit-element'

@customElement('custom-container')
export class CustomContainer extends LitElement {
  @query('#btn') btnEle: HTMLButtonElement

  firstUpdated () {
    console.log('this.btn is:', this.btnEle)
    this.btnEle.addEventListener('click', () => {
      console.log('click')
    })
  }

  render () {
    return html`
      <div>
        <button id="btn">button</button>
      </div>
    `
  }
}
```

**webapck.config.js**:

```javascript
module.exports = {
  // ...
  module: {
   rules: [
      {
        test: /\.ts?$/,
        loader: 'babel-loader',
        exclude: /node_modules/,
        options: {
          cacheDirectory: true,
          cacheCompression: false
        }
      }
    ]
  }
}
```
**babel.config.json**:

```json
{
  "presets": [
    "@babel/env",
    "@babel/preset-typescript"
  ],
  "plugins": [
    [
      "@babel/plugin-proposal-decorators", {
        "legacy": true
      }
    ],
    [
      "@babel/plugin-proposal-class-properties"
    ]
  ]
}
```

[Demo source file](https://github.com/WUSO01/query-babel-loader-issues)

The issues：

![image](https://user-images.githubusercontent.com/21070775/101334033-4ea98a00-38b2-11eb-8929-07a21fb16467.png)

sorvell:
Can you see if the decorator settings here work? This is what we use for testing: https://github.com/Polymer/lit-element/blob/master/.babelrc#L3.

If you'd like additional help debugging the issue, it'd be great if you could provide a minimal runnable reproduction. A simple github repo would be fine.

Hope that helps.

sorvell:
Closing due to age. Please open a new issue with a reproduction if the problem still exists. Thanks.

PR #1138, opened 2021-01-22
fix babel-plugin-proposal-decorators use legacy the query's decorator…

WUSO01:
The original issue #1121 

I Write lit-element like this:

```typescript
@customElement('custom-div')
export class CustomDiv extends LitElement {
  @query('#btn') btn: HTMLButtonElement

  // ....
}
```

And I also used `babel(v7)` and `webpack(v4)`,  I set `@babel/plugin-proposal-decorators`  legacy was true

**babel.config.json**:

```json
{
  "presets": [
    "@babel/env",
    "@babel/preset-typescript"
  ],
  "plugins": [
    [
      "@babel/plugin-proposal-decorators", {
        "legacy": true
      }
    ],
    [
      "@babel/plugin-proposal-class-properties"
    ]
  ]
}
```

When I builded i got an error: `this.btn is undefined`. If you change `legacy` to`decoratorsBeforeExport`,  this issue won't be show.

I founded the issue cased by [legacyQuery](https://github.com/Polymer/lit-element/blob/master/src/lib/decorators.ts#L345)  function didn't use 'return' Statement.


WUSO01:
Fixes #1121

sorvell (changes requested):
This change looks ok, but we need to add testing for babel decorators in legacy mode. Think this is something you could add? Thanks.

WUSO01:
Legacy mode will cause other problems, eg: `updated` lifecycle  changedProperties  always empty Map

PR #1143, opened 2021-01-26
Fix cached query decorator (#1142)

### LINKED ISSUE #1142: Bug: @query broken with cache = true and babel decorators
When using `@query` with `cache: true`, it generates the `key` based on the passed in `name` - but `name` only exists for legacy decorators. This results in the key always being `__undefined`, so multiple uses of `@query` in one component result in all of them using using the same cached value. See the code [here](https://github.com/Polymer/lit-element/blob/master/src/lib/decorators.ts#L230). This should probably be replaced with:

```js
const prop = name === undefined ? protoOrDescriptor.key : name;
const key = typeof prop === 'symbol' ? Symbol() : `__${prop}`;
```

If this looks like a good solution, I can send a PR.

bschlenk:
### Reference Issue
Fixes #1142

### DIFF
--- a/src/lib/decorators.ts
+++ b/src/lib/decorators.ts
@@ -227,7 +227,9 @@ export function query(selector: string, cache?: boolean) {
       configurable: true,
     };
     if (cache) {
-      const key = typeof name === 'symbol' ? Symbol() : `__${name}`;
+      const prop =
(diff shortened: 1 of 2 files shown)


justinfagnani:
Thanks for the fix and test!

I added https://github.com/Polymer/lit-html/issues/1572 to track adding this fix to the next major version we're working on in the lit-html monorepo.

PR #1166, opened 2021-03-07
Update properties.md

danny-andrews:
Correct note on converting from attribute to property.

Attributes can't be `null`. They can only be strings.


justinfagnani:
Thanks for the PR!

Attributes can be `null` in that `getAttribute()` returns `null` if the value isn't present: https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute#non-existing_attributes. The phrasing could just say if the attribute is present or not, though I'd rather check and change the language in the new lit.dev site.

PR #1190, opened 2021-04-29
Prepare 2.5.0 release

justinfagnani:


package.json
```diff
@@ -38,9 +38,11 @@
     "publish-dev": "npm test && VERSION=${npm_package_version%-*}-dev.`git rev-parse --short HEAD` && npm version --no-git-tag-version $VERSION && npm publish --tag dev",
     "release": "np --any-branch --yolo"
   },
-  "author": "The Polymer Authors",
+  "author": "Google LLC",
```
aomarks, line 41: Should we also be updating the copyright headers in the source files, and the LICENSE file? Or maybe not update this here if we just keep the older versions with the old copyright attribution?

aomarks (approved):

package.json
justinfagnani: I'm not sure. I did this one because the package metadata shows up in other places. @graynorton ?

PR #1203, opened 2021-07-12
Run tests and benchmarks on GitHub actions.

### LINKED ISSUE #1202: Move tests and benchmarks to GitHub Actions
We need to move the tests and benchmarks to GitHub Actions now that travis-ci.org is no longer available.

bicknellr:
This PR updates this repo to run its tests and benchmarks on GitHub actions. Fixes #1202.

From #1207, which handled the benchmarks part and was merged into this PR:

> I originally tried the same approach as https://github.com/lit/lit/pull/1967 but found that the import specifiers were being mapped to paths that didn't exist, which caused the benchmarks to time out. I'm not entirely sure why this didn't work, but I suspect it was due to the way that tachometer expects the repo it will benchmark is set up. Particularly, tachometer seems to expect that the package it runs within stays the same and that different versions of the dependencies are what should be effectively benchmarked.
>
> ~~I've updated this repo's benchmark workflow to check out the benchmark package separately so that it can be the root package with lit-element as its dependencies.~~
>
> I've merged the LitElement benchmarks from https://github.com/Polymer/benchmarks into this repo under `benchmarks` and made that folder a private NPM package, which seems to be tachometer's preferred setup.


bicknellr:
Hmm, the benchmarks aren't completing for some reason.

bicknellr:
Also Safari 10 (OS X 10.12) and Firefox (Linux) are consistently timing out.

kevinpschaaf (approved):
LGTM

PR #1208, opened 2021-07-26
Replace local type declarations for polyfill APIs with those imported from the polyfills.

bicknellr:
This PR removes the local types describing the polyfill APIs in favor of the new TS externs from the polyfill repos.

- [x] https://github.com/webcomponents/polyfills/pull/457
- [x] Remove local tarball and use released polyfill packages instead.


bicknellr:
For some reason, I can't add @sorvell as a reviewer.

src/env.d.ts
```diff
@@ -1,20 +1,4 @@
...
-
-interface ShadyDOM {
-  inUse: boolean;
-  flush: () => void;
-}
-
 interface Window {
-  ShadyCSS?: ShadyCSS;
-  ShadyDOM?: ShadyDOM;
   ShadowRoot: typeof ShadowRoot;
```
kevinpschaaf, line 2: Why do we still need `ShadowRoot` on window? Seems like that's in `lib.dom.d.ts`...

kevinpschaaf (approved):
LGTM, just the one question about `ShadowRoot`.

src/env.d.ts
bicknellr: I was limiting the PR to only removing the parts that are now provided externally, but it builds without a lot of other stuff in here. Should I remove those also?

bicknellr: (FWIW, this declaration of `ShadowRoot` is not necesary, but the part below that adds `adoptedStyleSheets` still is.)

bicknellr: Actually, I think I'd rather wait to clean unrelated extra declarations out of here in another PR.

bicknellr: Side note: Now that I think about it, https://github.com/lit/lit/pull/1904 was probably a mistake that will become repairable once this PR is in. Originally, I found that those types were conflicting internally with something (the original JS externs for the polyfills, I think?) so that PR avoided the problem by not adding to `Window`. Now that the polyfill types all have exactly one definition, it might be better to go back to adding Lit-specific special properties to `Window` because they _are_ actually there and a collision would indicate a real problem.

bicknellr: (We talked in a meeting about this, I'm going to go ahead and merge.)

PR #1211, opened 2021-10-31
feat: make banner more apparent

e111077:
From this:

image

To this:

image

a11y contrast checker shows that it's the same a11y rating - there were already problems with blue on white and the rating is not degraded with this bg


arthurevans (approved):
LGTM. I'll do deep links in a follow-up PR.

Issue #867, opened 2019-12-20
Support style-src 'nonce' security policy


redallen:
### Description
Inline styles are not applied and the following console error is thrown when using `static get styles`:
```
Content Security Policy: The page’s settings blocked the loading of a resource at inline (“style-src”).
```

#### Live Demo
https://stackblitz.com/edit/lit-element-example-qpfhsi?file=index.html

#### Steps to Reproduce
1. Add `<meta http-equiv="Content-Security-Policy" content="style-src 'nonce-2726c7f26c'" >` to `<head>` in a browser that does not support constructable stylesheets such as Firefox. For more details on this security policy, see [MDN.](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src)

#### Expected Results
You can pass a nonce to the [`css` tagged template function](https://github.com/Polymer/lit-element/blob/master/src/lib/css-tag.ts) so it creates `<style nonce="2726c7f26c>` when [inserting the `<style>` tag into the document.](https://github.com/Polymer/lit-element/blob/master/src/lit-element.ts#L211)

#### Actual Results
There's no API available to set a nonce using `static get styles`.

### Browsers Affected

- [ ] Chrome
- [X] Firefox
- [X] Edge
- [X] Safari 11
- [X] Safari 10
- [X] IE 11

### Versions
- lit-element: v2.2.1
- webcomponents: v2.4.0

justinfagnani:
I think we could handle this with a `LitElement.styleNonce` static that we apply to all injected `<style>` tags.

mdownes:
Has there been any update on this? Or has the issue been fixed in the new lit repo? It's really stopping me adopt CSP while using lit-element

samaniehsan:
has anybody found a reliable solution to this?

sorvell:
Fixed via https://github.com/lit/lit/pull/2134

andy-blum:
@justinfagnani  - Is there any chance the PR linked above could get back-ported to this repo? Our team is still running on `lit-element@^2.5.1` and we're hitting this issue.

justinfagnani:
@andy-blum yeah, I think we can do that.

Is there any reason you haven't upgraded to LitElement 3.x yet? Things are mostly compatible, so I'm curious what incompatibilities you've run into.

andy-blum:
[...]

PR #1230, opened 2022-11-25
fix: backport CSP support from lit/lit

andy-blum:
### Reference Issue
#867

This PR backports the CSP support from lit/lit by adding `nonce` attributes to injected stylesheets with the value of `window.litNonce`. This should allow projects to use LitElement versions prior to 3.x.x with Safari and CSPs preventing unsafe inline styles.

### DIFF
--- a/src/lit-element.ts
+++ b/src/lit-element.ts
@@ -12,6 +12,9 @@
  * http://polymer.github.io/PATENTS.txt
  */
 
+ const NODE_MODE = false;
+ const global = NODE_MODE ? globalThis : window;
+
 /**
  * The main LitElement module, which defines the [[`LitElement`]] base class and
  * related APIs.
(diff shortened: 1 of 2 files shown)


andy-blum:
@justinfagnani any feedback on this or an estimate when it might be able to be merged in?

andy-blum:
@justinfagnani any chance you can take a look at this?

justinfagnani:
Yes, I can take a look

Any reason why you're still on lit-element 2.0?

justinfagnani:
We're going to want a link back to the original PR from here: https://github.com/lit/lit/pull/2134/files

@andy-blum we'll need to port over the test too

andy-blum:
@justinfagnani Seems like the test setup is different. Hopefully, this is at least on the right track?

PR #1246, opened 2023-04-07
Do not redefine class in `customElement` decorator if already defined

### LINKED ISSUE #3794: customElement decorator causes DOM exception when accidentally executed twice
### Which package(s) are affected?

Lit Core (lit / lit-html / lit-element / reactive-element)

### Description

When using the `@customElement('my-wc')`, I occasionally get DOM exceptions like as follows:

```
Uncaught DOMException: Failed to execute 'define' on 'CustomElementRegistry': the name "litwc-hello-world" has already been used with this registry
    at target.
```

Obviously, this is because I am importing and running the script twice somewhere. However it would be nice to have `customElement` check if it is defined before attempting to re-define it in the registry.

It would be quite simple to implement so I will attempt to make a PR.

### Reproduction

```html
<script type="module" src="my-wc.js"></script>
<!-- Accidentally include the same script twice --!>
<script type="module" src="my-wc.js"></script>
```

### Workaround

No workaround yet. We could change the implementation to check if the `MyWc` tag is registered before defining.

### Is this a regression?

No or unsure. This never worked, or I haven't tried before.

### Affected versions

2.7.2

### Browser/OS/Node environment

Browser: Chrome Version 111.0.5563.146 (Official Build) (arm64)
OS: Mac OS

tronicboy1:
### Reference Issue
Fixes [#3794](https://github.com/lit/lit/issues/3794)

Change `customElement` decorator to not define a custom element if it is already in the registry.

This eliminates DOM exceptions when multiple scripts are accidentally imported.

### DIFF
--- a/src/lib/decorators.ts
+++ b/src/lib/decorators.ts
@@ -48,6 +48,10 @@ interface ClassElement {
 
 const legacyCustomElement =
     (tagName: string, clazz: Constructor<HTMLElement>) => {
+      // Do not define if already registered
+      const isDefined = Boolean(window.customElements.get(tagName));
+      // tslint:disable-next-line:no-any
+      if (isDefined) return clazz as any;
       window.customElements.define(tagName, clazz);
       // Cast as any because TS doesn't recognize the return type as being a
       // subtype of the decorated class when clazz is typed as


justinfagnani:
Even though this is a small code change, it is a major behavior change that would require an RFC. `@customElement()` is only a declarative shorthand for `customElements.define()` - it's been important that the behavior matches. We also don't put features in decorators that are not available in plain JS.

Given a single global registry, and no way to tell if two registrations for the same name are compatible, throwing is the more correct and conservative thing to do. We may wish to explore some other option, but that would involve some way of element definition describing what other definitions they are compatible with and still throwing if they aren't known to be compatible.

tronicboy1:
@justinfagnani 
I totally understand where your coming from with that. You could have multiple packages trying to define the same name and have no way to see that as an error.

I'll close this PR, sorry for wasting your time!

justinfagnani:
@tronicboy1 not a waste. I think we should do something here for use cases that need it, it's just a little complicated. Hopefully scoped element registries solve most cases.

tronicboy1:
@justinfagnani 
First time hearing about scoped registries. Your proposal makes a lot of sense and I for one would be very happy to see it realized! 

thanks again for all of your work.

